PR #3, opened 2015-12-28
add option to prefer locally installed binaries

sindresorhus:
This enables you to `execa('foo')` if you `$ npm install foo` it.

Some open questions:

- Should it be enabled by default? That would really be the nicest solution, but not sure about the consequences and whether there are some downsides I haven't thought of.
- What should the option be named? I would like something better than current. Put your bikeshedding hats on.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,25 +3,37 @@ var childProcess = require('child_process');
 var crossSpawnAsync = require('cross-spawn-async');
 var stripEof = require('strip-eof');
 var objectAssign = require('object-assign');
+var npmRunPath = require('npm-run-path');
+var pathKey = require('path-key')();
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
 	return new Promise(function (resolve, reject) {
-		opts = objectAssign({
-			maxBuffer: TEN_MEBIBYTE
-		}, opts);
-
 		var parsed = crossSpawnAsync._parse(cmd, args, opts);
 
+		opts = objectAssign({
+			maxBuffer: TEN_MEBIBYTE,
+			stripEof: true,
+			preferLocal: true
+		}, parsed.options);
+
 		var handle = function (val) {
-			if (parsed.options.stripEof !== false) {
+			if (opts.stripEof) {
 				val = stripEof(val);
 			}
 
 			return val;
 		};
 
-		var spawned = childProcess.execFile(parsed.command, parsed.args, parsed.options, function (err, stdout, stderr) {
+		if (opts.preferLocal) {
+			opts.env = opts.env || {};
+			opts.env[pathKey] = npmRunPath({
+				cwd: parsed.cwd,
+				path: opts.env[pathKey]
+			});
+		}
+
+		var spawned = childProcess.execFile(parsed.command, parsed.args, opts, function (err, stdout, stderr) {
 			if (err) {
 				err.stdout = stdout;
 				err.stderr = stderr;
--- a/package.json
+++ b/package.json
@@ -31,11 +31,14 @@
   ],
   "dependencies": {
     "cross-spawn-async": "^2.1.1",
+    "npm-run-path": "^1.0.0",
     "object-assign": "^4.0.1",
+    "path-key": "^1.0.0",
     "strip-eof": "^1.0.0"
   },
   "devDependencies": {
     "ava": "*",
+    "cat-names": "^1.0.2",
     "xo": "*"
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -10,6 +10,7 @@
 - Supports [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) binaries cross-platform.
 - [Improved Windows support.](https://github.com/IndigoUnited/node-cross-spawn-async#why)
 - Higher max buffer. 10 MB instead of 200 KB.
+- [Executes locally installed binaries by name.](#preferlocal)
 
 
 ## Install
@@ -76,10 +77,18 @@ Additional exposed options:
 
 #### stripEof
 
-Type: `boolean`  
+Type: `boolean`<br>
 Default: `true`
 
 [Strip EOF](https://github.com/sindresorhus/strip-eof) (last newline) from the output.
+
+#### preferLocal
+
+Type: `boolean`<br>
+Default: `true`
+
+Prefer locally installed binaries when looking for a binary to execute.<br>
+If you `$ npm install foo`, you can then `execa('foo')`.
 
 
 ## License
--- a/test.js
+++ b/test.js
@@ -21,12 +21,22 @@ test('stdout/stderr available on errors', async t => {
 	}
 });
 
+test('execa.shell()', async t => {
+	const {stdout} = await fn.shell('echo foo');
+	t.is(stdout, 'foo');
+});
+
 test('stripEof option', async t => {
 	const {stdout} = await fn('echo', ['foo'], {stripEof: false});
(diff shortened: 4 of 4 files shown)


sindresorhus:
@SamVerschueren @ariporad @vdemedes @jamestalmage Could use some feedback on this if any of you have some extra time.

ariporad:
@sindresorhus: I'd say that it should be on by default, I can't think of any reason you wouldn't want it. Maybe just call it "local"?

sindresorhus:
> Maybe just call it "local"?

Hmm, imagine reading it in code `execa('foo', {local: false});`. Would you have any idea what it meant?

vdemedes:
Haha, I also wanted to suggest `preferLocal` until I looked at the diff.

The behavior should definitely be turned off by default and enabled via an option. Otherwise looks good to me, I'd also go with `preferLocal`. You have the docs, I don't think there will be any confusion.

sindresorhus:
> The behavior should definitely be turned off by default

Why do you think that?

vdemedes:
Because it's not a standard & expected behavior. That's why devs need to know what they really want and how to enable it via an option.

vdemedes:
I mean, it should be off by default to avoid confusion. Like in bash, if I type 'ava', I expect the binary to be in $PATH. If I want "local" file, I use './ava'. Same in Node.js exec()s. 

So when users enable preferLocal, they need to be aware what they're doing.

jamestalmage:
I am not thrilled with `preferLocal`, but I am also not coming up with much better:

- `searchLocally`
- `includeLocal`
- `preferLocalModules`
- `preferLocalBinaries`

 `preferLocallyInstalledBinaries` says everything I want it to, but it's just too long.

sindresorhus:
@vdemedes The goal of this package is to be a better exec, not the same. It already does many things different and unexpected, like stripping of the last EOL. I'm more curious if there's any downside that would cause actual problems by having it on by default? I prefer to have good defaults and let devs turns things off in the off-chance they don't need it, instead of having everything off by default and forcing the dev to manually turn things on.

jamestalmage:
Let me pose a scenario:

I have build `awesome-module` that wants to execute a binary created by `awesome-dep`

```
/Users/user/dev/project/
  node_modules/
    .bin/
      awesome-dep
  awesome-dep@0.3.0/
    ...
  awesome-module/
    node_modules/
      .bin/
        awesome-dep
      awesome-dep@0.4.0/
```

When the `preferLocal` module is used inside `awesome-module`, which one should it pick? Right now I think it goes for `0.3.0`, but I think it would be surprising  for the `awesome-module` developer if they pinned their dependency.

index.js
```diff
@@ -21,6 +23,14 @@ module.exports = function (cmd, args, opts) {
 			return val;
 		};
 
+		if (opts.preferLocal) {
```
SamVerschueren: `if(opts.preferLocal === true)` maybe?

Not sure what you want to do if someone passes in `{preferLocal: 'hello world'}`. The value is truthy, but it's not `true` :).

SamVerschueren:
It's more or less the same discussion we had in the XO repo. What binary should be executed if you type `xo`, the global or the local one?

I think the local option set to `true` would be a better default then set to `false`. At the moment, I can't think of a use case where I want it to be turned off (and off course, there will be use cases enough).

Naming option(s)
- `localBin`

index.js
jamestalmage: Nah, then you force users to cast truthy values to Boolean.

```js
// I want to preferLocal if the user provided a message - for whatever reason.
{
  preferLocal: Boolean(message)
}
```

jamestalmage:
I'm fine with it "on" by default. I do think we should figure out what to do about my example [above](https://github.com/sindresorhus/execa/pull/3#issuecomment-167672873).

index.js
SamVerschueren: Good point :+1:

SamVerschueren:
> When the preferLocal module is used inside awesome-module, which one should it pick? Right now I think it goes for 0.3.0, but I think that would be surprising for the awesome-module developer if they pinned their dependency.

Good question.

What if we change your `awesome-dep` to `AVA` to make it more concrete. In this situation, the developer that installed `awesome-module` has also installed a more current version of `AVA`, for example `AVA@0.8.0`. The dependency is still running `AVA@0.3.0`. Which one should it pick? Because the developer installed a more current version, I would say the version the developer installed (`AVA@0.8.0`). So in your example, `awesome-dep@0.3.0`. But what if `AVA@0.8.0` came from another dependency in the project and wasn't installed by the developer itself? If I'm not mistaken, this is possible with npm@3 because it tries to deduplicate the entire dependency tree (please tell me if I'm wrong :)). Then it might be more suitable to use the pinned dependency of `awesome-module`.

It really feels like it's use case dependent of what's best. So maybe to keep it consistent, I might go for the pinned dependency, in your example `awesome-dep@0.4.0`.

jamestalmage:
> If I'm not mistaken, this is possible with npm@3 because it tries to deduplicate the entire dependency tree (please tell me if I'm wrong :)).

I can't tell from your description if you are wrong or not, but I do know how npm 3 works. :smile:. 

`npm@3` does try to dedupe, but doesn't take it so far it breaks your semver. Let's say you have two dependencies `A` and `B`, which both specify `ava@^0.6.0`, and a two dependencies `C` and `D` which want `ava@^0.9.0`, and the project itself pins to `ava@0.9.1`, you end up with this tree:

```
node_modules/
  A/
    node_modules/
      ava@0.6.1
  B/
    node_modules/
      ava@0.6.1
   C/
     node_modules/
       # no ava here - it got deduped
   D/
     node_modules/
       # no ava here - it got deduped
    ava@0.9.1 # deduped!
```

`A` and `B` do not get deduped dependencies, because their semver spec is incompatible with what is specified in the base project. `C` does get the deduped version because it is semver compatible. If the base projects did not specify `ava` at all (i.e. it wasn't a direct dependency), then `npm` would pick one to dedupe (how / which one it would pick is not entirely clear to me).

SamVerschueren:
> then npm would pick one to dedupe (how / which one it would pick is not entirely clear to me

I tried to test it and I think it picks the one that is installed first, but not 100% sure though.

So in your example, if `B` uses `execa` to execute `AVA`, should it use the deduped version or the pinned version?

jamestalmage:
I think we should do `resolveFrom(dirnam(caller()), ...)`.

sindresorhus:
> When the preferLocal module is used inside awesome-module, which one should it pick? Right now I think it goes for 0.3.0, but I think that would be surprising for the awesome-module developer if they pinned their dependency.

It's up to you as a module author to set `cwd`, or provide an option for it. You set the `cwd` of where you want to execute the binary. Sometimes you'd want to use your local bin, in other situations you'd want for the user to be able to use theirs. And without the `preferLocal` option on it wouldn't find any of those binaries, as none of them would be in the PATH.

> I think we should do resolveFrom(dirnam(caller()), ...).

I don't think that's a good idea because of the above.

sindresorhus:
While I have you here. Are there any other annoyances you've had with child processes that could potentially be fixed in `execa`? And if not. Please do open issues here if you encounter anything annoying with the builtin childprocess module.

jamestalmage:
OK, how about a `resolveFrom` option that allows you to manipulate the lookup path without manipulating `cwd`?

sindresorhus:
> how about a resolveFrom option that allows you to manipulate the lookup path without manipulating cwd?

Do you have an actual use for this? I can't think of any scenario when I'd need to locally spawn local binaries from different location. Maybe open a new issue so not to derail this PR?

PR #5, opened 2016-01-05
Add `all` property to returned object

### LINKED ISSUE #1: Return an `.all` property with both stdout and stderr intermixed

``` js
execa('ava').then(result => {
    console.log(result.all);
});
```
## 

sotojuan:
This fixes #1 by returning an additional `all` property that is just the `stderr` and `stdout` together, updating the readme, and adding two tests that test this feature.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,10 @@
 'use strict';
-var childProcess = require('child_process');
 var crossSpawnAsync = require('cross-spawn-async');
 var stripEof = require('strip-eof');
 var objectAssign = require('object-assign');
 var npmRunPath = require('npm-run-path');
 var pathKey = require('path-key')();
+var spawn = require('cross-spawn-async');
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
@@ -40,7 +40,7 @@ module.exports = function (cmd, args, opts) {
 		var all = '';
 		var out = '';
 		var err = '';
-		var spawned = childProcess.spawn(parsed.command, parsed.args, parsed.options);
+		var spawned = spawn(parsed.command, parsed.args, parsed.options);
 
 		spawned.stdout.setEncoding('utf8');
 		spawned.stderr.setEncoding('utf8');
--- a/index.js
+++ b/index.js
@@ -18,8 +18,12 @@ module.exports = function (cmd, args, opts) {
 		}, parsed.options);
 
 		var handle = function (val) {
-			if (opts.stripEof) {
+			if (parsed.options.stripEof !== false) {
 				val = stripEof(val);
+			}
+
+			if (parsed.options.encoding !== undefined) {
+				val = new Buffer(val, parsed.options.encoding);
 			}
 
 			return val;
@@ -33,17 +37,34 @@ module.exports = function (cmd, args, opts) {
 			});
 		}
 
-		var spawned = childProcess.execFile(parsed.command, parsed.args, opts, function (err, stdout, stderr) {
-			if (err) {
-				err.stdout = stdout;
-				err.stderr = stderr;
-				reject(err);
-				return;
-			}
+		var all = '';
+		var out = '';
+		var err = '';
+		var spawned = childProcess.spawn(parsed.command, parsed.args, parsed.options);
 
+		spawned.stdout.on('data', function (data) {
+			out += data.toString();
+			all += data.toString();
+		});
+
+		spawned.stderr.on('data', function (data) {
+			err += data.toString();
+			all += data.toString();
+		});
+
+		spawned.on('error', function (error) {
+			error.stdout = handle(out);
+			error.stderr = handle(err);
+			error.all = handle(all);
+
+			reject(error);
+		});
+
+		spawned.on('close', function () {
 			resolve({
-				stdout: handle(stdout),
-				stderr: handle(stderr)
+				stdout: handle(out),
+				stderr: handle(err),
+				all: handle(all)
 			});
 		});
 
--- a/index.js
+++ b/index.js
@@ -42,14 +42,17 @@ module.exports = function (cmd, args, opts) {
 		var err = '';
 		var spawned = childProcess.spawn(parsed.command, parsed.args, parsed.options);
 
+		spawned.stdout.setEncoding('utf8');
+		spawned.stderr.setEncoding('utf8');
+
 		spawned.stdout.on('data', function (data) {
-			out += data.toString();
-			all += data.toString();
+			out += data;
+			all += data;
 		});
 
 		spawned.stderr.on('data', function (data) {
-			err += data.toString();
-			all += data.toString();
+			err += data;
+			all += data;
 		});
 
 		spawned.on('error', function (error) {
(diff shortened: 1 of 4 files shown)


sindresorhus:
Ok, I should have been more clear in the original issue. It's not that simple. The user could already do this themselves. What I'm after is the combination of stdout and stderr as it's outputted, just like what you would see in your terminal.

```js
process.stdout.write('1');
process.stderr.write('2');
process.stdout.write('3');
```

Would result in `132` your way. I want `123`.

Will probably have to spawn and gather it manually.

sotojuan:
I was wondering why you said you didn't have enough time to do this! I'll look into it. Thanks for clearing it  up.

sotojuan:
All right, so I use `spawn` and then listen on `data` events in `stdout` and `stderr` and add each to a string for `stdout`, `stderr`, and `all` respectively. I got it, but I am trying to think of something I can use to test it. For some reason if I do `echo good; ls x; echo good` for example (assuming `x` does not exist), the two `echo`s get put together, with the error from `ls` put last.

However, if I do `echo good; ls x; ls .` then it works, with the error from `ls x` in the middle and the result from `ls .` put last.

I am not sure if this is an error in my implementation, but let me know!

sindresorhus:
> but I am trying to think of something I can use to test it

Just create a `fixture.js` file that contains my above code example.

sotojuan:
All right, so this passes all the tests and seems to work, but (IMO) probably could be refactored/cleaned up.

`spawn` [does not seem to have an option for encoding](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options), so all the outputs are saved as strings and if the user set an encoding (anything not `undefined`) the string is converted into a buffer with that encoding. I did this in the `handle` function. I tried using buffers instead of strings but it got messy/wasn't working.

A lot of stuff I did was new to me (I learned stuff at least!), so let me know what you think.

sindresorhus:
> spawn does not seem to have an option for encoding

You use e.g. `stdout.setEncoding()` on the streams instead.

index.js
```diff
@@ -18,20 +18,41 @@ module.exports = function (cmd, args, opts) {
...
-				reject(err);
-				return;
-			}
+		var spawned = childProcess.spawn(parsed.command, parsed.args, parsed.options);
+		var out = '';
+		var err = '';
+		var all = '';
+
+		spawned.stdout.on('data', function (data) {
+			out += data.toString();
```
sindresorhus: Using `.toString()` is a common mistake in Node.js, but this will totally break on unicode where one character might be split between different `data` chunks. Instead use `.setEncoding()` upfront.

index.js
```diff
@@ -18,20 +18,41 @@ module.exports = function (cmd, args, opts) {
...
 		};
 
-		var spawned = childProcess.execFile(parsed.command, parsed.args, parsed.options, function (err, stdout, stderr) {
-			if (err) {
-				err.stdout = stdout;
-				err.stderr = stderr;
-				reject(err);
-				return;
-			}
+		var spawned = childProcess.spawn(parsed.command, parsed.args, parsed.options);
```
sindresorhus: You might as well use the https://github.com/IndigoUnited/node-cross-spawn-async/blob/master/index.js API directly now that we're just spawning.

index.js
```diff
@@ -18,20 +18,41 @@ module.exports = function (cmd, args, opts) {
...
 			}
 
+			if (parsed.options.encoding !== undefined) {
+				val = new Buffer(val, parsed.options.encoding);
+			}
+
 			return val;
 		};
 
-		var spawned = childProcess.execFile(parsed.command, parsed.args, parsed.options, function (err, stdout, stderr) {
```
sindresorhus: Using spawning misses out on a lot of the execFile stuff, like maxBuffer, which will need to be implemented manually, I think. https://github.com/nodejs/node/blob/cc82e5e3d9aefb5e31dbe688b0038327bb7d9b80/lib/child_process.js#L118-L300

Might be easier to just use spawn for the `.all` property.

package.json
```diff
@@ -37,5 +37,8 @@
   "devDependencies": {
     "ava": "*",
     "xo": "*"
+  },
+  "xo": {
+    "ignores": ["fixture.js"]
```
sindresorhus: Already ignored by default: https://github.com/sindresorhus/xo/blob/3462ab6d98adb74c26d2009788d28b03706b55f6/options-manager.js#L18

test.js
```diff
@@ -30,3 +31,8 @@ test('execa.shell()', async t => {
 	const {stdout} = await fn.shell('echo foo');
 	t.is(stdout, 'foo');
 });
+
+test('all', async t => {
```
sindresorhus: title could be a bit more descriptive

index.js
sotojuan: So you'd want to use `spawn` and `execFile` for each command?

index.js
jamestalmage: Aren't you supposed to use string-decoder for that? Or is that what setEncoding uses under the hood?

sindresorhus: @jamestalmage You could. I think it uses it under the hood, but `.setEncoding()` is the correct way regardless.

jamestalmage: http://stackoverflow.com/a/9046164

sindresorhus: Yup. I blame the Node.js docs. They could be a lot clearer. Especially since it's easy to do it wrong and would cause subtle bugs.

index.js
sindresorhus: My last sentence makes no sense. We of course don't want to execute it twice for a command. I think I meant to say that `execFile` returns the child process. So could maybe listen to .stdout/.stderr and gather it and then include it in the `execFile` callback.

Maybe @jamestalmage has a better suggestion.

sotojuan:
Hey Sindre, so my old `all` branch of my fork somehow disappeared/got messed up, so uh I just started with a fresh one based on latest master and forced push. Probably a bad thing to do but the reason I am commenting is because apparently for some reason the `all` output works. It works on the tests, tried it with external files and the fixture, and even shell commands. Not sure why since it's the same code.

sindresorhus:
Force push is fine, and often useful, on non-master branches. Your changes does not yet address all the previous inline feedback.

sotojuan:
Sure, was just explaining the force push! I'll get to work on the rest of the feedback. Thanks!

sotojuan:
All right, I took care of some of the smaller things. The `all` works when running a file but not when doing a shell command (`echo` at least). Still investigating that.

Also, can you clarify on what I should do about [this](https://github.com/sindresorhus/execa/pull/5#discussion-diff-48919697)?

sindresorhus:
Note to self: This should be a lot easier now that we use `spawn` by default. We could just [merge](https://github.com/teambition/merge2) the stdout/stderr streams before passing the merged stream into `get-stream`. This should wait until we're done with all the other work though.

jamestalmage:
https://www.npmjs.com/package/merge-stream

sotojuan:
Ha, I had forgotten about this PR. Yeah there's probably a better way to do this now. Might be worth to just close this one and I (or someone else?) can work with the new proposed solutions.

PR #9, opened 2016-03-09
Fix execa.shell on Windows

### LINKED ISSUE #8: execa.shell not running on windows
This code from the `readme` was tested on windows 7:

```js
execa.shell('echo unicorns').then(result => {
    console.log(result.stdout);
});
```
and the result was no output at all. No error, `unicorns` or anything. 

I also tested known windows commands like `dir`, and `echo`:

```js
execa.shell('dir').then(result => {
    console.log(result.stdout);
});

execa.shell('echo unicorns').then(result => {
    console.log(result.stdout);
});
```
None of these commands produced any results. Just nothing in the console at all. 

However, when I run the `execa` example, it works:

```js
execa('echo', ['unicorns']).then(result => {
    console.log(result.stdout);
});
//=> unicorns
```

Am I missing something simple?

strikeentco:
When you're trying pass `cmd` as command and command as one of the arguments into the `crossSpawnAsync._parse` it makes some terrible things.

Because [`crossSpawnAsync._parse`](https://github.com/IndigoUnited/node-cross-spawn-async/blob/master/lib/parse.js#L111-L112) also adds command as one of the arguments and then uses `cmd` as the command.

After all these steps, this:
C:\Windows\system32\cmd.exe /s /c "echo unicorns"
turns to this:
C:\Windows\system32\cmd.exe /s /c ""C:\Windows\system32\cmd.exe" "/s" "/c" "\"echo unicorns\"""

This PR fix it and closes #8.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,7 +8,20 @@ var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
 	return new Promise(function (resolve, reject) {
-		var parsed = crossSpawnAsync._parse(cmd, args, opts);
+		var parsed;
+
+		if (opts && opts.__winShell === true) {
+			delete opts.__winShell;
+			parsed = {
+				command: cmd,
+				args: args,
+				options: opts,
+				file: cmd,
+				original: cmd
+			};
+		} else {
+			parsed = crossSpawnAsync._parse(cmd, args, opts);
+		}
 
 		opts = objectAssign({
 			maxBuffer: TEN_MEBIBYTE,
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -8,7 +8,20 @@ var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
 	return new Promise(function (resolve, reject) {
-		var parsed = crossSpawnAsync._parse(cmd, args, opts);
+		var parsed;
+
+		if (opts && opts.__winShell === true) {
```
SamVerschueren: Why the `opts.__winShell`? Can't you just do `if (opts && process.platform === 'win32')` instead?

strikeentco: Ooops :scream:

strikeentco: Oh. Stop. This option only needed when we use execa.shell. In execa it shouldn't be.

sindresorhus:
Thanks :)

strikeentco:
You're welcome ;)

PR #11, opened 2016-03-10
Attach pid and kill() to returned promise

### LINKED ISSUE #10: Getting the pid of a long-running process
I'm using execa to start a process that runs "forever" (a web server). I would like to send a kill signal to the process about 30 seconds after launch, but I don't know how/where to get its process id from execa. Is it possible?

At the moment, execa returns a promise that is only resolved after the process ends, which isn't really useful for indefinitely-running processes. It would be nice if execa returned some additional info along with the promise, like the pid.

lukehorvat:
Resolves #10.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,8 @@ var npmRunPath = require('npm-run-path');
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
-	return new Promise(function (resolve, reject) {
+	var spawned;
+	var promise = new Promise(function (resolve, reject) {
 		var parsed;
 
 		if (opts && opts.__winShell === true) {
@@ -45,7 +46,7 @@ module.exports = function (cmd, args, opts) {
 			});
 		}
 
-		var spawned = childProcess.execFile(parsed.command, parsed.args, opts, function (err, stdout, stderr) {
+		spawned = childProcess.execFile(parsed.command, parsed.args, opts, function (err, stdout, stderr) {
 			if (err) {
 				err.stdout = stdout;
 				err.stderr = stderr;
@@ -62,6 +63,11 @@ module.exports = function (cmd, args, opts) {
 
 		crossSpawnAsync._enoent.hookChildProcess(spawned, parsed);
 	});
+
+	promise.kill = spawned.kill.bind(spawned);
+	promise.pid = spawned.pid;
+
+	return promise;
 };
 
 module.exports.shell = function (cmd, opts) {
--- a/readme.md
+++ b/readme.md
@@ -61,7 +61,7 @@ Execute a file directly.
 
 Same options as [`child_process.execFile`](https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback).
 
-Returns a promise for a result object with `stdout` and `stderr` properties.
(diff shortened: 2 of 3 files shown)


SamVerschueren:
Just wondering but it might be cleaner if you did this on [line 52](https://github.com/sindresorhus/execa/pull/11/files#diff-168726dbe96b3ce427e7fedce31bb0bcR51)

```js
promise.kill = spawned.kill;
promise.pid = spawned.pid;
```

Don't really see the added value of `Object.assign`, but maybe I'm missing something :).

Might be nice if you added a test to see of `result.kill` is a function and `result.pid` returns a number?

sindresorhus:
:+1: to what Sam said.

Could you also update the docs about it?

lukehorvat:
I agree with your points; I'll make those changes soon.

I had a thought — do you guys think it might be better if we simply attach the [ChildProcess](https://nodejs.org/api/child_process.html#child_process_class_childprocess) instance (`spawned` in this case) to the promise instead (e.g. `promise.process`)? It already has `kill` and `pid`, as well as other useful methods.

sindresorhus:
> It already has kill and pid, as well as other useful methods.

I thought about it, but then you would have to dig one level deeper. I idea here is that the return value is both a promise and kinda a ChildProcess. Meaning we just attach anything useful from the ChildProcess to the promise. What else should we attach?

lukehorvat:
Updated my commit with those changes. :sunglasses:

I'll leave attaching anything else from ChildProcess to someone else.

index.js
```diff
@@ -7,7 +7,9 @@ var npmRunPath = require('npm-run-path');
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
 module.exports = function (cmd, args, opts) {
-	return new Promise(function (resolve, reject) {
+	var spawned;
+
```
kevva: Remove this line.

index.js
```diff
@@ -62,6 +64,10 @@ module.exports = function (cmd, args, opts) {
 
 		crossSpawnAsync._enoent.hookChildProcess(spawned, parsed);
 	});
+
+	promise.kill = spawned.kill.bind(spawned);
+	promise.pid = spawned.pid;
```
kevva: Add line break after this row.

kevva:
After fixing those small code style issues noted in my inline comments, I think this is looks good to merge. Not sure what else we would need to attach.

lukehorvat:
@kevva Updated my commit. But CI is breaking because my test uses `test.only`. I _need_ to use `test.only` because there is already another test that uses it. Please advise.

SamVerschueren:
Just remove the [`.only`](https://github.com/sindresorhus/ava#running-specific-tests).

lukehorvat:
Test won't run though?

SamVerschueren:
You have to use `async`/`await`.

```js
test('execa() returns a promise with kill() and pid', async t => {
	const promise = await m('echo', ['foo']);
	t.is(typeof promise.kill, 'function');
	t.is(typeof promise.pid, 'number');
});
```

lukehorvat:
Nah, `await` will resolve the promise to a value. I need the promise instance itself in this case.

I'm talking about this line: https://github.com/sindresorhus/execa/blob/d6f0a42/test.js#L24

SamVerschueren:
Remove that one as well. I guess it was committed by accident.

lukehorvat:
@SamVerschueren Done.

SamVerschueren:
LGTM

sindresorhus:
Thanks @lukehorvat :)

PR #14, opened 2016-04-22
add `execa.stdout()` shortcut

sindresorhus:
Not sure about this one. The thinking is that stdout is the most common use-case so might be nice to have a shortcut for it. This is not about the amount of characters, as that's mostly the same, but rather clarity.

Example:

```js
test(async t => {
	t.is((await execa('./cli.js')).stdout, 'foo');
});
```

```js
test(async t => {
	t.is(await execa.stdout('./cli.js'), 'foo');
});
```

You can see how the second one is more readable because of less parens.

-

@kevva @SamVerschueren @jamestalmage @novemberborn Could use your opion 🙌

### DIFF
--- a/index.js
+++ b/index.js
@@ -123,6 +123,20 @@ module.exports = function (cmd, args, opts) {
 	return spawned;
 };
 
+module.exports.stdout = function () {
+	// TODO: set `stderr: 'ignore'` when that option is implemented
+	return module.exports.apply(null, arguments).then(function (x) {
+		return x.stdout;
+	});
+};
+
+module.exports.stderr = function () {
+	// TODO: set `stdout: 'ignore'` when that option is implemented
+	return module.exports.apply(null, arguments).then(function (x) {
+		return x.stderr;
+	});
+};
+
 module.exports.shell = function (cmd, opts) {
 	return handleShell(module.exports, cmd, opts);
 };
--- a/readme.md
+++ b/readme.md
@@ -65,6 +65,14 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
 
 The `child_process` instance is enhanced to also be promise for a result object with `stdout` and `stderr` properties.
 
+### execa.stdout(file, [arguments], [options])
+
+Same as `execa()`, but returns only `stdout`.
+
+### execa.stderr(file, [arguments], [options])
+
+Same as `execa()`, but returns only `stderr`.
+
 ### execa.shell(command, [options])
 
 Execute a command through the system shell. Prefer `execa()` whenever possible, as it's both faster and safer.
--- a/fixtures/noop-err
+++ b/fixtures/noop-err
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+'use strict';
+console.error(process.argv[2]);
--- a/test.js
+++ b/test.js
@@ -17,6 +17,16 @@ test('buffer', async t => {
 	t.is(stdout.toString(), 'foo');
 });
 
+test('execa.stdout()', async t => {
+	const stdout = await m.stdout('noop', ['foo']);
+	t.is(stdout, 'foo');
+});
+
+test('execa.stderr()', async t => {
+	const stderr = await m.stderr('noop-err', ['foo']);
+	t.is(stderr, 'foo');
+});
+
 test('stdout/stderr available on errors', async t => {
 	const err = await t.throws(m('exit', ['2']));
 	t.is(typeof err.stdout, 'string');


jamestalmage:
The one advantage to the first, is that `power-assert` will show you the output of `stderr` in the graph. If you do this, you might as well create the equivalent stderr shortcut. Just so it's easy to flip a test to print stderr instead of stdout without reworking your call structure.

sindresorhus:
Sure, I guess I could add that, but the important question is; Do you think this is a valid addition?

kevva:
Regardless of the parens, I actually don't think one way is more clearer than the other. The first example is how I would *expect* it to work, the latter one would just be a nice shortcut. I could live without it.

jamestalmage:
I don't think it's that helpful, but I haven't written a lot of tests using it.

novemberborn:
Alternatively change the `execa` return value to be a decorated thenable, e.g. add a `stdout` getter that returns a promise:

```js
test(async t => {
    t.is(await execa('./cli.js').stdout, 'foo');
});
```

jamestalmage:
I think Mark's idea is the best so far.

jamestalmage:
Note: it needs to be implemented as a getter that defers creation of additional promises so you don't end up with unhandled rejections.

sindresorhus:
> I think Mark's idea is the best so far.

How is it better?

> Alternatively change the execa return value to be a decorated thenable, e.g. add a stdout getter that returns a promise

What if you want both `stdout` and `stderr` at the same time? Not easily doable, unless I'm missing how this will work.

jamestalmage:
> Why is it better?

Subjective, I like the trailing `.stdout` - it reads like synchronous code.

> What if you want both stdout and stderr at the same time?

both:

```js
const {stdout, stderr} = await execa('./cli');
```

Just one:

```js
const stdout = await execa('./cli').stdout;
```

---

Or maybe just leave it as is:

```js
const {stdout} = await execa('./cli');
```

sindresorhus:
How can `const stdout = await execa('./cli').stdout;` both be a string and promise?

jamestalmage:
It can't - it would always be a promise.

sindresorhus:
@jamestalmage But, then your example wouldn't work:

```js
const {stdout} = await execa('./cli');
```

`stdout` is a string here.

jamestalmage:
Both would work - the await is evaluated before the assignment:

```js
const {stdout} = await execa('./cli');
// =>
const {stdout} = await Promise.resolve({stdout:string, stderr: string});
// =>
const {stdout} = {stdout: string, stderr: string};
// => 
const stdout = string
```

and

```js
const stdout = await exec('./cli').stdout;
// =>
const stdout = await Promise.resolve(string);
// =>
const stdout = string;
```

jamestalmage:
See https://github.com/jamestalmage/execa/commit/70629c6225ce5278184fa8136a45769e7a04fbec#diff-1dd241c4cd3fd1dd89c570cee98b79ddR115 

It can definitely work both ways. Still, not sure it's the best idea

jamestalmage:
My new preference is to merge this, and #18.

#17 was a fun idea, but it conflicts with #18, which I think is more important.

sindresorhus:
So if we're going with this it would be `execa.stdout()` and `execa.stderr()`. I'm not so sure anymore it's worth the extra API bloat and potential confusion. We now have two very similar ways to get stdout/stderr.

jamestalmage:
One potential advantage: https://github.com/sindresorhus/execa/pull/18#issuecomment-213799758

Ignoring the stream you don't care about.

sindresorhus:
Another argument for adding this. When returning the result of `execa` and you only want stdout you currently have to do:

```js
return execa('bin').then(result => result.stdout);
```

This would be nicer:

```js
return execa.stdout('bin');
```

sindresorhus:
@jamestalmage Ready for review.

index.js
```diff
@@ -123,6 +123,20 @@ module.exports = function (cmd, args, opts) {
 	return spawned;
 };
 
+module.exports.stdout = function () {
+	// TODO: set `stderr: 'ignore'` when that option is implemented
```
sindresorhus: Deferring this until https://github.com/sindresorhus/execa/issues/24 is done.

readme.md
```diff
@@ -65,6 +65,14 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
 
 The `child_process` instance is enhanced to also be promise for a result object with `stdout` and `stderr` properties.
 
+### execa.stdout(file, [arguments], [options])
+
+Same as `execa()`, but returns only `stdout`.
+
```
sindresorhus: You mentioning something about memory benefits. Should we mention that? Do you have any source for it?

jamestalmage: It's not true until you set `stderr: ignore` and stop buffering the output.

sindresorhus: Yes, I know, but it will be true, so let's pretend it is, and we can add that info later. Just wondering now.

jamestalmage: No source, I just assumed if you ignore it it dumps the output and does not concat it all together in memory. I think it's a safe assumption.

sindresorhus:
Thanks for the great feedback :)

PR #15, opened 2016-04-22
add `input` option

### LINKED ISSUE #12: `input` option
[`child_process.execFileSync`](https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback) already have this. Would be useful to have it for the async methods too.

You can already do it with `child.stdin.write(stdin)`, but for tests especially, it would be convenient to specify it upfront.

Right now I'm doing this: https://github.com/sindresorhus/strip-indent-cli/blob/4998c5a41810e7316b1765b490f12f73fc4e4f39/test.js#L10

This would be so much nicer:

```js
await execa('./cli.js', {input: '  foo\\n  bar', cwd: __dirname});
```

-

// @kevva @SamVerschueren

jamestalmage:
Fixes #12.

None of the new tests are setup to run on Windows. Is there a cross platform command I can execute that will work well on Travis and AppVeyor?


kevva:
Lol https://github.com/sindresorhus/execa/pull/16.

sindresorhus:
Lol, this is scary close: 🐐

screen shot 2016-04-23 at 05 25 36

// @kevva

kevva:
This has better tests and account for streaming too so :+1: on this.

index.js
```diff
@@ -45,6 +46,21 @@ function handleArgs(cmd, args, opts) {
...
+
+	if (input === null || input === undefined) {
+		return;
+	}
+
+	if (isStream(input)) {
+		input.pipe(spawned.stdin);
+	} else {
+		spawned.stdin.write(input);
+		spawned.stdin.end();
```
sindresorhus: You can pass `input` directly to `.end()`

sindresorhus:
Re [fixture](https://github.com/sindresorhus/execa/blob/input-option/fixtures/stdin), maybe improve it to be:

```
#!/usr/bin/env node
'use strict';
process.stdin.pipe(process.stdout);
```

index.js
```diff
@@ -116,6 +134,9 @@ module.exports.spawn = function (cmd, args, opts) {
 };
 
 module.exports.sync = function (cmd, args, opts) {
+	if (opts && isStream(opts.input)) {
+		throw new TypeError('opts.input can not be a stream in sync mode');
```
```js
throw new TypeError('The `input` option cannot be a stream in sync mode');
```

index.js
```diff
@@ -116,6 +133,9 @@ module.exports.spawn = function (cmd, args, opts) {
 };
 
 module.exports.sync = function (cmd, args, opts) {
+	if (opts && isStream(opts.input)) {
+		throw new TypeError('The `input` option cannot be a stream in sync mode');
+	}
```
kevva: Just nitpicking here, but an extra line break here :).

test.js
```diff
@@ -62,6 +63,48 @@ test.serial('preferLocal option', async t => {
 	process.env.PATH = _path;
 });
 
+if (process.platform !== 'win32') {
```
kevva: Using the fixture, these tests could probably work in Windows too.

fixtures/stdin
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+'use strict';
+process.stdin.pipe(process.stdout);
```
sindresorhus: needs a trailing newline

not warned by XO since it doesn't have an extension.

readme.md
```diff
@@ -108,6 +108,11 @@ Default: `true`
 Prefer locally installed binaries when looking for a binary to execute.<br>
 If you `$ npm install foo`, you can then `execa('foo')`.
 
+#### input
+
+Type: `string` or `buffer` or a `ReadableStream`
```
```
Type: `string` `buffer` `ReadableStream`
```

readme.md
```diff
@@ -108,6 +108,11 @@ Default: `true`
...
+
+Write input to `stdin` of your binary. Streams are not allowed when using `execa.sync(...)`.
```
sindresorhus: > to **the**

sindresorhus: > when using the synchronous methods.

jamestalmage:
Updated.

index.js
```diff
@@ -45,6 +46,20 @@ function handleArgs(cmd, args, opts) {
...
+
+	if (input === null || input === undefined) {
+		return;
+	}
+
+	if (isStream(input)) {
+		input.pipe(spawned.stdin);
+	} else {
+		spawned.stdin.end(input);
+	}
```
kevva: Also nitpicking:

```js
if (isStream(input)) {
    input.pipe(spawned.stdin);
    return;
}

spawned.stdin.end(input);
```

jamestalmage: Hmm. Is there value to that though? I am all for preferring `return` over nesting some huge else block, but for a short one liner, I think `if`/`else` is more readable.

Depending on what we decide, we could make XO catch this one pretty easy.

kevva: Nah, I guess it's a personal preference. If I can, I skip `else` and `return` early. To me at least, it's somewhat clearer.

sindresorhus:
Yay. Thanks James! This is going to make my CLI tests a lot nicer.

index.js
sindresorhus: @kevva Open an issue on https://github.com/sindresorhus/eslint-plugin-xo and we can discuss.

PR #17, opened 2016-04-23
[DO NOT MERGE]: proof of concept - #14 alternative

jamestalmage:
Implements an alternative to #14.

This conflicts with #18, since there would be a namespace conflict for `stdin`, `stdout`, etc, if we did this.

If we think #18 is valuable, then #14 is the better solution for shortcutting access to `stdin` (because it does not conflict)..


sindresorhus:
Closing in favor of #18.

PR #18, opened 2016-04-23
return the child process instead of a promise.

jamestalmage:
We are guessing at the desired properties of the child-process object.

Instead, why not just copy over `.then` and `.catch` from the promise to `spawned` and return that.

This would allow stuff like this:

```js
var spawned = execa('./cli.js');
spawned.stdin.write('foo');
spawned.stdin.end();

const {stdout} = await spawned;
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,10 +117,10 @@ module.exports = function (cmd, args, opts) {
 		handleInput(spawned, parsed.opts);
 	});
 
-	promise.kill = spawned.kill.bind(spawned);
-	promise.pid = spawned.pid;
+	spawned.then = promise.then.bind(promise);
+	spawned.catch = promise.catch.bind(promise);
 
-	return promise;
+	return spawned;
 };
 
 module.exports.shell = function (cmd, opts) {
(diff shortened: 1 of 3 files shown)


kevva:
This is pretty cool :+1:

sindresorhus:
👍 That's a cool idea! I always forget you can thenify something.

sindresorhus:
Can you update the docs? Mainly update what the methods return and remove `The promise instance has a pid property (ID of the child process) and a kill method (for sending signals to the child process).`.

jamestalmage:
Go ahead and merge #14. I'll update this once it's merged so `execa.stdout(...)` returns a `thenified` child process as well.

jamestalmage:
> Can you update the docs?

Yes, absolutely. I was just getting this out there as an idea. It was easier to do a PR than open an issue in this case.

sindresorhus:
@jamestalmage Go ahead and finish this. I'm not ready to merge #14 yet. Haven't decided whether #14 is worth the extra API bloat and overhead.

jamestalmage:
> Haven't decided whether #14 is worth the extra API bloat and overhead.

I think it is. Specifically, if someone does `execa.stdout`, then we can set `stderr` to `'ignore'` or maybe `inherit`. This would be useful if some program prints some very verbose debugging information that we don't want causing buffer overruns. 

> Go ahead and finish this

Will do

jamestalmage:
Docs updated.

sindresorhus:
tumblr_mkwooez15j1qdlh1io1_400

PR #22, opened 2016-04-23
[BREAKING] for consistency return a result object from `execa.sync()` instead of string - fixes #19

### LINKED ISSUE #19: `execa` and `execa.sync` have return type mismatches.
I think it would be better if the following were basically equivalent:

```js
const val = execa.sync(...args);

const val = await execa(...args);
```

This would mean a breaking change to `execa.sync`, using `child.spawnSync` instead, but I think it's worth it to have a more consistent API.

jamestalmage:
Includes commits from #18, so avoid merging until after that is, and I can rebase.

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,9 +143,14 @@ module.exports.sync = function (cmd, args, opts) {
 		throw new TypeError('The `input` option cannot be a stream in sync mode');
 	}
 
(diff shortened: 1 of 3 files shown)


jamestalmage:
Nevermind - there were no conflicts between this and #18 (rebase drop went without a hitch). Merge either/both at will.

sindresorhus:
👍

Issue #20, opened 2016-04-23
Observable support with customizable through filters


jamestalmage:
``` js
var child = execa('./cli.js', [], {stdio: [null, 'observable', null]});

t.plan(5); // should get 5 lines
await child.stdin
  .filter(line => /^s/.test(line)) // only lines that start with "s"
  .forEach(line => t.true(fitsSpecifiedFormat(line)));
```

Additionally, it would be cool if you could use `DuplexStream`s to modify what reaches the observable, I'm thinking "newline-separated-json" byte stream converted to a JSON object stream, etc.

There are a number of "split-stream-XXX" modules already on NPM, but they convert to a stream, would be cool to write `object-stream-to-observable`, and then use that to make CLI output processing awesome.

sindresorhus:
I like it. I always dread when I have to do anything with Node streams. Observables seem perfect for doing this kind of thing.

ehmicky:
I am wondering whether the modern way to do this now (as this issue is 7 years old) would be to use async iterables. Both streams (including `process.stdin|stdout|stderr` and `readline.createInterface()`) are async iterables. Node.js added some functional methods to streams, including `.filter()`, `.map()` and so on.

Based on this, it seems like we might be able to close this issue. What do you think?

sindresorhus:
Agreed

ehmicky:
I have been digging into this issue, and it turns out it is actually quite complicated for users to get right.

First, the `.filter()`, `.map()` functional methods I was mentioning are only available for `Readable` streams, so cannot be used for `process.stdin`. 

Another problem with those functional methods is that they do not allow init and flush logic. This is quite problematic. For example, this prevents defining a mapping method that splits the input line by line (as suggested in #21).

Getting error handling right, especially ensuring that an error thrown in the user mapping logic is propagated is difficult. That's because any error will abort streams, and often a generic `AbortError` will be shown instead. That error is completely opaque, which makes debugging pretty difficult.

Yet another problem is that this does not work when combined with other `stdin`/`stdout`/`stderr`/`stdio` option values. One important case is: it is difficult for users to map `stdout: 'inherit'`, since `childProcess.stdout` would then not be defined.

Other options which could benefit from having this built-in: the `maxBuffer` and `encoding` options. For example, if a transform has a big output, it would be beneficial to make the `maxBuffer` option take this into account.

[...]

PR #26, opened 2016-04-24
DO NOT MERGE - [WIP,Experimental] - Add observable support.

jamestalmage:
Quick hack to make it work.

 Proof of concept for #20.

 I think I really like what this can do for testing

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,7 +131,7 @@ module.exports = function (cmd, args, opts) {
 		handleInput(spawned, parsed.opts);
 	});
 
-	if (opts && opts.stdout === 'observable') {
+	if (opts && (opts.stdout === 'observable' || (opts.stdout && opts.stdout.observable))) {
 		var Observable = tryRequireObservable();
 
 		if (!Observable) {
@@ -140,6 +140,7 @@ module.exports = function (cmd, args, opts) {
 			throw new Error('Observable Not Found');
 		}
 
+		var observableTransform = opts.stdout.transform || split();
 		var stdout = spawned.stdout;
 
 		promise.stdout = new Observable(function (observer) {
@@ -154,7 +155,7 @@ module.exports = function (cmd, args, opts) {
 				});
 
 			stdout
-				.pipe(split())
+				.pipe(observableTransform)
 				.on('data', function (line) {
 					observer.next(line);
 				});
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -117,6 +131,38 @@ module.exports = function (cmd, args, opts) {
 		handleInput(spawned, parsed.opts);
 	});
 
+	if (opts && (opts.stdout === 'observable' || (opts.stdout && opts.stdout.observable))) {
```
jamestalmage: The options handling here is not well thought out. Just wanted to get something working.

test.js
```diff
@@ -108,3 +109,26 @@ test('execa() returns a promise with kill() and pid', t => {
...
 });
+
+test('observable on stdin', t => {
+	t.plan(2);
+
+	return m('stdin', [], {stdout: 'observable', input: 'hello\ngoodbye\nhello'})
+		.stdout
+		.filter(line => /h/.test(line))
+		.map(line => line.toUpperCase())
+		.forEach(line => t.is(line, 'HELLO'));
```
jamestalmage: Overdone example here, just showing the power of `Observables`.

Understand that the `.filter`, `.map` and `.forEach` methods are not synchronous like Array methods. And they fire as lines are encountered. The second call to the `.filter` method could happen long after the first call to `.forEach`.

index.js
```diff
@@ -5,7 +5,21 @@ var stripEof = require('strip-eof');
...
 var pathKey = require('path-key')();
+
+function tryRequireObservable() {
+	if (typeof global.Observable === 'function') {
+		return global.Observable;
+	}
+
+	try {
+		return require('zen-observable');
+	} catch (e) {
```
jamestalmage: - `global.Observable`
- `require('zen-observable')`
- `require('rxjs/Observable')`?

Which order?

sindresorhus:
This will be awesome for testing! I do think we should finish the `execFile` => `spawn` stuff first though.

jamestalmage:
Oh yes - this is throwaway now. I will redo using `stream-to-observable`.

I also want to land an in-progress option handling PR before this as well. (opts.stdin vs opts.stdio).

We need to rework option-handling to make this easy to use anyways.

BarryThePenguin:
Curious how this will differ to [spawn-rx](/paulcbetts/spawn-rx)?

sindresorhus:
Closing this as it hasn't seen any activity in a long time.

PR #27, opened 2016-04-26
Use spawn instead of execFile

jamestalmage:
childProcess.execFile pays no attention `opts.stdio`. This means users can't set `stdio` options when calling `exca()`. You can set `stdio` options when calling `execa.spawn()`, but that does not return a promise.

Rather than having two separate methods `execa()` and `execa.spawn()` - I think we should work towards a single method that allows you to configure how it behaves via `stdio` options. `opts.stdio` or `opts.stdin`, etc; Those should determine whether or not the stream is buffered and made part of the promise result, or if they are just streamed (or sent to an Observable), or ignored. 

To that end, I've reimplemented much of the functionality of `child_process.execFile`. `execFile` is really just a wrapper around `spawn` anyways. It's main utility is that in concats `stdio` and `stdout`. Which is pretty easy for us to achieve on our own using `get-stream`.

This simplifies the API as we can pretty much deprecate `execa.spawn()`. And it offers more control over the behavior as well. Want to use streams/observables for `stdout` because you know it will contain lots of data and you don't want to waste memory buffering? Would it be nice if it still buffered `stderr` for you though? This makes that doable.

Remaining items to be feature complete with execFile:
- [ ] timeout support (should be easy to add back).
- [ ] errname resolution. See TODO note on that. I'm not sure we want to use `process.bind('uv')`. I don't really know what it does.
- [ ] some stream cleanup code (see childProces.execFile's internal `kill` method. It calls `stream.destroy()`. I am not sure how that will affect `get-stream`)
- [ ] add a bunch of tests around the logic copied out of child_process. The new code has nearly 100% coverage, but that's deceptive. Most of these aren't terribly consequential (Error properties: `message`, `killed`, `code`, etc).

Going forward, we'll still need to bikeshed on a few API details, like instead of just `ignore`, `pipe` and `inherit`, we should add `buffer` and `observable`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -194,7 +194,7 @@ module.exports.shell = function (cmd, opts) {
 	return handleShell(module.exports, cmd, opts);
 };
 
-module.exports.spawn = util.deprecate(module.exports, 'execa.spawn: just use execa instead.');
+module.exports.spawn = util.deprecate(module.exports, 'execa.spawn() is deprecated. Use execa() instead.');
 
 module.exports.sync = function (cmd, args, opts) {
 	var parsed = handleArgs(cmd, args, opts);
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 'use strict';
 var childProcess = require('child_process');
+var util = require('util');
 var crossSpawnAsync = require('cross-spawn-async');
 var stripEof = require('strip-eof');
 var objectAssign = require('object-assign');
@@ -179,14 +180,7 @@ module.exports.shell = function (cmd, opts) {
 	return handleShell(module.exports, cmd, opts);
 };
 
-module.exports.spawn = function (cmd, args, opts) {
-	var parsed = handleArgs(cmd, args, opts);
-	var spawned = childProcess.spawn(parsed.cmd, parsed.args, parsed.opts);
-
-	crossSpawnAsync._enoent.hookChildProcess(spawned, parsed);
-
-	return spawned;
-};
+module.exports.spawn = util.deprecate(module.exports, 'execa.spawn: just use execa instead.');
 
 module.exports.sync = function (cmd, args, opts) {
 	var parsed = handleArgs(cmd, args, opts);
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ var stripEof = require('strip-eof');
 var objectAssign = require('object-assign');
 var npmRunPath = require('npm-run-path');
 var isStream = require('is-stream');
-var getStream = require('get-stream');
+var _getStream = require('get-stream');
 var pathKey = require('path-key')();
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
 
@@ -93,24 +93,38 @@ function handleShell(fn, cmd, opts) {
 	return fn(file, args, opts);
 }
 
+function getStream(stream, encoding) {
+	if (!stream) {
+		return null;
+	}
+
+	if (encoding) {
+		return _getStream(stream, {encoding: encoding});
+	}
+	return _getStream.buffer(stream);
+}
+
+function processDone(spawned) {
+	return new Promise(function (resolve) {
+		spawned.on('exit', function (code, signal) {
+			resolve({code: code, signal: signal});
+		});
+
+		spawned.on('error', function (err) {
+			resolve({err: err});
+		});
+	});
+}
+
 module.exports = function (cmd, args, opts) {
 	var parsed = handleArgs(cmd, args, opts);
 	var encoding = parsed.opts.encoding;
 	var spawned = childProcess.spawn(parsed.cmd, parsed.args, parsed.opts);
-	var gs = encoding ? getStream : getStream.buffer;
 
 	var promise = Promise.all([
-		new Promise(function (resolve) {
-			spawned.on('exit', function (code, signal) {
-				resolve({code: code, signal: signal});
-			});
-
-			spawned.on('error', function (err) {
-				resolve({err: err});
-			});
-		}),
-		spawned.stdout && gs(spawned.stdout, encoding && {encoding: encoding}),
-		spawned.stderr && gs(spawned.stderr, encoding && {encoding: encoding})
+		processDone(spawned),
+		getStream(spawned.stdout, encoding),
+		getStream(spawned.stderr, encoding)
 	]).then(function (arr) {
 		var result = arr[0];
 		var stdout = arr[1];
--- a/index.js
+++ b/index.js
@@ -205,10 +205,8 @@ module.exports.sync = function (cmd, args, opts) {
 
 	var result = childProcess.spawnSync(parsed.cmd, parsed.args, parsed.opts);
 
-	if (parsed.opts.stripEof) {
-		result.stdout = stripEof(result.stdout);
-		result.stderr = stripEof(result.stderr);
-	}
+	result.stdout = handleOutput(parsed.opts, result.stdout);
+	result.stderr = handleOutput(parsed.opts, result.stderr);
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
Very cool! I like this a lot. Will give a more thorough review and response later today.

This will also make it easier to achieve https://github.com/sindresorhus/execa/pull/5/files.

jamestalmage:
Can't figure out why Windows is failing. If someone else could give troubleshooting it a go, it would be appreciated

jamestalmage:
Using the technique described [here](http://stackoverflow.com/questions/27688804/how-do-i-debug-error-spawn-enoent-on-node-js), I'm getting the following output:

This Branch(failing):

```
Z:\WebstormProjects\execa>node node_modules/ava/cli.js
spawn called
[ 'C:\\Windows\\system32\\cmd.exe',
  [ '/s',
    '/c',
    '"node "Z:\\WebstormProjects\\execa\\fixtures\\error-message.js""' ],
  { maxBuffer: 10485760,
    stripEof: true,
    preferLocal: true,
    encoding: 'utf8',
    windowsVerbatimArguments: true,
    env: { Path: 'Z:\\WebstormProjects\\execa\\node_modules\\.bin;Z:\\WebstormPr
ojects\\node_modules\\.bin;Z:\\node_modules\\.bin;C:\\Program Files\\nodejs;Z:\\
WebstormProjects\\execa\\fixtures;C:\\Program Files\\Parallels\\Parallels Tools\
\Applications;C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\
Windows\\System32\\WindowsPowerShell\\v1.0\\;c:\\Program Files\\Microsoft SQL Se
rver\\100\\Tools\\Binn\\;c:\\Program Files\\Microsoft SQL Server\\100\\DTS\\Binn
\\;C:\\Program Files\\Microsoft SQL Server\\110\\Tools\\Binn\\;C:\\Users\\Admini
strator\\AppData\\Roaming\\nvm;C:\\Program Files\\nodejs;C:\\Program Files\\Git\
\cmd;C:\\Program Files\\OpenVPN\\bin;C:\\Users\\Administrator\\AppData\\Roaming\
\npm;C:\\Users\\Administrator\\AppData\\Roaming\\nvm;C:\\Program Files\\nodejs'
} } ]

   1 failed

  1. include stdout in errors for improved debugging

  t.regex(err.message, /stdout/)
              |
              "spawn undefined ENOENT"

      _callee6$ (test.js:52:4)

Z:\WebstormProjects\execa>
```

Master (passing)

```
Z:\WebstormProjects\execa>node node_modules/ava/cli.js
execFile called
[ 'C:\\Windows\\system32\\cmd.exe',
  [ '/s',
    '/c',
    '"node "Z:\\WebstormProjects\\execa\\fixtures\\error-message.js""' ],
  { maxBuffer: 10485760,
    stripEof: true,
    preferLocal: true,
    encoding: 'utf8',
    windowsVerbatimArguments: true,
    env: { Path: 'Z:\\WebstormProjects\\execa\\node_modules\\.bin;Z:\\WebstormPr
ojects\\node_modules\\.bin;Z:\\node_modules\\.bin;C:\\Program Files\\nodejs;Z:\\
WebstormProjects\\execa\\fixtures;C:\\Program Files\\Parallels\\Parallels Tools\
\Applications;C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\
Windows\\System32\\WindowsPowerShell\\v1.0\\;c:\\Program Files\\Microsoft SQL Se
rver\\100\\Tools\\Binn\\;c:\\Program Files\\Microsoft SQL Server\\100\\DTS\\Binn
\\;C:\\Program Files\\Microsoft SQL Server\\110\\Tools\\Binn\\;C:\\Users\\Admini
strator\\AppData\\Roaming\\nvm;C:\\Program Files\\nodejs;C:\\Program Files\\Git\
\cmd;C:\\Program Files\\OpenVPN\\bin;C:\\Users\\Administrator\\AppData\\Roaming\
\npm;C:\\Users\\Administrator\\AppData\\Roaming\\nvm;C:\\Program Files\\nodejs'
} } ]

   1 passed

Z:\WebstormProjects\execa>
```

jamestalmage:
I just... I can't... I don't get it. https://github.com/sindresorhus/execa/pull/27/commits/dcb00dca304e56220a5b1160c929830229029d77

jamestalmage:
Previous commit did not trigger travis. Force pushed another.

jamestalmage:
@sindresorhus - How do you want to proceed with this?

Do you want me to keep pushing commits until it's feature complete with `execFile`? (the only thing of real consequence is `timeout` support - which I don't think is widely used).

My only concern is that it's already a pretty major refactor, and I would like to do it in smaller chunks if possible. We can convert my check list from the OP to issues, and just work to fix them before the next release.

sindresorhus:
> I just... I can't... I don't get it. 49065ca

I tried to debug this on my Windows VM, but still no idea why it happen. Fwiw it doesn't happen with `childProcess.spawn()`.

index.js
```diff
@@ -141,14 +194,7 @@ module.exports.shell = function (cmd, opts) {
...
 
-module.exports.spawn = function (cmd, args, opts) {
-	var parsed = handleArgs(cmd, args, opts);
-	var spawned = childProcess.spawn(parsed.cmd, parsed.args, parsed.opts);
-
-	crossSpawnAsync._enoent.hookChildProcess(spawned, parsed);
-
-	return spawned;
-};
+module.exports.spawn = util.deprecate(module.exports, 'execa.spawn: just use execa instead.');
```
sindresorhus: > `execa.spawn()` is deprecated. Use `execa()` instead.

sindresorhus:
This looks really good. Let's merge now and open issues for what's remaining. I too prefer work done in smaller chunks. And you've already covered to important parts. Can you update the readme, removing spawn and any other changes?

kevva:
The tests pass on Windows for me (even with `process.exit(1)`).

sindresorhus:
@kevva What Windows & Node.js version?

kevva:
@sindresorhus, never mind. Ran it in `bash`. Trying to run it in `cmd` gave me `spawn undefined ENOENT` too.

jamestalmage:
Should we open an issue on the Node repo about the weird process.exit behavior? I haven't tried creating a minimal reproduction yet. That would be a first step.

sindresorhus:
Definitely, if you can recreate it. Sounds like a horrible bug if valid.

kevva:
I did create a simple script exiting with error code `1` and got `ENOENT` too. Same result using the core `childProcess.spawn`. Feels like a bug.

jamestalmage:
@kevva - Can you share the script. I'm not able to reproduce with a simple spawn.

sindresorhus:
🙌 🎉

kevva:
:clap:

derimagia:
Make sure when you add links to lines on github you press y to get the current commit - https://github.com/sindresorhus/execa/blob/56a20b1f933b4e9be28f0f47333cd4c008b0f6e4/index.js#L160 is outdated at the moment.

sindresorhus:
@jamestalmage One thing we didn't consider here is that if the user does `execa('echo', ['unicorns']).stdout.pipe(process.stdout);`, to use it like they used `child_process.spawn`, it will still buffer the result, waste memory, and eventually hit the max buffer. We should probably have a `buffer: false` option or something, right?

jamestalmage:
Yeah. Or we could automatically discard the buffer if they pipe.

sindresorhus:
Was thinking about that too, but can we detect that?

PR #28, opened 2016-04-27
Test on Node 6

jamestalmage:


sindresorhus:
AppVeyor doesn't yet have Node.js 6 for some reason. Seems like they manually package them instead of just relying on a version manager. Why am I not surprised...

sindresorhus:
https://twitter.com/appveyor/status/725164672383590400

So we can just keep this open until then.

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - '6'
   - '5'
```
sindresorhus: Can you drop `5`?

> Node.js v5 will continue to be supported for the next two months in order to give developers currently using v5 time to transition to Node.js v6. - https://nodejs.org/en/blog/release/v6.0.0/

jamestalmage: I was thinking of doing that in 2 months! 😄 

Seems I side with @novemberborn here.

jamestalmage:
> AppVeyor doesn't yet have Node.js 6 for some reason
> ...
> So we can just keep this open until then.

Notice that and had the same conclusion

.travis.yml
sindresorhus: Ok, I don't feel strongly either way. Mostly just annoying having to do this update twice :p

kevva:
Node 6 should be available on AppVeyor now.

PR #37, opened 2016-04-30
add max-buffer support

### LINKED ISSUE #33: Reimplement maxBuffer
We lost this when we abandoned execFile.

It should probably be a feature of `getStream`. I will open an issue there.

Reference: https://github.com/nodejs/node/blob/31600735f40e02935b936802ff42d66d49e9769b/lib/child_process.js#L274

jamestalmage:
Fixes #33 

Wait to merge until https://github.com/sindresorhus/get-stream/pull/12 lands.

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,15 +93,24 @@ function handleShell(fn, cmd, opts) {
 	return fn(file, args, opts);
 }
 
-function getStream(stream, encoding) {
-	if (!stream) {
+function getStream(process, stream, encoding, maxBuffer) {
+	if (!process[stream]) {
 		return null;
 	}
 
+	var p;
+
 	if (encoding) {
-		return _getStream(stream, {encoding: encoding});
+		p = _getStream(process[stream], {encoding: encoding, maxBuffer: maxBuffer});
+	} else {
+		p = _getStream.buffer(process[stream], {maxBuffer: maxBuffer});
 	}
-	return _getStream.buffer(stream);
+
+	return p.catch(function (e) {
+		e.stream = stream;
+		e.message = stream + ' ' + e.message;
+		throw e;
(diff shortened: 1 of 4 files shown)


sindresorhus:
LGTM

sindresorhus:
New version of get-stream (2.2.0) is published.

index.js
```diff
@@ -93,15 +93,24 @@ function handleShell(fn, cmd, opts) {
...
 	if (encoding) {
-		return _getStream(stream, {encoding: encoding});
+		p = _getStream(process[stream], {encoding: encoding, maxBuffer: maxBuffer});
+	} else {
+		p = _getStream.buffer(process[stream], {maxBuffer: maxBuffer});
 	}
-	return _getStream.buffer(stream);
+
+	return p.catch(function (e) {
+		e.stream = stream;
```
sindresorhus: `e` => `err`

sindresorhus: Why isn't https://github.com/sindresorhus/eslint-plugin-xo/blob/master/docs/rules/catch-error-name.md catching this?

index.js
```diff
@@ -93,15 +93,24 @@ function handleShell(fn, cmd, opts) {
...
 }
 
-function getStream(stream, encoding) {
-	if (!stream) {
+function getStream(process, stream, encoding, maxBuffer) {
+	if (!process[stream]) {
 		return null;
 	}
 
+	var p;
```
sindresorhus: `p` => `ret`

index.js
jamestalmage: Need to bump `eslint-plugin-xo` [here](https://github.com/sindresorhus/xo/blob/master/package.json#L76)

PR #38, opened 2016-04-30
reimplement errname lookup.

### LINKED ISSUE #31: re-implement errname resolution
#27 dropped execFile in favor of using spawn directly, so we've lost some features.

One is whatever these two lines do:

1. https://github.com/nodejs/node/blob/31600735f40e02935b936802ff42d66d49e9769b/lib/child_process.js#L8
2. https://github.com/nodejs/node/blob/31600735f40e02935b936802ff42d66d49e9769b/lib/child_process.js#L204

Can we reliably just do `process.binding('uv')`? Or is that asking for problems. I know it means importing some native lib, but not much beyond that.

Maybe we look back all the way to Node `0.12` and make sure it was still imported and used the same back then?

jamestalmage:
Fixes #31

I can not for the life of me figure out how to create a test that actually triggers the affected code in `index.js`. I can't create a negative error code in a way that causes `!err` to be true.

### DIFF
--- a/lib/errname.js
+++ b/lib/errname.js
@@ -1,0 +1,38 @@
+'use strict';
+// The Node team wants to deprecate `process.bind(...)`.
+//   https://github.com/nodejs/node/pull/2768
+//
+// However, we need the 'uv' binding for errname support.
+// This is a defensive wrapper around it so `execa` will not fail entirely if it stops working someday.
+//
+// If this ever stops working. See: https://github.com/sindresorhus/execa/issues/31#issuecomment-215939939 for another possible solution.
+var uv;
+
+try {
+	uv = process.binding('uv');
+	if (typeof uv.errname !== 'function') {
+		throw new Error('uv.errname is not a function');
+	}
+} catch (e) {
+	console.error('execa/lib/errname: unable to establish process.binding("uv")', e);
+	uv = null;
+}
+
+function errname(uv, code) {
+	if (uv) {
+		return uv.errname(code);
+	}
+
+	if (!(code < 0)) {
+		throw new Error('err >= 0');
+	}
+
+	return 'UNKNOWN CODE: ' + code;
+}
+
+module.exports = function getErrname(code) {
+	return errname(uv, code);
+};
+
+// used for testing the fallback behavior.
+module.exports._test = errname;
--- a/lib/errname.js
+++ b/lib/errname.js
@@ -10,11 +10,12 @@ var uv;
 
 try {
 	uv = process.binding('uv');
+
 	if (typeof uv.errname !== 'function') {
 		throw new Error('uv.errname is not a function');
 	}
-} catch (e) {
-	console.error('execa/lib/errname: unable to establish process.binding("uv")', e);
+} catch (err) {
+	console.error('execa/lib/errname: unable to establish process.binding(\'uv\')', err);
 	uv = null;
 }
 
@@ -27,7 +28,7 @@ function errname(uv, code) {
 		throw new Error('err >= 0');
 	}
 
-	return 'UNKNOWN CODE: ' + code;
+	return 'Unknown system error ' + code;
 }
 
 module.exports = function getErrname(code) {
@@ -35,4 +36,4 @@ module.exports = function getErrname(code) {
 };
 
 // used for testing the fallback behavior.
-module.exports._test = errname;
+module.exports.__test__ = errname;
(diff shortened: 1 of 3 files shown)


lib/errname.js
```diff
@@ -0,0 +1,38 @@
...
+
+	return 'UNKNOWN CODE: ' + code;
+}
+
+module.exports = function getErrname(code) {
+	return errname(uv, code);
+};
+
+// used for testing the fallback behavior.
+module.exports._test = errname;
```
sindresorhus: Maybe we should have a convention of `__test__` for when you have to attach something for testing?

lib/errname.js
```diff
@@ -0,0 +1,38 @@
...
+// If this ever stops working. See: https://github.com/sindresorhus/execa/issues/31#issuecomment-215939939 for another possible solution.
+var uv;
+
+try {
+	uv = process.binding('uv');
+	if (typeof uv.errname !== 'function') {
+		throw new Error('uv.errname is not a function');
+	}
+} catch (e) {
+	console.error('execa/lib/errname: unable to establish process.binding("uv")', e);
```
sindresorhus: `console.error('execa/lib/errname: unable to establish process.binding('uv')', e);`


and `e` => `err`

lib/errname.js
```diff
@@ -0,0 +1,38 @@
...
+//   https://github.com/nodejs/node/pull/2768
+//
+// However, we need the 'uv' binding for errname support.
+// This is a defensive wrapper around it so `execa` will not fail entirely if it stops working someday.
+//
+// If this ever stops working. See: https://github.com/sindresorhus/execa/issues/31#issuecomment-215939939 for another possible solution.
+var uv;
+
+try {
+	uv = process.binding('uv');
```
sindresorhus: empty line here

test/errname.js
```diff
@@ -0,0 +1,23 @@
...
+
+// simulates failure to capture process.binding('uv');
+function fallback(code) {
+	return errname._test(null, code);
+}
+
+function makeTests(name, m, expected) {
+	test(name, t => {
+		// throws >= 0
+		t.throws(() => m(0), /err >= 0/);
```
sindresorhus: I don't see `m` imported anywhere.

jamestalmage: It's a parameter to `makeTests`

jamestalmage:
Updated.

jamestalmage:
Hmm. Do you think there's more on `stderr` or `stdout` that's not being printed for [this error](https://ci.appveyor.com/project/sindresorhus/execa/build/55/job/ni8woa6980dftrhe#L808)?

test/errname.js
```diff
@@ -1,9 +1,12 @@
+import os from 'os';
 import test from 'ava';
 import errname from '../lib/errname';
 
+const isWin = os.platform() === 'win32';
```
sindresorhus:
```js
const isWin = process.platform === 'win32';
```

sindresorhus:
Not really sure what you're asking.

test/errname.js
```diff
@@ -0,0 +1,28 @@
+import os from 'os';
+import test from 'ava';
+import errname from '../lib/errname';
+
+const isWin = os.platform() === 'win32';
+
+// simulates failure to capture process.binding('uv');
+function fallback(code) {
```
kevva: `const fallback = code = {` looks nicer imo.

sindresorhus: @kevva Join the discussion in https://github.com/sindresorhus/eslint-plugin-xo/issues/3 ;)

test/errname.js
```diff
@@ -0,0 +1,28 @@
+import os from 'os';
+import test from 'ava';
+import errname from '../lib/errname';
+
+const isWin = os.platform() === 'win32';
```
sindresorhus:
```js
const isWin = process.platform === 'win32';
```

jamestalmage:
> Not really sure what you are asking

I couldn't figure out why there was a test error. Turns out I was testing on Node 4 locally when I thought I had switched to Node 0.12.

It seems `uv.errname(-2)` causes a `process.exit` on Windows Node `0.12`.

I'm not even sure how to get negative error codes in Windows (or at all really).

jamestalmage:
I think this is ready for merge. Not sure what to do about the `0.12` issue, but the JS code for the part of `execFile` that calls `uv.errname` is the same all the way back to `0.12`, so I don't think it's a problem.

sindresorhus:
> Not sure what to do about the 0.12 issue

Nothing. We're doing exactly what Node.js is doing and I personally couldn't care less about 0.12.

sindresorhus:
👍

PR #40, opened 2016-05-16
Add `reject` option

### LINKED ISSUE #13: `reject` option
Currently `execa` rejects the promise on non-zero exit code. Sometimes though, you want to get the output regardless, like e.g. running ESLint and reading the result regardless of it having linter issues.

Not totally sold on the name. Suggestions welcome.

Does this option make sense?

kevva:
I think this is all that's needed. Not sure if you're satisfied with the option name but I'll happy change it if you want to.

Fixes #13.

### DIFF
--- a/index.js
+++ b/index.js
@@ -166,6 +166,7 @@ module.exports = function (cmd, args, opts) {
 
 			err.stdout = stdout;
 			err.stderr = stderr;
+			err.failed = true;
 
 			if (!parsed.opts.reject) {
 				return err;
@@ -176,7 +177,9 @@ module.exports = function (cmd, args, opts) {
 
 		return {
 			stdout: handleOutput(parsed.opts, stdout),
-			stderr: handleOutput(parsed.opts, stderr)
+			stderr: handleOutput(parsed.opts, stderr),
+			code: 0,
+			failed: false
 		};
 	});
 
--- a/index.js
+++ b/index.js
@@ -31,7 +31,8 @@ function handleArgs(cmd, args, opts) {
 		maxBuffer: TEN_MEBIBYTE,
 		stripEof: true,
 		preferLocal: true,
-		encoding: 'utf8'
+		encoding: 'utf8',
+		reject: true
 	}, parsed.options);
 
 	if (opts.preferLocal) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
LGTM

index.js
```diff
@@ -165,6 +166,11 @@ module.exports = function (cmd, args, opts) {
 
 			err.stdout = stdout;
 			err.stderr = stderr;
+
+			if (!parsed.opts.reject) {
+				return err;
```
jamestalmage: Should we also add the following:

```js
err.failed = true;
```

And `failed: false` in the passing result?

Also, probably worth adding `code: 0` to the passing result?

If you are using this option, you are probably going to write conditionals around the result, and it would be nice to have as many consistent properties between the `pass` and `fail` results as possible.

kevva: Hmm, yeah, some props to differentiate this from a succeeding result would be nice. Dunno about changing the `code` though. You might want the original one.

sindresorhus: > Also, probably worth adding code: 0 to the passing result?

@kevva Read `passing result` ;)

kevva: Oh, yeah. In that case it makes sense.

kevva: Added. Hopefully it was what you guys meant :).

sindresorhus:
From the readme example:

```js
{
        message: 'Command failed: /bin/sh -c exit 3'
        killed: false,
        code: 3,
        signal: null,
        cmd: '/bin/sh -c exit 3',
        stdout: '',
        stderr: ''
    }
```

Is `killed` the same as `failed`? 

It's also clear https://github.com/sindresorhus/execa/pull/40/files#diff-168726dbe96b3ce427e7fedce31bb0bcR178 is missing some properties.

jamestalmage:
No, I believe `killed` means you called `process.kill()` on it.

jamestalmage:
Also, the timeout option (once implemented) would set killed. That's how it is in execFile

kevva:
So, do we want the other props on the passing result as well or is this good for now?

jamestalmage:
Meh, let's just address that going forward.

PR #42, opened 2016-05-18
`cross-spawn-async` => `cross-spawn`

kevva:
The synchronous stuff was removed from `cross-spawn` as of `3.0.0` so we should be able to use that package now without having to worry about the build toolchain upon install that `spawnSync` needed.


albinekb:
Could this one possibly be released? 0.4.1? 
So that we don't get the annoying warning in all packages using execa 👼

PR #48, opened 2016-07-15
add tests

### LINKED ISSUE #34: Add some tests around the logic copied out of `execFile`
#27 really only went so far as copying the execFile implementation where necessary to get tests passing using spawn.

Most of the required test aren't terribly consequential, but should still be completed. The one set of tests I can think of right now is the creation of the Error property. It builds the command by concatting args together, then uses that to build the error message. Also, it attaches a few extra properties: `killed`, `code`, etc.

Examine the diff of #27 for a complete picture of what should be tested.

I will tackle this, just recording for posterity.

jamestalmage:
Fixes #34

- Adds some tests for functionality copied from `child_process` in #27.
- Normalises the result and error objects further. Each now has `killed`, `signal` and `cmd` properties.
- NYC was not ignoring the `fixtures` directory. Fixed that, and configured nyc in package.json instead of via CLI.

### DIFF
--- a/test.js
+++ b/test.js
@@ -5,8 +5,6 @@ import test from 'ava';
 import getStream from 'get-stream';
 import m from './';
 
-const isWindows = process.platform === 'win32';
-
 process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
 
 test('execa()', async t => {
@@ -193,7 +191,7 @@ test('err.killed is false if process was killed indirectly', async t => {
 	t.false(err.killed);
 });
 
-if (!isWindows) {
+if (process.platform === 'darwin') {
 	test.cb('sanity check: child_process.exec also has killed.false if killed indirectly', t => {
 		const cp = childProcess.exec('forever', err => {
 			t.truthy(err);
@@ -205,7 +203,9 @@ if (!isWindows) {
 			process.kill(cp.pid, 'SIGINT');
 		}, 100);
 	});
+}
 
+if (process.platform !== 'win32') {
 	test('err.signal is SIGINT', async t => {
 		const cp = m('forever');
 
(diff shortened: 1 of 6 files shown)


jamestalmage:
Travis tests just appear to hang. AVA is not exiting. Works fine locally.

sindresorhus:
Maybe try adding `--serial` to pinpoint where it hangs? We really need to add better output when tests hang...

test.js
```diff
@@ -163,3 +165,114 @@ test(`use relative path with '..' chars`, async t => {
...
+		cp.kill();
+	}, 100);
+
+	const err = await t.throws(cp);
+
+	t.true(err.killed);
+});
+
+// TODO: Should this really be the case, or should we improve on child_process?
+test('err.killed is false if process was killed indirectly', async t => {
```
jamestalmage: @sindresorhus - What do you think about this?

Right now, we have this behavior:

```js
cp.kill(); // err.killed will be true
process.kill(cp.pid); // err.killed will be false
```

This mimics the behavior of `child_process.exec`, but it doesn't make a lot of sense to me. If the process is killed, shouldn't `killed` be true regardless of how it was done?

SamVerschueren: I'm with you on this James. If it's killled, it's killed, regardless of how it was done. And if it is killed, jou woild expect that flag to be true.

sindresorhus: Agreed

sindresorhus:
@jamestalmage :shipit: when https://github.com/sindresorhus/execa/pull/48/files#r71064835 is resolved :)

sindresorhus:
Landing this to unblock work on `execa`. We can continue the unresolved issue in https://github.com/sindresorhus/execa/issues/52.

Issue #47, opened 2016-07-15
error from execFile is swallowed


KnisterPeter:
If the file given to execFile is not executable the error result from execFile is swallowed.

sindresorhus:
Can you provide minimal testcase? Can you try `master`? It has a lot of large changes.

KnisterPeter:
I'll give it a try tomorrow and will provide even a PR if thats okay.

kevva:
ping @KnisterPeter

kevva:
I assume this is fixed?

> In master this does not fail anymore, but a test case if it may help to avoid regression

sindresorhus:
A testcase was added in https://github.com/sindresorhus/execa/pull/49

KnisterPeter:
Yes, my testcase is already merges. This could be closed.

Kevin Mårtensson <notifications@github.com> schrieb am Di., 6. Sep. 2016,
14:14:

PR #49, opened 2016-07-16
add test case for throwing on non-executables

KnisterPeter:
add a test case for #47 
In master this does not fail anymore, but a test case if it may help to avoid regression

### DIFF
--- a/test.js
+++ b/test.js
@@ -167,12 +167,11 @@ test(`use relative path with '..' chars`, async t => {
 	t.is(stdout, 'foo');
 });
 
-test('execa() throws if running non executable', async t => {
-	// windows systems do not know executable-flags
-	if (os.platform() !== 'win32') {
+if (os.platform() !== 'win32') {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Would you be able to fix the merge conflict?

KnisterPeter:
Sure :)

KnisterPeter:
I think it may be hard to test on windows boxes :(

KnisterPeter:
I will give it some thoughts tomorrow.

test.js
```diff
@@ -166,6 +167,13 @@ test(`use relative path with '..' chars`, async t => {
 	t.is(stdout, 'foo');
 });
 
+test('execa() throws if running non executable', async t => {
+	// windows systems do not know executable-flags
+	if (os.platform() !== 'win32') {
```
sindresorhus: Better to wrap the whole test in that if statement. Also no need to import `os`. Just use `process.platform`.

Issue #46, opened 2016-06-28
Release?


callumlocke:
Could you release a new version with the `reject` option available?

sindresorhus:
We need to reimplement some features that got lost (https://github.com/sindresorhus/execa/pull/27) when we stopped using the core `execFile` method:

- [ ] https://github.com/sindresorhus/execa/issues/30
- [ ] https://github.com/sindresorhus/execa/issues/32
- [ ] https://github.com/sindresorhus/execa/issues/34

Before that I'm not entirely confident we haven't broken stuff.

jamestalmage:
Agreed.

We are still pre `1.0.0`, so as long as people are using sensible semver strings, a minor bump is basically a breaking one at this point anyways, but we are bound to get complaints if we upgrade before those are fixed.

sindresorhus:
I still haven't had the time to look into the referenced issues and nobody has offered to help out. I've done a release now as we can't keep those issues holding up a release forever. Next time I won't merge such major changes before they're feature equal.

PR #53, opened 2016-08-03
Add `cleanup` option

### LINKED ISSUE #51: Kill child when exiting
If I use `execa.shell` to spawn a long running process, but the original process dies for some reason it would be great if the spawned process gets terminated automatically.

A small example

```js
'use strict'

const execa = require('execa')

execa.shell('ipfs daemon')
  .then((res) => {
    console.log(res.stdout)
  })
  .catch((err) => {
    console.log('failed', err)
  })

console.log('starting')
setTimeout(() => {
  console.log('exiting')
  process.exit()
}, 10 * 1000)
```

Running this will print `exiting` after `10s` and exit but when I grep for `ipfs dameon` the process is still running.

dignifiedquire:
As spawned processes are not cleaned up when the main process dies
this option adds the abilitiy to kill those when the spawning process
dies.

Closes #51

### DIFF
--- a/fixtures/sub-process
+++ b/fixtures/sub-process
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+'use strict';
+var m = require('../');
+
+var cp = m('forever');
+console.log(cp.pid);
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var npmRunPath = require('npm-run-path');
 var isStream = require('is-stream');
 var _getStream = require('get-stream');
 var pathKey = require('path-key')();
+var onExit = require('signal-exit');
 var errname = require('./lib/errname');
 
 var TEN_MEBIBYTE = 1024 * 1024 * 10;
@@ -33,7 +34,8 @@ function handleArgs(cmd, args, opts) {
 		stripEof: true,
 		preferLocal: true,
 		encoding: 'utf8',
-		reject: true
+		reject: true,
+		cleanup: true
 	}, parsed.options);
 
 	if (opts.preferLocal) {
@@ -146,6 +148,13 @@ module.exports = function (cmd, args, opts) {
 	var maxBuffer = parsed.opts.maxBuffer;
 	var spawned = childProcess.spawn(parsed.cmd, parsed.args, parsed.opts);
 
+	var removeExitHandler;
+	if (parsed.opts.cleanup) {
+		removeExitHandler = onExit(function () {
+			spawned.kill();
+		});
+	}
+
 	var promise = Promise.all([
 		processDone(spawned),
 		getStream(spawned, 'stdout', encoding, maxBuffer),
@@ -158,6 +167,10 @@ module.exports = function (cmd, args, opts) {
 		var err = result.err;
 		var code = result.code;
 		var signal = result.signal;
+
+		if (removeExitHandler) {
+			removeExitHandler();
+		}
 
 		if (err || code !== 0 || signal !== null) {
 			if (!err) {
--- a/readme.md
+++ b/readme.md
@@ -11,6 +11,7 @@
 - [Improved Windows support.](https://github.com/IndigoUnited/node-cross-spawn#why)
 - Higher max buffer. 10 MB instead of 200 KB.
 - [Executes locally installed binaries by name.](#preferlocal)
+- [Cleans up spawned processes when the parent process dies.](#cleanup)
 
 
 ## Install
@@ -138,6 +139,12 @@ Default: `true`
 
 Setting this to `false` resolves the promise with the error instead of rejecting it.
 
+#### cleanup
+
+Type: `boolean`<br>
+Default: `true`
+
+Keep track of the spawned process and `kill` it when the parent process exits.
 
 ## License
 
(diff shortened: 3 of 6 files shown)


dignifiedquire:
Looks like this is not working on CI :( passed fine for me locally will investigate what the issue is.

index.js
```diff
@@ -33,7 +34,8 @@ function handleArgs(cmd, args, opts) {
 		stripEof: true,
 		preferLocal: true,
 		encoding: 'utf8',
-		reject: true
+		reject: true,
+		cleanup: false
```
sindresorhus: Any reason we wouldn't want this as `true` by default? I already assumed that's how the core one worked. We could clearly document it in the readme.

readme.md
```diff
@@ -11,7 +11,7 @@
 - [Improved Windows support.](https://github.com/IndigoUnited/node-cross-spawn#why)
 - Higher max buffer. 10 MB instead of 200 KB.
 - [Executes locally installed binaries by name.](#preferlocal)
-
```
sindresorhus: Don't remove the newline

readme.md
```diff
@@ -11,7 +11,7 @@
...
+- [Cleans up spawned processes when the parent process dies](#cleanup)
```
sindresorhus: Should end in a `.`

readme.md
```diff
@@ -138,6 +138,12 @@ Default: `true`
 
 Setting this to `false` resolves the promise with the error instead of rejecting it.
 
+#### cleanup
+
+Type: `boolean`<br>
+Default: `false`
+
+Setting this to `true` keeps track of the spawned process and `kill`s it when the parent process exits.
```
sindresorhus: > Keep track of the spawned process and `kill`s it when the parent process exits.

test.js
```diff
@@ -276,3 +277,24 @@ cmd.title = (message, expected) => `cmd is: ${JSON.stringify(expected)}`;
 test(cmd, ' foo bar', 'foo', 'bar');
 test(cmd, ' baz quz', 'baz', 'quz');
 test(cmd, '');
+
+async function spawnAndKill(t, signal) {
+	const cp = m('sub-process');
+	let pid;
+
```
sindresorhus: `cp.stdout.setEncoding('utf8')`

Then you don't have to `.toString()` it below.

test.js
```diff
@@ -276,3 +277,24 @@ cmd.title = (message, expected) => `cmd is: ${JSON.stringify(expected)}`;
...
+	cp.stdout.on('data', chunk => {
+		pid = parseInt(chunk.toString(), 10);
+		t.is(typeof pid, 'number');
+
+		setTimeout(function () {
```
sindresorhus: use arrow function

fixtures/sub-process
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+'use strict';
+
```
sindresorhus: remove this newline

index.js
dignifiedquire: I was just conscious of this being a breaking change and thought you might not want that, happy to default it to `true`.

sindresorhus: We already have other breaking changes in master. So will be a breaking release regardless.

dignifiedquire: Okay, will change then

dignifiedquire:
@sindresorhus squashed commits and travis seems to be happy now that I introduced a small delay for the pid check.

I've also added a test to make sure it behaves as expected when using `cleanup: false`

test.js
```diff
@@ -276,3 +277,54 @@ cmd.title = (message, expected) => `cmd is: ${JSON.stringify(expected)}`;
...
+
+		setTimeout(() => {
+			process.kill(cp.pid, signal);
+		}, 100);
+	});
+
+	await t.throws(cp);
+
+	// Give everybody some time to breath and kill things
+	const timeout = delay => new Promise(resolve => setTimeout(resolve, delay));
```
sindresorhus: Just use https://github.com/sindresorhus/delay

test.js
```diff
@@ -276,3 +277,54 @@ cmd.title = (message, expected) => `cmd is: ${JSON.stringify(expected)}`;
...
+
+	await timeout(100);
```
sindresorhus: 100=>200 just to be safe

test.js
```diff
@@ -276,3 +277,54 @@ cmd.title = (message, expected) => `cmd is: ${JSON.stringify(expected)}`;
...
+
+	cp.stdout.setEncoding('utf8');
+	cp.stdout.on('data', chunk => {
+		pid = parseInt(chunk, 10);
+		t.is(typeof pid, 'number');
+
+		setTimeout(() => {
+			process.kill(cp.pid, 'SIGINT');
+		}, 100);
+	});
```
sindresorhus: This is almost a duplicate of the other test. Can you extract this into a function both could use?

test.js
dignifiedquire: I knew there was a module somewhere ;)

dignifiedquire:
Squashed and green (at least before squash)

dignifiedquire:
@sindresorhus ping

sindresorhus:
Thank you! :)

Sorry about the delay. I've been quite busy.

Unfortunately we can't do a new release with this until https://github.com/sindresorhus/execa/issues/46 is resolved... Any help welcome.

dignifiedquire:
> Sorry about the delay. I've been quite busy.

No need to be sorry :) Thank you for merging

PR #67, opened 2016-12-20
Throw error in sync method - fixes #66

### LINKED ISSUE #66: Should `execa.sync()` throw on error?
Currently it returns the error as an `.error` property, but I think that's a bit surprising (the async methods reject). At least I didn't realize: https://github.com/sindresorhus/clipboardy/commit/82acb3b7d362098a454db95a1e1fbf2dc6d940c5

Thoughts anyone?

SamVerschueren:
This should fix #66 by throwing an `Error` object if something is written to `stderr`. Also added a test for this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -231,6 +231,10 @@ module.exports.sync = (cmd, args, opts) => {
 
 	const result = childProcess.spawnSync(parsed.cmd, parsed.args, parsed.opts);
 
+	if (result.error || result.status !== 0) {
+		throw (result.error || new Error(result.stderr === '' ? result.stdout : result.stderr));
+	}
+
 	result.stdout = handleOutput(parsed.opts, result.stdout);
 	result.stderr = handleOutput(parsed.opts, result.stderr);
 
--- a/readme.md
+++ b/readme.md
@@ -95,6 +95,8 @@ Same options as [`child_process.execFileSync`](https://nodejs.org/api/child_proc
 
 Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
 
+This method throws an `Error` if the command fails.
+
 ### execa.shellSync(file, [options])
 
 Execute a command synchronously through the system shell.
--- a/test.js
+++ b/test.js
@@ -57,6 +57,10 @@ test('execa.sync()', t => {
 	t.is(stdout, 'foo');
 });
 
+test('execa.sync() throws error if written to stderr', t => {
+	t.throws(() => m.sync('foo'), process.platform === 'win32' ? /'foo' is not recognized as an internal or external command/ : 'spawnSync foo ENOENT');
+});
+
 test('execa.shellSync()', t => {
 	const {stdout} = m.shellSync('node fixtures/noop foo');
 	t.is(stdout, 'foo');


SamVerschueren:
Oh, I totally missed `test.js` in the root. Let me move the test I wrote to that file.

SamVerschueren:
Odd, this works on my Mac and on Windows but apparently not on Linux...

SamVerschueren:
The test doesn't seem to be reliable. Is there any other way I can make sure something is printed to `stderr` instead?

SamVerschueren:
Oh yes, you're totally right! My bad.

I fixed the implementation but couldn't come up with a command that had an error object on Windows. I'm really not familiar with that system. Everything I tried just exited with `status` `1` and data in the `stderr`. So if you have any better ideas on which command we could use that fails on Windows, let me know and I'll fix the tests.

sindresorhus:
Try spawning something that doesn't exist.

SamVerschueren:
Unfortunately that doesn't give an `error`...

> $ cd foo
```
{ status: 1,
  signal: null,
  output: [ null, '', 'The system cannot find the path specified.\r\n' ],
  pid: 684,
  stdout: '',
  stderr: 'The system cannot find the path specified.'
}
```

> $ foo
```
{ status: 1,
  signal: null,
  output:
   [ null,
     '',
     '\'foo\' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n' ],
  pid: 1760,
  stdout: '',
  stderr: '\'foo\' is not recognized as an internal or external command,\r\noperable program or batch file.'
}
```

sindresorhus:
That sounds like a Node.js bug. The `status` (exit code) is not `0`, so it should be an error. You could check the `status` though.

SamVerschueren:
Do you mean in the code? Something like this:

```js
if (result.error || result.status !== 0) {
    throw result.error || new Error(result.stderr);
}
```

sindresorhus:
👍

SamVerschueren:
Finally got it working on both Travis and AppVeyor :p.

index.js
```diff
@@ -231,6 +231,10 @@ module.exports.sync = (cmd, args, opts) => {
 
 	const result = childProcess.spawnSync(parsed.cmd, parsed.args, parsed.opts);
 
+	if (result.error || result.status !== 0) {
+		throw result.error || new Error(result.stdout === '' ? result.stderr : result.stdout);
```
sindresorhus: Wrap it in parens for clarity:

```js
throw (result.error || new Error(result.stdout === '' ? result.stderr : result.stdout));
```

readme.md
```diff
@@ -95,6 +95,8 @@ Same options as [`child_process.execFileSync`](https://nodejs.org/api/child_proc
 
 Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
 
+This method throws an `Error` if something is written to `stderr`.
```
sindresorhus: This needs to reflect the actual code.

index.js
SamVerschueren: Would it make sense to switch the `stdout/stderr` check to something like this

```js
new Error(result.stderr === '' ? result.stdout : result.stderr);
```

sindresorhus: No, I think it's more likely something is on stderr on error.

sindresorhus: Yes, actually, that would make sense.

LinusU:
Can this be merged now? I would love to have it in ❤️

sindresorhus:
It's waiting on https://github.com/sindresorhus/execa/pull/67#discussion_r94870963

SamVerschueren:
@sindresorhus I fixed that. Forgot to notify you.

sindresorhus:
Cool. Thanks Sam :)

SamVerschueren:
You're welcome!

PR #68, opened 2017-01-02
Reimplement stream cleanup behavior from node core. Fixes #32

### LINKED ISSUE #32: reimplement stream cleanup procedures from execFile
Another feature we lost with #27 

See [these lines](https://github.com/nodejs/node/blob/31600735f40e02935b936802ff42d66d49e9769b/lib/child_process.js#L212-L230) from `child_process`.

Right now we aren't destroying those streams. Probably should.

jamestalmage:
I'm not sure how to write a meaningful test for this, as I'm not sure how to end a process without the streams also being destroyed for me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ const stripEof = require('strip-eof');
 const npmRunPath = require('npm-run-path');
 const isStream = require('is-stream');
 const _getStream = require('get-stream');
+const pFinally = require('p-finally');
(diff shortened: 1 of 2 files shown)


sindresorhus:
Yay! Thanks James. Happy xmas and new years :)

LinusU:
Amazing! @sindresorhus can this be released as 0.5.1? ❤️

PR #73, opened 2017-03-01
Add localDir option to specify location of locally installed binaries

### LINKED ISSUE #72: `preferLocal` fails for globally installed CLI programs
When you have a globally installed binary that uses `execa` within it, the `preferLocal` option doesn't work correctly. 

Example: `foo` is a js cli program that depends on `a` (which has a binary installed to `foo`'s node_modules/.bin/ folder.

```sh
npm link # within foo directory
foo doSomethingThatCallsA # within another directory
output: spawn a ENOENT
```

The reason for this is because `npmRunPath` uses the CWD by default. This isn't an issue in npm-run-path necessarily though, because execa should provide an option (along with `preferLocal` to specify the directory to look into.

For example:
```js
// foo.js

execa('a', ['some', 'options'], {
    preferLocal: true,
    localDir: __dirname 
});
```

More than happy to PR this, just want to get approval on the `localDir` option name or to have a better name suggested before I submit a PR 😄

acburdine:
closes #72
- add localDir option that is passed to `npmRunPath`

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,13 +32,14 @@ function handleArgs(cmd, args, opts) {
 		maxBuffer: TEN_MEGABYTES,
 		stripEof: true,
 		preferLocal: true,
+		localDir: parsed.options.cwd || process.cwd(),
 		encoding: 'utf8',
 		reject: true,
 		cleanup: true
 	}, parsed.options);
 
 	if (opts.preferLocal) {
-		opts.env = npmRunPath.env(opts);
+		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
 	return {
--- a/readme.md
+++ b/readme.md
@@ -125,6 +125,13 @@ Default: `true`
 Prefer locally installed binaries when looking for a binary to execute.<br>
 If you `$ npm install foo`, you can then `execa('foo')`.
 
+#### localDir
+
+Type: `string`<br>
+Default: `process.cwd()`
+
+Preferred path to find locally installed binaries in (use with `preferLocal`).
+
 #### input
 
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -38,7 +38,7 @@ function handleArgs(cmd, args, opts) {
 	}, parsed.options);
```
sindresorhus: Can you define the defaults for `cwd` and `localDir` in the Object.assign call instead? Better to have a unified place for defaults.

index.js
```diff
@@ -32,13 +32,14 @@ function handleArgs(cmd, args, opts) {
...
 		preferLocal: true,
+		localDir: parsed.options.cwd || process.cwd(),
 		encoding: 'utf8',
 		reject: true,
 		cleanup: true
 	}, parsed.options);
 
 	if (opts.preferLocal) {
-		opts.env = npmRunPath.env(opts);
+		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
```
acburdine: @sindresorhus I defined the default for `localDir` up above, but we still need the `Object.assign` call here, because `npmRunPath` needs `cwd` to be set to define where it looks, and if we override `cwd` with the value of `opts.localDir` it would mess up the execution of the actual command.

sindresorhus:
I think this could also use a test to prevent future regressions. You could do something like https://github.com/sindresorhus/execa/blob/21b418b67fc896d28fc9befa4bdc9bece2cb0d98/test.js#L75, but create a package.json in `fixtures/local-dir` and use the [`self-path`](https://github.com/sindresorhus/self-path) to assert the relative path is correct.

sindresorhus:
Friendly ping :)

PR #74, opened 2017-03-12
Writing to fast-terminating process causes EPIPE

floatdrop:
For some reason, spawning process that failing on `bash` under Windows does not behaves well – it can crash process with `EPIPE` error (see https://github.com/xavi-/node-copy-paste/pull/53 and https://github.com/zeit/serve/pull/130#issuecomment-285909778).


floatdrop:
How to reproduce:

- Compile `int main() { return 0; }` binary with `gcc main.cpp -o fast-exit`
- Run `execa('fast-exit', [], {input: 'some data'})`

You should get `Uncaught Exception: write EPIPE`.

floatdrop:
@sindresorhus I think this is sufficient for now, because writing to process, that exited with `0` code should not throw `EPIPE` error.

schnittstabil:
> writing to process, that exited with `0` code should not throw `EPIPE` error.

Hmm, at first it seems your patch is independent of the concrete returned status code, isn't it? And secondly, for me, personally, I believe you describe the behavior I would expect :confused:

@floatdrop When do you expect a broken pipe error if not in this case?

floatdrop:
@schnittstabil sadly, at point, when `EPIPE` is raised there is no information about exit code of process. It can be improved a little – `destroyed` property is flipped to `false` in error handler, but I don't think it signals about writing to exited process, it just always be `false` on `EPIPE` errors.

For now executing fast-terminating processes with `input` is broken – I think this is more important thing to fix now.

schnittstabil:
I see, but even with `0` status code, I expect an `EPIPE` error in this scenario. IMO the caller process should be responsible to decide if that error can be ignored or not – neither the called process nor `execa` should be responsible.

floatdrop:
@schnittstabil oh, this _is_ actually better solution. Thanks for pointing it out.

sindresorhus:
Released as 0.6.1

PR #76, opened 2017-03-19
Fix `spawn undefined ENOENT` bug in Windows - fixes #35

### LINKED ISSUE #35: Hunt down weird Windows bug.
See https://github.com/sindresorhus/execa/pull/27#issuecomment-215011451, and subsequent comments.

@kevva - If you could post a reproduction that proves this is a Node bug, we can file an issue in that repo and close this.

SamVerschueren:
This should fix #35.

Tests will fail, will rebase once #77 is merged.

// @jamestalmage @kevva

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,8 @@ function handleArgs(cmd, args, opts) {
 	return {
 		cmd: parsed.command,
 		args: parsed.args,
-		opts
+		opts,
+		parsed
 	};
 }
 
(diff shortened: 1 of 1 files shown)


schnittstabil:
Do we already have a test for this? Maybe the time has come to refactor `module.exports = (…) => {…}` as well – IMO [135 loc](https://github.com/sindresorhus/execa/blob/3025dfce6e907e3239f47066999d99cf70f348e1/index.js#L118-L252) are less-than-ideal.

SamVerschueren:
Don't think we have a test yet. Will have a look if I can add one later today.

sindresorhus:
> Maybe the time has come to refactor module.exports = (…) => {…} as well – IMO 135 loc are less-than-ideal.

Yes, that's sorely needed. Pull request welcome if anyone here has time or interest in doing it :)

PR #79, opened 2017-03-20
Do not output stderr and stdout on stdio inherit - fixes #70

### LINKED ISSUE #70: Inherited stdio results in null stdout, stderr in Error
Ran across a funky error case today while converting gulp-mocha to use execa. If `{stdio: 'inherit'}` is used in execa's options, and the command fails, the resulting Error message contains `nullnull`

Line in question: https://github.com/sindresorhus/execa/blob/master/index.js#L201

Example output:

```
{ Error: Command failed: mocha /.../github/gulp-mocha/test/fixtures/fixture-fail.js
nullnull
    at Promise.all.then.arr (/.../github/gulp-mocha/node_modules/execa/index.js:201:11)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```

FWIW, for commands that succeed, `result.stdout` and `result.stderr` are also both null in the `.then(result =>` handler when `{stdio: 'inherit'{` is used. That isn't mentioned anywhere in the README, and the option appears to be the only way to preserve terminal formatting.

SamVerschueren:
Fixes #70 by not outputting `stderr` and `stdout` when `stdio` is set to `inherit`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,7 +206,8 @@ module.exports = (cmd, args, opts) => {
 
 		if (err || code !== 0 || signal !== null) {
 			if (!err) {
-				err = new Error(`Command failed: ${joinedCmd}\n${stderr}${stdout}`);
+				const output = parsed.opts.stdio === 'inherit' ? '' : `\n${stderr}${stdout}`;
+				err = new Error(`Command failed: ${joinedCmd}${output}`);
 				err.code = code < 0 ? errname(code) : code;
 			}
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

SamVerschueren:
Wasn't sure about how to test this. Check if a newline is not present if `stdio` is set to `inherit`?

SamVerschueren:
Added one

sindresorhus:
👍

PR #36, opened 2016-04-28
[WIP] stdio option handling

jamestalmage:
Fixes #24.

I've only written an options helper method. Not integrated into index.js yet. Started this before #27, and it's backburnered until we get some of the remaining issues from that PR resolved. I'll pick this back up once we've made progress on those.

PR #80, opened 2017-03-22
Better stdio option handling - fixes #24

### LINKED ISSUE #24: Better `stdio` option.
I think the `stdio` array is a pretty poor UI, you have to remember the order of things. I'm constantly pulling up that piece of documentation. Instead, of the three member array, let's just give each one it's own named property:

``` js
execa('./cli.js', [], {
  stdin: someReadableStream,
  stdout: 'pipe',
  stderr: 'inherit'
});
```

SamVerschueren:
Cherry picked from #36 (Thanks @jamestalmage for the initial work!).

Also fixed some issues regarding #79 as it didn't support `stdio` to being an Array.

Feel free to provide feedback on anything, not sure about the docs either.


readme.md
```diff
@@ -146,6 +146,27 @@ Default: `true`
 
 Keep track of the spawned process and `kill` it when the parent process exits.
 
+#### stdin
+
+Type: `string`<br>
+Default: `pipe`
+
+Configure the `stderr` pipe.
```
jamestalmage: We should link to the relevant Node docs and provide a reminder of the available options: https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio

We should also allow positive integers, `undefined`, and `null`.

SamVerschueren:
Made a little change to the docs, is this better? Or should we do something like

> # stdin
> Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio). Shorter and more descriptive form for `stdio: ['inherit', null, null]`.

jamestalmage:
That works. Do we have tests making sure we cover the number and stream scenarios?

SamVerschueren:
I will have a look tomorrow to cover that. Any ideas how we could test the streams?

jamestalmage:
Hmmm.

Using streams with `child_process` does not work how I thought. You can actually only pass native like streams (`net` / `dgram` / `tty`, etc).

I am working on a module to reliably detect those types of streams. (stealing from the code [here](https://github.com/nodejs/node/blob/master/lib/internal/child_process.js#L254-L261)).

If they pass a userland stream, I think we can still accommodate, but it's more of a PITA.

jamestalmage:
Just go with this for now. We can discuss providing more user friendly stream handling later

SamVerschueren:
Added 2 more types of tests

1. Writing content to a file via a file descriptor `fs.openSync('file.txt', 'w');`
2. Add tests to see if parsing numbers works. This wasn't working as expected so I fixed that

I was looking into the stream thing before I read your comment. From the docs

> Share a readable or writable stream that refers to a tty, file, socket, or a pipe with the child process.

So I was expecting I could do something like this

```js
const writeStream = fs.createWriteStream('file.txt');

await execa('foo', {stdout: writeStream});
```

But as you figured out, that is not possible.

test.js
```diff
@@ -69,6 +71,18 @@ test('do not include `stderr` in errors when set to `inherit`', async t => {
...
+test('pass `stdout` to a file descriptor', async t => {
+	const file = tempfile('.txt');
+	await m('fixtures/noop', ['foo bar'], {stdout: fs.openSync(file, 'w')});
+	t.is(fs.readFileSync(file, 'utf8'), 'foo bar\n');
+});
+
+test('pass `stderr` to a file descriptor', async t => {
+	const file = tempfile('.txt');
+	await m('fixtures/noop-err', ['foo bar'], {stderr: fs.openSync(file, 'w')});
+	t.is(fs.readFileSync(file, 'utf8'), 'foo bar\n');
```
jamestalmage: Ah. Good test strategy. File streams are allowed. I opened an issue to handle userland streams better

jamestalmage: Wait, I missed your comment. So this is just a descriptor, not a stream. The way `child_process` handles streams just gets weirder and weirder

jamestalmage:
I like it. Let's merge

test.js
SamVerschueren: Yes exactly. Either the docs are wrong, or I did it wrong. But the docs lik to [Stream](https://nodejs.org/dist/latest-v6.x/docs/api/stream.html#stream_stream) which states

screen shot 2017-03-23 at 19 32 53

> Incorrect value for stdio stream: WriteStream

But but... Ugh, nevermind.

PR #83, opened 2017-03-30
Make killed property true regardless of how it was killed - fixes #52

### LINKED ISSUE #52: Make `.killed === true` regardless of how it was killed
See: https://github.com/sindresorhus/execa/pull/48/files#r71061011

SamVerschueren:
I looked into #52. I really need feedback on this though because I'm not sure if this is 100% correct. My idea is that if the process was exitted with a signal, this means it was killed by something and `killed` should be true.

### DIFF
--- a/index.js
+++ b/index.js
@@ -163,6 +163,7 @@ module.exports = (cmd, args, opts) => {
 
 	const processDone = new Promise(resolve => {
 		spawned.on('exit', (code, signal) => {
+			spawned.killed = signal !== null;
 			cleanupTimeout();
 			resolve({code, signal});
 		});
--- a/test.js
+++ b/test.js
@@ -233,8 +233,19 @@ test('err.killed is true if process was killed directly', async t => {
 	t.true(err.killed);
 });
 
-// TODO: Should this really be the case, or should we improve on child_process?
-test('err.killed is false if process was killed indirectly', async t => {
+test('err.killed is true if process was killed indirectly', async t => {
+	const cp = m('forever');
+
+	setTimeout(() => {
+		process.kill(cp.pid);
+	}, 100);
+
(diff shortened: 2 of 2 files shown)


SamVerschueren:
Off course, Windows... Will check it later today when I have access to my Windows laptop.

index.js
```diff
@@ -163,6 +163,7 @@ module.exports = (cmd, args, opts) => {
 
 	const processDone = new Promise(resolve => {
 		spawned.on('exit', (code, signal) => {
+			spawned.killed = signal !== null;
```
jamestalmage: [See here](https://nodejs.org/api/process.html#process_signal_events)

> Note: Windows does not support sending signals, but Node.js offers some emulation with process.kill(), and ChildProcess.kill(). Sending signal 0 can be used to test for the existence of a process. Sending SIGINT, SIGTERM, and SIGKILL cause the unconditional termination of the target process.

I am guessing `signal` will always be `null` on Windows.

jamestalmage: If this turns out to be true, we may want to reconsider the whole idea. With the current implementation, at least `cp.killed` is consistent across platforms. Maybe we just attach `signal`.

I worry magically doing this for the user on only one platform will create confusing bugs.

sindresorhus: > I am guessing signal will always be null on Windows.

@SamVerschueren Can you check if this is the case?

SamVerschueren:
Lost this one out of my sight. Will have a look again when on my Windows machine.

sindresorhus:
I can't think of a way to do this consistently across platforms.

sindresorhus:
Closing for the above reason. Happy to reopen if anyone can think of a way to do this reliably.

PR #88, opened 2017-05-14
Extend environment by default - fixes #87

### LINKED ISSUE #87: Add `extendEnv` option
Many people, including me, forgets that the `env` option overrides all the environment of the parent. This can cause many weird problem, like binaries not being found because there's no PATH. Would be nice to have a shortcut option that just extends the existing environment instead of replacing it.

From: https://github.com/sindresorhus/clipboardy/pull/16#issuecomment-301237176

SamVerschueren:
This PR extends from the global environment variables by default and implements #87.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,10 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 function handleArgs(cmd, args, opts) {
 	let parsed;
 
+	if (opts && opts.env && opts.envExtend !== false) {
+		opts.env = Object.assign(process.env, opts.env);
+	}
+
 	if (opts && opts.__winShell === true) {
 		delete opts.__winShell;
 		parsed = {
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 function handleArgs(cmd, args, opts) {
 	let parsed;
 
-	if (opts && opts.env && opts.envExtend !== false) {
+	if (opts && opts.env && opts.extendEnv !== false) {
 		opts.env = Object.assign(process.env, opts.env);
 	}
 
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ function handleArgs(cmd, args, opts) {
 	let parsed;
 
 	if (opts && opts.env && opts.extendEnv !== false) {
-		opts.env = Object.assign(process.env, opts.env);
+		opts.env = Object.assign({}, process.env, opts.env);
 	}
 
 	if (opts && opts.__winShell === true) {
--- a/readme.md
+++ b/readme.md
@@ -169,6 +169,13 @@ Default: `true`
 Prefer locally installed binaries when looking for a binary to execute.<br>
 If you `$ npm install foo`, you can then `execa('foo')`.
 
+#### extendEnv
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -16,6 +16,10 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 function handleArgs(cmd, args, opts) {
 	let parsed;
 
+	if (opts && opts.env && opts.envExtend !== false) {
+		opts.env = Object.assign(process.env, opts.env);
+	}
```
SamVerschueren: I did this as the first part of the `handleArgs` in the `Object.assign` block at line `36` (or after that block) because I wasn't sure that `cross-spawn` isn't using that variable internally. But if you want to move it, feel free to suggest so.

SamVerschueren:
I will rebase on master once #89 gets merged.

sindresorhus:
Can you document the `env` option too? Since it differs from the `child_process` one, we should be very clear about it.

SamVerschueren:
Rebased and documented `env`. Any suggestion on the description?

Trying to fix the tests in the mean time on AppVeyor as well.

readme.md
```diff
@@ -125,6 +125,20 @@ Default: `true`
 Prefer locally installed binaries when looking for a binary to execute.<br>
 If you `$ npm install foo`, you can then `execa('foo')`.
 
+#### env
+
+Type: `object`<br>
+Default: `process.env`
+
+Environment key-value pairs. This object is being merged with `process.env`. Set `extendEnv` to `false` if you don't want this.
```
SamVerschueren: Or

> Environment key-value pairs. Extends automatically from `process.env`. Set `extendEnv` to `false` if you don't want this.

SamVerschueren:
I refactored the tests a bit because `process.env.PWD` doesn't seem to exist on Windows.

readme.md
sindresorhus: > Environment key-value pairs. Extends automatically from process.env. Set extendEnv to false if you don't want this.

👍

test.js
```diff
@@ -412,3 +413,15 @@ if (process.platform !== 'win32') {
 		}
 	});
 }
+
+test('extend environment variables by default', async t => {
+	const result = await m.stdout('environment', [], {env: {BAR: 'bar'}});
+
+	t.is(result, `foo\nbar`);
```
sindresorhus: Nitpick, but would be better to split on `\n` and use `t.deepEqual()`.

index.js
```diff
@@ -16,6 +16,10 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 function handleArgs(cmd, args, opts) {
 	let parsed;
 
+	if (opts && opts.env && opts.extendEnv !== false) {
+		opts.env = Object.assign(process.env, opts.env);
```
SamVerschueren: When looking back at the code, this way we extend `process.env` itself as well. Might be better to do `Object.assign({}, process.env, opts.env)` to make sure we don't alter external things.

kevva: :+1:

SamVerschueren:
Fixed the last issue with the `process.env` altering, so I guess it's good to go now.

sindresorhus:
Awesome

PR #92, opened 2017-06-05
Document all the `child_process` options - fixes #90

### LINKED ISSUE #90: Document all the `child_process` options
People don't seem to notice `execa` supports all the `child_process` options. Would be nice to have the documentation for them here.

SamVerschueren:
I documented all the `child_process` options as described in #90. Feedback appreciated :)!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -108,7 +108,59 @@ Returns the same result object as [`child_process.spawnSync`](https://nodejs.org
 
 ### options
 
-Additional options:
+Type: `object`
+
+#### cwd
+
+Type: `string`<br>
+Default: `process.cwd()`
+
+Current working directory of the child process.
+
+#### env
+
+Type: `object`<br>
+Default: `process.env`
+
+Environment key-value pairs.
+
+#### argv0
+
+Type: `string`
+
+Explicitly set the value of `argv[0]` sent to the child process. This will be set to `command` or `file` if not specified.
+
+#### stdio
+
+Type: `Array` `string`<br>
+Default: `pipe`
+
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -108,7 +108,59 @@ Returns the same result object as [`child_process.spawnSync`](https://nodejs.org
 
 ### options
 
-Additional options:
+Type: `object`
```
sindresorhus: `object` => `Object`

sindresorhus:
With this, you can remove all the `Same options as ...` sentences.

You also need to document the following:

```
encoding <string> (Default: 'utf8')
timeout <number> (Default: 0)
maxBuffer <number> Largest amount of data in bytes allowed on stdout or stderr.
```

SamVerschueren:
What about `killSignal`?

```
killSignal <string> | <integer> The signal value to be used when the spawned process will be killed. (Default: 'SIGTERM')
```

SamVerschueren:
What about `killSignal`?

```
killSignal <string> | <integer> The signal value to be used when the spawned process will be killed. (Default: 'SIGTERM')
```

kevva:
> What about killSignal?

Good catch. Actually, I don't think that one works as it stands https://github.com/sindresorhus/execa/blob/master/index.js#L160. `parsed.killSignal` should be `parsed.opts.killSignal` I think.

SamVerschueren:
I think you are right. The `parsed` object doesn't return `killSignal` like that... https://github.com/sindresorhus/execa/blob/master/index.js#L47

sindresorhus:
🙌

PR #94, opened 2017-06-10
Ability to turn on infinite buffer when .then() is used

### LINKED ISSUE #84: Ability to turn off buffering
If you only pipe the output it's a waste of resources to buffer it and the user can also easily hit the `maxBuffer`.

Either `{buffer: false}` or detect when `.then()` is used. I would prefer the latter, but not sure how feasible it is.

---

This would have prevented https://github.com/sindresorhus/gulp-mocha/issues/169

gabrielcsapo:
fixes #84 😄 

`large.txt` is a 20mb file, didn't have time to think of a better solution, so I am all ears for a better way to test the fix!

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,6 +115,9 @@ function getStream(process, stream, encoding, maxBuffer) {
 	} else {
 		ret = _getStream.buffer(process[stream], {maxBuffer});
 	}
+	process[stream].on('infiniteBuffer', () => {
+		process[stream]._readableState.pipes.getBufferedLength = () => 0;
+	});
 
 	return ret.catch(err => {
 		err.stream = stream;
@@ -267,8 +270,17 @@ module.exports = (cmd, args, opts) => {
 
 	handleInput(spawned, parsed.opts);
 
-	spawned.then = promise.then.bind(promise);
 	spawned.catch = promise.catch.bind(promise);
+
+	Object.defineProperty(spawned, 'then', {
+		get: () => {
+			// If the maxBuffer is the default set it to infinity
+			if (maxBuffer === TEN_MEGABYTES && spawned.stdout) {
+				spawned.stdout.emit('infiniteBuffer');
+			}
+			return promise.then.bind(promise);
+		}
+	});
 
 	return spawned;
 };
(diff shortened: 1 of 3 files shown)


gabrielcsapo:
@sindresorhus looks like the appveyor job is broken, won't clone correctly to run tests?

SamVerschueren:
I had a different idea in mind. Will experiment with it first before taking conclusions as to which approach is better. This solution feels more tacky to be honest. The thing is that we don't want to buffer at all when streams are being used.

gabrielcsapo:
@SamVerschueren I don't know what you mean by tacky but 😅 I get not wanting to buffer output but the stream is already being drained, but it makes no sense to stop the stream and alter the data flow. 

Anyway good luck on another solution, I thought this one was minimally invasive and did not alter too much since it is only 14 lines.

gabrielcsapo:
@sindresorhus thoughts on this? I don't think it is hacky, because in your issue you wanted to only change type based on the context of someone getting `.then`.

SamVerschueren:
Like I said before, it shouldn't buffer at all. Like mentioned in the issue itself, buffering is a waste of resources if it's not used. Increasing maxBuffer to Infinity doesn't fix that. Imagine streaming 100GB of data, it doesn't fit the memory. That's why I said it feels hacky, it was not meant to be rude. Sorry if it felt like that.

To fix this, I believe that [this block of code](https://github.com/sindresorhus/execa/blob/master/index.js#L198-L264) shouldn't be executed by default and should be handled lazy, only when `.then` or `.catch` is called. I have something in the works for this.

gabrielcsapo:
@SamVerschueren the code execution does not happen synchronously, I could initiate `execa` and then 4 lines later call .then.

I was hoping for more feedback rather than just being told it doesn't work. I can't learn that way 😉.

I was going to suggest there be a way to pause the stream in get-stream, but the issue with that module is that it returns a promise and not a raw stream. I was going to replace it, but wanted to get feedback before I went ahead with a more complex rewrite of that file.

gabrielcsapo:
@SamVerschueren this might be a good module to use https://github.com/dominictarr/pause-stream, it would allow the stream to be paused, when .then is called, we would drain the stream and then pipe the remaining content.

PR #95, opened 2017-06-10
Do not buffer when streams are used - fixes #84

### LINKED ISSUE #84: Ability to turn off buffering

SamVerschueren:
Alternative implementation to #94. Didn't wrote tests for it yet because I have no time now. Just wanted to get feedback in case I missed something. Guess this should work because the promise is now lazy instead of eager.

### DIFF
--- a/index.js
+++ b/index.js
@@ -267,13 +267,8 @@ module.exports = (cmd, args, opts) => {
 
 	handleInput(spawned, parsed.opts);
 
-	spawned.then = (onfulfilled, onrejected) => {
-		return handlePromise().then(onfulfilled, onrejected);
-	};
-
-	spawned.catch = onrejected => {
-		return handlePromise().catch(onrejected);
-	};
+	spawned.then = (onfulfilled, onrejected) => handlePromise().then(onfulfilled, onrejected);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -267,8 +267,13 @@ module.exports = (cmd, args, opts) => {
 
 	handleInput(spawned, parsed.opts);
 
-	spawned.then = promise.then.bind(promise);
-	spawned.catch = promise.catch.bind(promise);
+	spawned.then = (onfulfilled, onrejected) => {
+		return handlePromise().then(onfulfilled, onrejected);
+	};
```
sindresorhus: Use inline arrow function.

sindresorhus:
It looks like the correct solution to me.

SamVerschueren:
Fixed and added a test.

PR #102, opened 2017-07-13
Fix extending from process.env when no env option is provided - fixes #101

### LINKED ISSUE #101: Child process environment is not extended from process.env
Child process environment is not extended from process.env.

On line 19 : `if (opts && opts.env && opts.extendEnv !== false)` evaluates to `false` because `opts.env` is undefined.

SamVerschueren:
Refactored the `opts` checks a little bit so it is more clear on what is going on. Fixes #101

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,11 +16,16 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 function handleArgs(cmd, args, opts) {
 	let parsed;
 
-	if (opts && opts.env && opts.extendEnv !== false) {
+	opts = Object.assign({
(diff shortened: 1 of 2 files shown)


SamVerschueren:
@sindresorhus @jfmengels Any idea why the build on AppVeyor is failing with the error

> Error: Failed to load plugin ava: Cannot find module 'eslint-plugin-ava'

sindresorhus:
> Error: Failed to load plugin ava: Cannot find module 'eslint-plugin-ava'

I think it was an issue with a npm version.

PR #107, opened 2017-08-19
Enhance error object in sync methods - fixes #98

### LINKED ISSUE #98: shellSync error should also return stdout
When throwing an error, `execa.shellSync` seems to return the error message as a string.

IMO it should also return stdout, as programs can return both stdout and stderr (for example `ls ; cat does-not-exist.txt`)

FWIW, that's `child_process.execSync()` behavior.

SamVerschueren:
This PR fixes #98 by enhancing the error object in the `sync` methods just like we did in the async methods.

I also believe we should extract the utility functions like `handleArgs`, `handleInput`, ... to a `utils` file or something to make the `index` file a little lighter and easier to process by people.

### DIFF
--- a/index.js
+++ b/index.js
@@ -341,7 +341,6 @@ module.exports.sync = (cmd, args, opts) => {
 		stderr: handleOutput(parsed.opts, result.stderr),
 		code: 0,
 		failed: false,
-		killed: false,
 		signal: null,
 		cmd: joinedCmd,
 		timedOut: false
--- a/index.js
+++ b/index.js
@@ -128,16 +128,62 @@ function getStream(process, stream, encoding, maxBuffer) {
 	});
 }
 
-module.exports = (cmd, args, opts) => {
+function makeError(result, options) {
+	const stdout = result.stdout;
+	const stderr = result.stderr;
+
+	let err = result.error;
+	const code = result.code;
+	const signal = result.signal;
+
+	const parsed = options.parsed;
+	const joinedCmd = options.joinedCmd;
+	const timedOut = options.timedOut || false;
+
+	if (!err) {
+		let output = '';
+
+		if (Array.isArray(parsed.opts.stdio)) {
+			if (parsed.opts.stdio[2] !== 'inherit') {
+				output += output.length > 0 ? stderr : `\n${stderr}`;
+			}
+
+			if (parsed.opts.stdio[1] !== 'inherit') {
+				output += `\n${stdout}`;
+			}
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -179,13 +225,13 @@ module.exports = (cmd, args, opts) => {
 
 		spawned.on('error', err => {
 			cleanupTimeout();
-			resolve({err});
+			resolve({error: err});
```
SamVerschueren: Renamed `err` to `error` because of the `makeError` function which uses `error`. The reason is that `spawnSync` uses `error` and I thought it would be easier to rename it here. 

I can also do `const err = result.err || result.error` in the `makeError` function. Don't really have a strong opinion on this one.

index.js
```diff
@@ -292,21 +314,38 @@ module.exports.shell = (cmd, opts) => handleShell(module.exports, cmd, opts);
...
 
-	if (result.error || result.status !== 0) {
-		throw (result.error || new Error(result.stderr === '' ? result.stdout : result.stderr));
-	}
+	if (result.error || result.status !== 0 || result.signal !== null) {
+		const err = makeError(result, {
+			joinedCmd,
+			parsed
+		});
+
```
SamVerschueren: The async methods have one more property though, the `killed` property.

https://github.com/sindresorhus/execa/blob/master/index.js#L240-L243   

I'm not sure what to do with it because it doesn't really look like it is set or something. Any ideas on this one?

index.js
sindresorhus: > Renamed err to error because of the makeError function which uses error. The reason is that spawnSync uses error and I thought it would be easier to rename it here.

👍

index.js
sindresorhus: That's because `killed` is part of the sub-process object, but the sync methods only return the result, not the sub-process:

screen shot 2017-09-29 at 13 19 38

It doesn't really make sense on the sync methods, so we should just document that's it's only available on the async methods.

SamVerschueren:
Removed the `killed` property entirely. Should we document it or is it okay as is?

sindresorhus:
Can you add an example in the readme? Like the `execa.shell('exit 3')`, but for the sync method and without `killed`.

SamVerschueren:
Done

sindresorhus:
> I also believe we should extract the utility functions like handleArgs, handleInput, ... to a utils file or something to make the index file a little lighter and easier to process by people.

Agreed

sindresorhus:
Looks great 🦄

PR #108, opened 2017-08-19
Add fork method - fixes #65

### LINKED ISSUE #65: Add `.fork()` method
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/execa/issues/65)

We could just use `execa('node', ['becomeTheOne.js'])`, but that would mean not capitalising on the added optimization that happens when using fork, as mentioned [here](http://stackoverflow.com/a/17861879/1920325).

There is no spoon matrix

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/execa/issues/65).

SamVerschueren:
It's a start for the `execa.fork` method. Still need some extra work though but would like some feedback at the initial approach.

One thing I noticed with the `stdio` option of the original `child_process.fork` method is that it looks like it ignores the option whenever it's a string. So passing `ignore` to `stdio` results in `[0, 1, 2, 'ipc']`. Even passing `stdio: 'foo'` results in the same behaviour.

### DIFF
--- a/fixtures/ipc-process
+++ b/fixtures/ipc-process
@@ -1,0 +1,9 @@
+#!/usr/bin/env node
+'use strict';
+process.on('message', m => {
+	if (m === 'ping') {
+		process.send('pong');
+	} else {
+		process.send('rainbow');
+	}
+});
--- a/fixtures/ipc-process.js
+++ b/fixtures/ipc-process.js
@@ -1,8 +1,7 @@
-#!/usr/bin/env node
 'use strict';
 process.on('message', m => {
 	if (m === 'ping') {
-		process.send('pong');
+		process.send(process.argv[2] || 'pong');
 	} else {
 		process.send('rainbow');
 	}
--- a/index.js
+++ b/index.js
@@ -278,18 +278,32 @@ module.exports = (cmd, args, opts) => {
 	return spawned;
 };
 
-module.exports.fork = function (cmd, args, opts) {
+module.exports.fork = function (modulePath, args, opts) {
+	if (!Array.isArray(args)) {
+		opts = args;
+		args = [];
+	}
+
 	opts = Object.assign({
-		stdio: 'ipc'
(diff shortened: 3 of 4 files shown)


SamVerschueren:
I expanded the functionality in the latest commit. Something I missed in the first place is that the first argument of `.fork()` is the `modulePath` to a JavaScript file while the first argument of `spawn` is the command. So currently, I implemented it the exact same way. Not sure if we want to improve this?

SamVerschueren:
I was thinking this through once again. Actually I don't like an extra method, it just confuses people. If you google for it, a lot of SO posts can be found regarding `fork Vs. spawn Vs. execFile`. So maybe it would be better if we just add an extra option to the default `execa` method, something like `ipc: true` which would always set the 4th argument of the `stdio` array to `ipc`. Just a suggestion, but I think it's worth a discussions.

// @kevva @jfmengels @jamestalmage

sindresorhus:
@SamVerschueren It's not just the IPC though, it also spawns the Node.js binary only, so it's a bit different from `execa()`. How about `execa.node()`? Is that clear enough?

SamVerschueren:
Totally missed this one :).

> How about `execa.node()`? Is that clear enough?

It's clear that it only spawns the node binary, but it's not clear that it opens an IPC channel. What we could do however, is adding an option to `execa.node()` like I suggested in my previous post, `ipc: true` to make it explicit. So using `execa.node()` would just work the same way as `execa('node')`. Setting `ipc: true` would add the IPC channel.

sindresorhus:
Even if it's not clear from just seeing the method, we could make it clear in the docs about the behavior. I don't see the point of a `.node()` method if I have to set `ipc: true` manually. Then `execa('node')` is just as short.

sindresorhus:
Closing this to clean up open stale PRs.

PR #114, opened 2017-11-09
Document windowsVerbatimArguments option

mickdekkers:
This option is already [part of](https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback) the child_process API (on `execFile`, `fork`, `spawn`, and `spawnSync`), this change just documents it 😄


sindresorhus:
Thanks :)

PR #117, opened 2017-12-19
Detached child process should not kill child - fixes #115

### LINKED ISSUE #115: Detaching child process from parent fails to execute child
We had an issue in Alfy haunting us for a while. Alfred only shows the results when the process exits. In order to detect if a workflow has an update, we spawn a new process which does this for us in the form of [alfred-notifier](https://github.com/SamVerschueren/alfred-notifier).

This is how we did it first

```js
execa('check.js');
```

Then we noticed what was going on and tried to `unref` the child process.

```js
const cp = execa('check.js');
cp.unref();
```

Unfortunately, this is not enough. In order to detach the child process, you must set the `detached` flag to `true` and make sure there is no communication channel between child and parent.

```js
const cp = execa('check.js', {
    detached: true,
    stdio: 'ignore'
});
cp.unref();
```

However, this results in `check.js` not being executed. Using `child_process` directly however, did the trick.

```js
const subProcess = cp.spawn('check.js', [], {
    detached: true,
    stdio: 'ignore'
});
subProcess.unref();
```

-------------------
## How to reproduce

Create two files, `child.js` and `index.js`.

### child.js
```js
#!/usr/bin/env node
'use strict';
const name = process.argv[2];

setTimeout(() => {
	console.log(`Hello World from ${name}`);
}, 3000);

```

### index.js
```js
'use strict';
const path = require('path');
const fs = require('fs');
const cp = require('child_process');

const outCp = fs.openSync('out-cp.log', 'w');
const outExeca = fs.openSync('out-execa.log', 'w');

// Spawn the child process with `child_process`
const subProcess = cp.spawn(path.join(__dirname, 'child.js'), ['child_process'], {
	detached: true,
	stdio: ['ignore', outCp, outCp]
});
subProcess.unref();

// Spawn the child process with `execa`
const subProcess2 = execa(path.join(__dirname, 'child.js'), ['execa'], {
	detached: true,
	stdio: ['ignore', outExeca, outExeca]
});
subProcess2.unref();
```

Run `node index.js`. The process exits immediately and you will notice that two files are being created, `out-cp.log` and `out-execa.log`. After 3 seconds, the content of `out-cp.log` changes but the file `out-execa.log` keeps being empty.

SamVerschueren:
This PR solves #115.

The test however is not 100% representive with the bug because the parent process (the test) is not being killed. This means that even without the fix in `index.js`, the test still passes. I looked into working around this test issue, but couldn't find a way to do it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,7 @@ function handleArgs(cmd, args, opts) {
 	}
 
 	if (opts.detached) {
-		// https://github.com/sindresorhus/execa/issues/115
+		// #115
 		opts.cleanup = false;
 	}
 
--- a/index.js
+++ b/index.js
@@ -54,6 +54,11 @@ function handleArgs(cmd, args, opts) {
 		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
+	if (opts.detached) {
+		// https://github.com/sindresorhus/execa/issues/115
+		opts.cleanup = false;
+	}
+
 	return {
 		cmd: parsed.command,
 		args: parsed.args,
--- a/fixtures/detach
+++ b/fixtures/detach
@@ -1,0 +1,13 @@
+#!/usr/bin/env node
+const fs = require('fs');
+const execa = require('..');
+
+const fd = fs.openSync(process.argv[2], 'w');
+
+const cp = execa('noop', ['foo'], {
+	detached: true,
+	stdio: ['ignore', fd]
+});
+cp.unref();
+
+process.exit();
--- a/test.js
+++ b/test.js
@@ -489,3 +489,18 @@ test('do not buffer when streaming', async t => {
 
 	t.is(result, '....................\n');
 });
+
+test('detach child process', async t => {
+	const file = tempfile('.txt');
+	const fd = fs.openSync(file, 'w');
+
+	const cp = m('noop', ['foo'], {
+		detached: true,
+		stdio: ['ignore', fd]
+	});
+	cp.unref();
+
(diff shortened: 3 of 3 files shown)


sindresorhus:
Maybe move this part:

```js
const file = tempfile('.txt');
	const fd = fs.openSync(file, 'w');
 
	const cp = m('noop', ['foo'], {
		detached: true,
		stdio: ['ignore', fd]
	});
	cp.unref();
```

...into a fixture file that you spawn in the test body?

That way you could kill it and then check if the file exists.

index.js
```diff
@@ -54,6 +54,11 @@ function handleArgs(cmd, args, opts) {
 		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
+	if (opts.detached) {
+		// https://github.com/sindresorhus/execa/issues/115
```
sindresorhus: Let's just do the issue number `#115`. We highlight it with Refined GitHub, so we don't really need the noise of the full URL.

SamVerschueren:
> Maybe move this part into a fixture file that you spawn in the test body?

I tried that, but didn't work either. Apparently I forgot to do `process.exit()` inside the fixture :p, now it works.

Processed all feedback.

sindresorhus:
d8ee7104-273c-11e5-9676-43241f2fc082

SamVerschueren:
Sweet!!

sindresorhus:
This seems to be failing in master.

SamVerschueren:
I'll have a look in the following days.

PR #118, opened 2017-12-19
Turn off echo on Windows - fixes #116

### LINKED ISSUE #116: `stdout` includes prompt + command on Windows
On this line: https://github.com/GoogleChromeLabs/jsvu/blob/905885652848add5f8fe27bb909057fe777974ab/engines/chakra/test.js#L29

`stdout` is equal to:

```js
"\r\nC:\\Users\\jsvu>\"C:\\Users\\jsvu/.jsvu/engines/chakra\\chakra.exe\" \"C:\\Users\\jsvu\\AppData\\Local\\Temp\\3a4c2232c333e7ab7da85331c1dfe4e4\" \r\nHi!"
```

The expected value is only the actual stdout, which is:

```js
"Hi!"
```

In case it matters, this happens after running a batch script `chakra.cmd` with the following contents:

```cmd
"C:\Users\jsvu\.jsvu\engines\chakra\chakra.exe" %*
```

SamVerschueren:
This is a failing test for #116. This PR will also be used to solve the issue. See #116 for the discussion.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 'use strict';
+const path = require('path');
 const childProcess = require('child_process');
 const util = require('util');
 const crossSpawn = require('cross-spawn');
@@ -54,7 +55,7 @@ function handleArgs(cmd, args, opts) {
 		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
-	if (process.platform === 'win32') {
+	if (path.basename(parsed.cmd) === 'cmd.exe') {
 		// #116
 		parsed.args.unshift('/q');
 	}
@@ -192,9 +193,6 @@ module.exports = (cmd, args, opts) => {
 
 	let spawned;
 	try {
-		console.log(parsed.cmd);
-		console.log(parsed.args);
-
 		spawned = childProcess.spawn(parsed.cmd, parsed.args, parsed.opts);
 	} catch (err) {
 		return Promise.reject(err);
--- a/index.js
+++ b/index.js
@@ -54,6 +54,11 @@ function handleArgs(cmd, args, opts) {
 		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
+	if (process.platform === 'win32') {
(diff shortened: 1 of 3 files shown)


SamVerschueren:
@sindresorhus Implemented!

// FYI @mathiasbynens

index.js
```diff
@@ -54,6 +55,11 @@ function handleArgs(cmd, args, opts) {
 		opts.env = npmRunPath.env(Object.assign({}, opts, {cwd: opts.localDir}));
 	}
 
+	if (process.platform === 'win32' && (parsed.command === process.env.comspec || path.basename(parsed.command) === 'cmd.exe')) {
```
SamVerschueren: Is this check ok? Do we need the `parsed.command === process.env.comspec` check or is the basename always `cmd.exe`?

sindresorhus: > is the basename always cmd.exe?

Basename should always be `cmd.exe`. I think only the location could change, hence `comspec`.

index.js
```diff
@@ -58,6 +59,11 @@ function handleArgs(cmd, args, opts) {
 		// #115
 		opts.cleanup = false;
 	}
+  
+  if (process.platform === 'win32' && (parsed.command === process.env.comspec || path.basename(parsed.command) === 'cmd.exe')) {
+		// #116
```
sindresorhus: Weird indentation

index.js
```diff
@@ -58,6 +59,11 @@ function handleArgs(cmd, args, opts) {
 		// #115
 		opts.cleanup = false;
 	}
+  
+  if (process.platform === 'win32' && (parsed.command === process.env.comspec || path.basename(parsed.command) === 'cmd.exe')) {
```
SamVerschueren: Tried to resolve the conflict through mobile but it's a horrible UX... Feel free to fix it, otherwise will do it in the coming days.

sindresorhus: I didn't even know you could fix conflicts on mobile. I've fixed the stuff, no problem.

SamVerschueren: Yes, but the editor is really horrible :)

PR #125, opened 2018-01-27
Avoid relying on Node’s internals

addaleax:
`process.binding()` is not a public API, and should
not be used.

Luckily, Node recently introduced an API that does
exactly what execa needs:
https://github.com/nodejs/node/pull/18186

So use that instead and keep the old path as a fallback.


sindresorhus:
Great. Thank you! :)

PR #129, opened 2018-02-11
Remove exit handler on exit

### LINKED ISSUE #128: execa leaks exit handlers unless .then() is called
Since #95, execa only removes the exit handler that kills the child process when `.then()` is called.

```js

function run() {
  const child = execa('ls');
  // uncomment this to get execa to unregister its exit handlers
  //child.then();
  return new Promise((resolve, reject) => {
    child.on('error', reject);
    child.on('exit', resolve);
  });
}

async function test() {
  for (;;) {
    await run();
    // signal-exit adds it's exit listeners here
    console.log(process.__signal_exit_emitter__.listenerCount('exit'));
  }
}

test().catch(e => console.log(e.stack));
```

This logs

```
1
2
3
4
...
```

Before #95, it would log

```
1
1
1
1
```

also:
Always call `removeExitHandler()` when the spawned process exits, not only if `.then()` is called.

There was already the `cleanupTimeout()` function that was called whenever the spawned process exited, so I repurposed that to clean up everything.

Fixes #128

### DIFF
--- a/test.js
+++ b/test.js
@@ -521,6 +521,6 @@ test('removes exit handler on exit', async t => {
     child.on('exit', resolve);
 	});
 
-	const included = ee.listeners('exit').includes(listener);
+	const included = ee.listeners('exit').indexOf(listener) !== -1;
 	t.false(included);
 });
--- a/test.js
+++ b/test.js
@@ -517,8 +517,8 @@ test('removes exit handler on exit', async t => {
 	const listener = ee.listeners('exit').pop();
 
 	await new Promise((resolve, reject) => {
-    child.on('error', reject);
-    child.on('exit', resolve);
+		child.on('error', reject);
+		child.on('exit', resolve);
 	});
 
 	const included = ee.listeners('exit').indexOf(listener) !== -1;
--- a/test.js
+++ b/test.js
@@ -507,3 +507,20 @@ test('detach child process', async t => {
 
 	t.is(fs.readFileSync(file, 'utf8'), 'foo\n');
 });
+
+// https://github.com/sindresorhus/execa/issues/128
+test('removes exit handler on exit', async t => {
+	// TODO this relies on signal-exit internals
+	const ee = process.__signal_exit_emitter__;
+
+	const child = m('noop');
+	const listener = ee.listeners('exit').pop();
+
+	await new Promise((resolve, reject) => {
+    child.on('error', reject);
+    child.on('exit', resolve);
+	});
+
+	const included = ee.listeners('exit').includes(listener);
+	t.false(included);
+});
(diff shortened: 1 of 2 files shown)


kevva:
Are you able to add a test for this?

test.js
```diff
@@ -507,3 +507,20 @@ test('detach child process', async t => {
...
+// https://github.com/sindresorhus/execa/issues/128
+test('removes exit handler on exit', async t => {
+	// TODO this relies on signal-exit internals
+	const ee = process.__signal_exit_emitter__;
+
+	const child = m('noop');
+	const listener = ee.listeners('exit').pop();
+
+	await new Promise((resolve, reject) => {
+    child.on('error', reject);
```
kevva: Wrong indentation.

test.js
```diff
@@ -507,3 +507,20 @@ test('detach child process', async t => {
...
+    child.on('exit', resolve);
+	});
+
+	const included = ee.listeners('exit').indexOf(listener) !== -1;
+	t.false(included);
```
kevva: `t.false(ee.listeners('exit').includes(listener));`

also: `includes` is not supported by Node.js 4.

Putting the expression in `t.false` produces a long and unhelpful message on test failure:

```
/Users/rberdeen/work/execa/test.js:524

   523:
   524:   t.false(ee.listeners('exit').indexOf(listener) !== -1);
   525: });

  Value is not `false`:

  true

  ee.listeners('exit').indexOf(listener) !== -1
  => true

  -1
  => -1

  ee.listeners('exit').indexOf(listener)
  => 15

  listener
  => Function {}

  ee.listeners('exit')
  => [
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
    Function {},
  ]

  ee
  => EventEmitter {
    _events: {
      afterexit: Function {},
      exit: [
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
        Function {},
      ],
    },
    _eventsCount: 2,
    _maxListeners: Infinity,
    count: 2,
    domain: null,
    emitted: {},
    infinite: true,
```

kevva: Oh, right. `Array.includes` isn't handled by `babel` (in AVA).

sindresorhus:
Thank you :)

PR #133, opened 2018-04-06
Add `buffer` option to be able to disable buffering

### LINKED ISSUE #127: Explicitly disabling buffering when .then() is used
execa's promise based interface is really useful, especially when used with async/await.

However, because calling `.then()` on the child enables buffering of the child's output, it isn't practical to use the promise-based API while also using the output streams.

It would be nice to be able to use the output streams while also taking advantage of the promise API. #84 suggested an explicit option like `{buffer: false}` *or* the current behavior of enabling buffering when `.then()` is called. I think it would be useful to support both, where `{buffer: false}` will disable buffering even if `.then()` is called.

This works fine, and creates a 20mb file:
```js
function test() {
  const child = execa('dd', ['if=/dev/zero', 'count=20480', 'bs=1024']);
  child.stdout.pipe(fs.createWriteStream('zeros'));
}
```

This throws `Error: stdout maxBuffer exceeded`, because `.then()` enables buffering:
```js
async function test() {
  const child = execa('dd', ['if=/dev/zero', 'count=20480', 'bs=1024']);
  child.stdout.pipe(fs.createWriteStream('zeros'));
  await child;
}
```

**Proposed solution: this should create a 20mb file without filling a buffer:**

```js
async function test() {
  const child = execa('dd', ['if=/dev/zero', 'count=20480', 'bs=1024'], {buffer: false});
  child.stdout.pipe(fs.createWriteStream('zeros'));
  await child;
}
```

spalger:
Fixes #127 

The issue describes adding a `buffer: false` option, which seemed to be in conflict with the `maxBuffer` option that's already supported, so rather than adding a new option this PR disables buffering of stdio when `maxBuffer: null` is used.

### DIFF
--- a/test.js
+++ b/test.js
@@ -239,8 +239,8 @@ test('maxBuffer affects stderr', async t => {
 	await t.notThrows(m('max-buffer', ['stderr', '12'], {maxBuffer: 12}));
 });
 
-test('do not buffer stdout when `maxBuffer` set to `null`', async t => {
-	const promise = m('max-buffer', ['stdout', '10'], {maxBuffer: null});
+test('do not buffer stdout when `buffer` set to `false`', async t => {
+	const promise = m('max-buffer', ['stdout', '10'], {buffer: false});
 	const [result, stdout] = await Promise.all([
 		promise,
 		getStream(promise.stdout)
@@ -250,8 +250,8 @@ test('do not buffer stdout when `maxBuffer` set to `null`', async t => {
 	t.is(stdout, '.........\n');
 });
 
-test('do not buffer stderr when `maxBuffer` set to `null`', async t => {
-	const promise = m('max-buffer', ['stderr', '10'], {maxBuffer: null});
+test('do not buffer stderr when `buffer` set to `false`', async t => {
+	const promise = m('max-buffer', ['stderr', '10'], {buffer: false});
 	const [result, stderr] = await Promise.all([
 		promise,
 		getStream(promise.stderr)
--- a/test.js
+++ b/test.js
@@ -380,7 +380,7 @@ test('err.code is 3', code, 3);
 test('err.code is 4', code, 4);
 
 test('timeout will kill the process early', async t => {
-	const err = await t.throws(m('delay', ['3000', '0'], {timeout: 1500}));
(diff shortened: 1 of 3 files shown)


sindresorhus:
`maxBuffer: null` could also mean that it should buffer indefinitely. It's ambiguous. I prefer `buffer: false` as it's explicit.

sindresorhus:
@spalger Still interested in finishing this?

spalger:
@sindresorhus yep, sorry, got lost in the pile. Should be good to go now

test.js
```diff
@@ -358,7 +380,7 @@ test('err.code is 3', code, 3);
 test('err.code is 4', code, 4);
 
 test('timeout will kill the process early', async t => {
-	const err = await t.throws(m('delay', ['3000', '0'], {timeout: 1500}));
+	const err = await t.throws(m('delay', ['5000', '0'], {timeout: 1500}));
```
spalger: Did this in an effort to fix the AppVeyor failure here: https://ci.appveyor.com/project/sindresorhus/execa/build/1.0.337

Not sure if this is a failure others see this type of failure regularly, but it passes locally and shouldn't be effected by my changes, so I'm not sure what else I should do about it.

test.js
```diff
@@ -358,7 +380,7 @@ test('err.code is 3', code, 3);
 test('err.code is 4', code, 4);
 
 test('timeout will kill the process early', async t => {
-	const err = await t.throws(m('delay', ['3000', '0'], {timeout: 1500}));
+	const err = await t.throws(m('delay', ['60000', '0'], {timeout: 1500}));
```
spalger: Did this in an effort to fix the AppVeyor failure here: https://ci.appveyor.com/project/sindresorhus/execa/build/1.0.337

Not sure if this is a failure others see this type of failure regularly, but it passes locally and shouldn't be effected by my changes, so I'm not sure what else I should do about it.

sindresorhus: AppVeyor is known for being really slow and annoying. The usual fix is what you've done and just increase the timeouts a *lot*.

sindresorhus:
Thank you 🙌

PR #156, opened 2018-11-12
added manually destroy spawn

Gcaufy:


sindresorhus:
Can you add a description to the pull request? And a test to prove that your fix does what you say.

Gcaufy:
I found this is not the solution.
please check the test case here: https://github.com/sindresorhus/execa/pull/159

PR #159, opened 2018-11-16
added a failure test case

Gcaufy:
when the network in down, npm info will no work.
but when I add timeout option to execa, it does not kill the npm info process.

here is a test case of it.


Gcaufy:
any progress here?

sindresorhus:
I checked out this PR locally and the test passes.

Gcaufy:
should not work, I'm still stuck here.

PR #161, opened 2018-11-26
ci: test on Linux and Windows

DanielRuf:
We can also test on Windows too now.


sindresorhus:
The Windows support is just an early release in Travis and I'm not really interested in dealing with any problems because of that.

DanielRuf:
I know. Why not keep it open for further testing? Also tests are all green.

Why so negative / against this?

PR #162, opened 2018-11-26
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Also test against current stable version (fail early).


sindresorhus:
The latest non-LTS version of Node.js breaks the ecosystem often and I'm not interested in random breakage when I run the CI.

DanielRuf:
Ok, but what about `allow_failures` then?

PR #171, opened 2019-01-20
Add `.all` property with interleaved stdout and stderr

### LINKED ISSUE #1: Return an `.all` property with both stdout and stderr intermixed

tomsotte:
Added test cases for `result.all` intermixed output.

Added the fixture `noop-132` which should output the sequence `123` (where `13` comes from stdout and `3` comes from stderr). Due to the unpredictable nature of the stdout/stderr it outputs `132` as the most probable case.
I've tried, as listed on `noop-132`, a different number of technique to make it predictable.

On `execa.sync`, as it's using `spawnSync` internally, I couldn't get the stream and merge them.

As such the output of result.all is just a concatenation of stdout + stderr.
This makes the result.all in different use case (ex. async vs. sync) unreliable for string comparison, but at least useful for logging or checking on the output.

**Edit**: Removed `result.all` on `execa.sync`, as it's just a concatenated stdout and stderr output, not very useful. Added `execa.all()` similar to `execa.stdout()`. Added `all` in errors (only for async). Added documentation.

Closes #1

### DIFF
--- a/fixtures/noop-132
+++ b/fixtures/noop-132
@@ -13,4 +13,4 @@ process.stderr.write('3');
 
 setTimeout(() => {
 	process.stdout.write('2');
-}, 5000);
+}, 1000);
--- a/fixtures/noop-132
+++ b/fixtures/noop-132
@@ -9,6 +9,8 @@
 // - console.log/error
 
 process.stdout.write('1');
-process.stdout.write('3');
-// Ideally, to demonstrate the interleaving, the next line should go before '3'
-process.stderr.write('2');
+process.stderr.write('3');
+
+setTimeout(() => {
+	process.stdout.write('2');
+}, 5000);
--- a/fixtures/noop-132
+++ b/fixtures/noop-132
@@ -1,0 +1,14 @@
+#!/usr/bin/env node
+'use strict';
+// Interleaving the output of stdout and stderr is unpredictable and can't be
+// expressed as a test. What I've tried so far to make it predictable:
+// - process.stdout._handle.setBlocking(true)
+// - process.stdout.once('drain', () => process.stderr.write('2'));
+// - process.stdout.write('1', () => process.stderr.write('2'));
+// - await delay() between calls; it works but with varying ms, based on load
+// - console.log/error
+
+process.stdout.write('1');
+process.stdout.write('3');
+// Ideally, to demonstrate the interleaving, the next line should go before '3'
+process.stderr.write('2');
--- a/fixtures/noop-132
+++ b/fixtures/noop-132
@@ -1,12 +1,5 @@
 #!/usr/bin/env node
 'use strict';
-// Interleaving the output of stdout and stderr is unpredictable and can't be
-// expressed as a test. What I've tried so far to make it predictable:
-// - process.stdout._handle.setBlocking(true)
-// - process.stdout.once('drain', () => process.stderr.write('2'));
-// - process.stdout.write('1', () => process.stderr.write('2'));
-// - await delay() between calls; it works but with varying ms, based on load
-// - console.log/error
 
 process.stdout.write('1');
 process.stderr.write('3');
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const path = require('path');
+const os = require('os');
 const childProcess = require('child_process');
 const crossSpawn = require('cross-spawn');
 const stripFinalNewline = require('strip-final-newline');
@@ -15,43 +16,37 @@ const stdio = require('./lib/stdio');
 const TEN_MEGABYTES = 1000 * 1000 * 10;
 
 function handleArgs(command, args, options) {
-	let parsed;
+	const parsed = crossSpawn._parse(command, args, options);
+	command = parsed.command;
+	args = parsed.args;
+	options = parsed.options;
 
-	options = Object.assign({
-		extendEnv: true,
-		env: {}
-	}, options);
-
-	if (options.extendEnv) {
-		options.env = Object.assign({}, process.env, options.env);
-	}
-
-	if (options.__winShell === true) {
-		delete options.__winShell;
-		parsed = {
-			command,
-			args,
-			options,
-			file: command,
-			original: {
-				command,
-				args
-			}
-		};
-	} else {
-		parsed = crossSpawn._parse(command, args, options);
-	}
-
-	options = Object.assign({
+	options = {
 		maxBuffer: TEN_MEGABYTES,
 		buffer: true,
 		stripFinalNewline: true,
 		preferLocal: true,
-		localDir: parsed.options.cwd || process.cwd(),
+		localDir: options.cwd || process.cwd(),
 		encoding: 'utf8',
 		reject: true,
-		cleanup: true
-	}, parsed.options, {windowsHide: true});
+		cleanup: true,
+		...options,
+		windowsHide: true
+	};
+
+	if (options.extendEnv !== false) {
+		options.env = {
+			...process.env,
+			...options.env
+		};
+	}
+
+	if (options.preferLocal) {
+		options.env = npmRunPath.env({
+			...options,
+			cwd: options.localDir
+		});
+	}
 
 	// TODO: Remove in the next major release
 	if (options.stripEof === false) {
@@ -60,26 +55,17 @@ function handleArgs(command, args, options) {
 
 	options.stdio = stdio(options);
 
-	if (options.preferLocal) {
-		options.env = npmRunPath.env(Object.assign({}, options, {cwd: options.localDir}));
-	}
-
 	if (options.detached) {
 		// #115
 		options.cleanup = false;
 	}
 
-	if (process.platform === 'win32' && path.basename(parsed.command) === 'cmd.exe') {
+	if (process.platform === 'win32' && path.basename(command) === 'cmd.exe') {
 		// #116
-		parsed.args.unshift('/q');
+		args.unshift('/q');
 	}
 
-	return {
-		command: parsed.command,
-		args: parsed.args,
-		options,
-		parsed
-	};
+	return {command, args, options, parsed};
 }
 
 function handleInput(spawned, input) {
@@ -103,24 +89,7 @@ function handleOutput(options, value) {
 }
 
 function handleShell(fn, command, options) {
-	let file = '/bin/sh';
-	let args = ['-c', command];
-
-	options = Object.assign({}, options);
-
-	if (process.platform === 'win32') {
-		options.__winShell = true;
-		file = process.env.comspec || 'cmd.exe';
-		args = ['/s', '/c', `"${command}"`];
-		options.windowsVerbatimArguments = true;
-	}
-
-	if (options.shell) {
-		file = options.shell;
-		delete options.shell;
-	}
-
-	return fn(file, args, options);
+	return fn(command, {...options, shell: true});
 }
 
 function makeAllStream(spawned) {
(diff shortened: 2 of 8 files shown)


sindresorhus:
@ehmicky Would you be able to help review this?

index.js
```diff
@@ -122,6 +123,24 @@ function handleShell(fn, command, options) {
 	return fn(file, args, options);
 }
 
+function makeAllStream(spawned) {
+	const mixed = mergeStream();
```
ehmicky: Maybe this should be initialized after the next three lines, so that it's not initialized if `null` is returned.

index.js
```diff
@@ -339,6 +367,10 @@ module.exports.sync = (command, args, options) => {
 	const result = childProcess.spawnSync(parsed.command, parsed.args, parsed.options);
 	result.code = result.status;
 
+	// `spawnSync` doesn't expose the stdout/stderr before terminating, which means
```
ehmicky: I think this would be better to only provide `result.all` for async then, because just concatenating the streams might not be good enough.

fixtures/noop-132
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
```
ehmicky: This does not test interleaving because `stderr` is not in-between two `stdout` calls.

fixtures/noop-132
```diff
@@ -0,0 +1,14 @@
...
+'use strict';
+// Interleaving the output of stdout and stderr is unpredictable and can't be
+// expressed as a test. What I've tried so far to make it predictable:
+// - process.stdout._handle.setBlocking(true)
+// - process.stdout.once('drain', () => process.stderr.write('2'));
+// - process.stdout.write('1', () => process.stderr.write('2'));
+// - await delay() between calls; it works but with varying ms, based on load
```
ehmicky: I think using delays of like 100ms might be good enough to make it somewhat predictable?

ehmicky:
Done! Also I think documentation would need to be added in the `README`.

index.js
sindresorhus: I agree. The readme should also document why it’s not available on the synchronous methods.

fixtures/noop-132
sindresorhus: 👍

index.js
sindresorhus: 👍

index.js
tomsotte: You're right, I've removed it as it does not really give any advantage over the user concatenating the streams at the end of the process. I've written a bit of doc for `execa.sync`, let me know how it can be completed or if it's enough.

fixtures/noop-132
tomsotte: It was my mistake, I should have left it in-between and let the test fail. As I said in the comments I was testing various way to implement the test to be correct but I couldn't find one for all (more on this on the next conversation).

fixtures/noop-132
tomsotte: Yes, the only way I found to get the test correct is to put a delay.
I'm testing right now on my machine and sometimes it passes the test with `3000ms` delay, but sometimes not. Which makes it unpredictable on other machines. I believe it's based on the current load on the system.
With `6000ms` delay, on my machine, it seems it always passes the test but slows down the whole testing. Is it enough as a solution? Have you any other idea on how it can be made predictable?

ehmicky: Would the other solutions you're suggesting, e.g. using the `stream.write()` callback, work instead?

ehmicky: Also I think it might be a good idea to do a more complex interleaving like stdout > stderr > stderr > stdout > stderr > stdout

fixtures/noop-132
ehmicky: Also I think it might be a good idea to do a more complex interleaving like stdout > stderr > stderr > stdout > stderr > stdout

index.js
ehmicky: On the documentation side maybe the following would be good?
  - Put interleaving support in the `Why` section because that's a nice differentiating feature
  - Update examples, as the `result` return values will now be different
  - Document `execa.all()` method like `execa.stdout`
  - Document `options.all` like `options.stdout`

fixtures/noop-132
tomsotte: None of those worked for me. Only an high enough delay worked so far.
I've tried nesting callbacks on `stream.write`.
I've just re-tried some of what I've thought could work [here](https://gist.github.com/tomsotte/1e5fb674d94dba1cd7af1314af658c9d) (commenting/decommenting chunks of code to get the sequence).
Most of them returns on my console the correct sequence, interleaving stdout and stderr, '132', but when trying to test it they fail, returning `123`.
I may be missing something in all of this, but at the moment I've no other ideas to try.

sindresorhus:

fixtures/noop-132
ehmicky: Did you try creating streams in the test, then passing those down to the executable file using the `stdout` and `stderr` options?

tomsotte: Any other stream type passed down results to `null` for `stdout`, `stderr` and `all`. I tried with fs.createFileStream, net.Socket, 'inherit'... only the default 'pipe' works.
On ava it works, with a delay of `6000ms` on my machine before the last `stdout.write` call. That's the only way I can reproduce the result on `all` being interleaved.

fixtures/noop-132
tomsotte: I don't think it needs to be more complex, if it works just once we should expect to interleave the rest of the outputs. Otherwise we know it's just concatenating the output strings.
From what I understand by the tests I'm doing is that we cannot guarantee the order of the interleaving.

tomsotte:
Ok, now I fixed it. I thought it was enough to fix them locally and then commit to remote.

fixtures/noop-132
ehmicky: Other possible idea: using `test.serial()` to ensure no other tests are fired in parallel? This might be an issue due to parallelism. If it's not, this might still allow you to decrease the delay.

fixtures/noop-132
ehmicky: What do you think @sindresorhus?

fixtures/noop-132
tomsotte: Good find, `test.serial()` may be the answer. I've lowered to 1000ms the delay for `noop-132` and it seems to pass test runs consistently. Let's see if the CI agree as well.

tomsotte:
Shall I add an `execa.all()` in the same vein as `execa.stdout()`?

test.js
```diff
@@ -46,6 +46,14 @@ test('execa.stderr()', async t => {
 	t.is(stderr, 'foo');
 });
 
+test.serial('result.all shows both `stdout` and `stderr` intermixed', async t => {
+	const result = await execa('noop-132');
+	// Due to the async nature of process.stdout/stderr on POSIX, this test
+	// is very unpredictable, although it should interleave the streams
+	// https://nodejs.org/api/process.html#process_a_note_on_process_i_os
```
ehmicky: There is a typo in the link: https://nodejs.org/api/process.html#process_a_note_on_process_i_o

fixtures/noop-132
ehmicky: Seems like it works in CI, problem solved!

ehmicky:
@tomsotte this seems like a good idea :) 

Thanks for your work on this, it's a very useful feature!

index.js
tomsotte: I've not implemented an `options.all`. Accepting a stream as an option should be easy to handle, by piping to the `all` stream. To accept other types as [`options.stdio` in child_process](https://nodejs.org/api/child_process.html#child_process_options_stdio) does will take longer and I believe I can't guarantee enough correctness for all the other use cases.

I've started looking references about those in [child_process stdio](https://nodejs.org/api/child_process.html#child_process_options_stdio) and [nodejs/node/lib/internal/child_process.js#_validateStdio](https://github.com/nodejs/node/blob/3c277d7e92adf1464e41d27677fcdc05fec41a2b/lib/internal/child_process.js#L861)

ehmicky: My mistake. I agree `options.all` does not sound very important (as opposed to `result.all`).

tomsotte:
Should errors have the `all` output? At the time I did not added as I thought `stdout` and `stderr` would have been enough, but it might be helpful or at least it could be there for consistency with the normal result. What do you think?

ehmicky:
Yes nice catch :+1:

index.js
```diff
@@ -332,6 +364,11 @@ module.exports.stderr = async (...args) => {
 	return stderr;
 };
 
+module.exports.all = async (...args) => {
+	const {all} = await module.exports(...args);
+	return all;
+};
```
sindresorhus: I don't want a shortcut method for `all`. The existing ones only exist because we didn't have destructuring before.

fixtures/noop-132
```diff
@@ -0,0 +1,16 @@
...
+// - console.log/error
```
sindresorhus: These comments need to be removed.

index.js
```diff
@@ -144,6 +163,10 @@ function makeError(result, options) {
 	error.cmd = joinedCommand;
 	error.timedOut = Boolean(timedOut);
 
+	if ('all' in result) {
```
sindresorhus: Why is this guarded? It should not be treated differently than `stdout` and `stderr`, unless I'm missing something?

readme.md
```diff
@@ -100,7 +102,9 @@ Execute a file.
 
 Think of this as a mix of `child_process.execFile` and `child_process.spawn`.
 
-Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess), which is enhanced to also be a `Promise` for a result `Object` with `stdout` and `stderr` properties.
+Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) which is enhanced to be a `Promise`.
+It exposes an additional `all` stream, with both `stdout` and `stderr` outputs combined.
```
sindresorhus: `combined` => `interleaved`

readme.md
```diff
@@ -120,9 +128,9 @@ The `child_process` instance is enhanced to also be promise for a result object
 
 ### execa.sync(file, [arguments], [options])
 
-Execute a file synchronously.
+Execute a file synchronously.
```
sindresorhus: ^

readme.md
```diff
@@ -120,9 +128,9 @@ The `child_process` instance is enhanced to also be promise for a result object
 
 ### execa.sync(file, [arguments], [options])
 
-Execute a file synchronously.
+Execute a file synchronously. 
 
-Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
+Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options). It's not possible to obtain a true interleaved `all` stream as `execa()` does, because `child_process.spawnSync` does not expose `stdout` and `stderr` to be combined until termination.
```
sindresorhus: I think the last sentence could be written better.

test.js
```diff
@@ -46,10 +46,24 @@ test('execa.stderr()', async t => {
...
+test.serial('execa.all()', async t => {
+	const all = await execa.all('noop-132');
+	t.is(all, '132');
+});
+
+test.serial('result.all shows both `stdout` and `stderr` intermixed', async t => {
+	const result = await execa('noop-132');
+	// Due to the async nature of process.stdout/stderr on POSIX, this test
+	// is very unpredictable, although it should interleave the streams.
+	// https://nodejs.org/api/process.html#process_a_note_on_process_i_o
```
sindresorhus: Since it's not unstable anymore, this comment can be removed.

readme.md
tomsotte: Could this be better?
> It does not expose the additional `all` interleaved stream as `execa()` does because the underlaying sync implementation cannot guarantee a true interleaved output from `stdout` and `stderr`.

sindresorhus: > It does not have the `.all` property that `execa()` has because the underlying synchronous implementation only returns `stdout` and `stdout` at the end of the execution, so they cannot be interleaved.

index.js
tomsotte: I did this because the `makeError` function is used by `execa.sync()` as well, which does not have the `all` stream in `result.all`. This guard was meant to prevent adding `all` property when it's not available on the `result` object at all.
I used the `in`, which is stricter, so that it guards from using `in` and `Object#hasOwnProperty` on the error object to, for example, check if the error came from the async or sync execa. But I don't have any other use cases in which it can be useful.
What I meant in the code was to highlight that it's not part of the original result and that it's more of an addition. This guard may be too much.

ehmicky: This makes sense to me.

sindresorhus:
Thanks for contributing, @tomsotte :)

sindresorhus:
@tomsotte You need to submit the PR URL to IssueHunt to claim the bounty.

tomsotte:
Thanks to you both @sindresorhus and @ehmicky for patience and guiding me through the review :)

ehmicky:
@tomsotte thanks for this nice addition!

PR #172, opened 2019-01-22
Return a .all property in the result

### LINKED ISSUE #1: Return an `.all` property with both stdout and stderr intermixed

lunaticmonk:
Adds a .all property along with the result.

Closes #1


sindresorhus:
I appreciate the PR, but this is not the correct way to implement it. It doesn't interweave the stdout and stderr streams.

sindresorhus:
Duplicate of #171

PR #178, opened 2019-02-28
Upgrade ava to 1.2.1

ehmicky:
The `0.25.0` version of `ava` is reported by `npm audit` as having a vulnerable deep dependency, so this also fixes that.

### DIFF
--- a/test.js
+++ b/test.js
@@ -12,8 +12,8 @@ import m from '.';
 process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
 process.env.FOO = 'foo';
 
-const NO_NEWLINES_REGEXP = /^[^\n]*$/
-const STDERR_STDOUT_REGEXP = /stderr[^]*stdout/
+const NO_NEWLINES_REGEXP = /^[^\n]*$/;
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -42,41 +42,41 @@ test('execa.stderr()', async t => {
...
 	t.is(typeof err.stdout, 'string');
 	t.is(typeof err.stderr, 'string');
 });
 
 test('include stdout and stderr in errors for improved debugging', async t => {
-	const err = await t.throws(m('fixtures/error-message.js'));
+	const err = await t.throwsAsync(m('fixtures/error-message.js'));
 	t.regex(err.message, /stdout/);
 	t.regex(err.message, /stderr/);
 	t.is(err.code, 1);
```
sindresorhus: Can you also use the new error matching format? https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expected-message Then you don't need to return the error.

ehmicky: Sure thing, will do. In some of the tests this can't be used though, e.g. when `error.stdout|stderr` is being tested, I don't think the new assertion format supports that.

ehmicky: Done

sindresorhus:
Thanks :)

Issue #179, opened 2019-02-28
[Feature proposal] Distinguishing between error strings and error numbers


ehmicky:
At the moment `result.code` is either a string like `EACCES`, an integer like `13` or `undefined`.

It would be nice to split this property into two: `code` for the string and `errno` for the integer. 

This would be a breaking change if the name `code` is re-used. Otherwise we can pick another property name.

If this sounds good, I will submit a PR.

sindresorhus:
Ah yeah... I should not have reused `.code` for something different.

I don't want to do a breaking change *just* for this, so I suggest we add a `.exitCode` property. I prefer that over `.errno`. [`.errno` in Node.js](https://nodejs.org/api/errors.html#errors_error_errno) can be both a string and a number.

ehmicky:
Sounds good. 

So `code` is currently either string or integer. Should `exitCode` be string? Which property name should be for integer?

sindresorhus:
`.exitCode` should be an integer. The exit code the process excited with.

ehmicky:
Since `code` can be either a string or an integer, should there a third property that can only be a string?

sindresorhus:
What would it contain? Example?

ehmicky:
For example `code` currently can be either a string (`"EACCESS"`), an integer (`2`) or `undefined`.
`exitCode` will only be an integer (`2`) or `undefined`.
A third property could be only a string (`"EACCESS"`) or `undefined`.

sindresorhus:
> A third property could be only a string ("EACCES") or undefined.

Do you have any use-case for this? You could mostly compare `.code` to something anyway, in which the type doesn't matter.

ehmicky:
The use case would be if a user wants to print the signal name but not number. Some users might consider that more user-friendly. Also signal names are cross-platform, whereas signal numbers aren't.

sindresorhus:
Alright, any suggestion on what to name this string-only property?

ehmicky:
`exitName`?

sindresorhus:
I think it should be `exitCode` for the number, since that's what it's called outside the Node.js ecosystem too. As for the string, maybe `exitCodeName` or `exitCodeId`?

ehmicky:
`code` (integer or string), `exitCode` (integer), `exitCodeName` (string) sounds good, I can open a PR.

Should `code` be marked as deprecated and removed in next major release or not?

sindresorhus:
[...]

PR #180, opened 2019-02-28
Improve returned `error.message`

### LINKED ISSUE #177: [Feature proposal] Improved error messages
The current code has [some logic](https://github.com/sindresorhus/execa/blob/master/index.js#L155) aimed at providing better error messages. It would be nice to add the following error messages:
  - `Command timed out after 5 seconds: ...` if `timedOut` is `true`.
  - `Command was killed with SIGTERM` if `signal` is not `null`.
  - `Command failed with exit code 2 (ENOENT)` otherwise.

The values (e.g. `SIGTERM`) in the messages above are for example purpose.

I can provide with a PR if this feature proposal is approved.

ehmicky:
Fix #177 

This improves returned `error.message` by adding information about potential timeouts, signals and exit codes.

This also fixes few issues with `result.code`:
  - it was an integer instead of a string most of time because of the line `error.code = code < 0 ? errname(code) : code`. `code` comes from `childProcess.on('exit')` which is a positive integer most of the times. The fixed code avoids breaking backward compatibility by using the integer first if it's available.
  - `error.errno` was not used. There can be instances where `process.exit()` is not called but `error.errno` is available, which can be used as error `code` instead.
  - `os.constants.errno` is used to retrieve the code string when possible.

This also fixes `stderr` and `stdout` being concatenated with a newline between them when `options.stdio` was not a string.

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,21 +122,21 @@ function getStream(process, stream, {encoding, buffer, maxBuffer}) {
 
 function makeError(result, options) {
 	const {stdout, stderr, code, signal} = result;
-	let {error} = result
-	const {joinedCommand, timedOut, parsed:{options:{timeout}}} = options;
+	let {error} = result;
+	const {joinedCommand, timedOut, parsed: {options: {timeout}}} = options;
 
-	const [codeString, codeNumber] = getCode(result, code)
+	const [codeString, codeNumber] = getCode(result, code);
 
 	if (!(error instanceof Error)) {
-		const message = [joinedCommand, stderr, stdout].filter(Boolean).join('\n')
+		const message = [joinedCommand, stderr, stdout].filter(Boolean).join('\n');
 		error = new Error(message);
 	}
 
-	const prefix = getErrorPrefix({ timedOut, timeout, signal, codeString, codeNumber })
-	error.message = `Command ${prefix}: ${error.message}`
-	console.log(error.message)
+	const prefix = getErrorPrefix({timedOut, timeout, signal, codeString, codeNumber});
+	error.message = `Command ${prefix}: ${error.message}`;
+	console.log(error.message);
 
-	error.code = codeNumber || codeString
+	error.code = codeNumber || codeString;
 	error.stdout = stdout;
 	error.stderr = stderr;
 	error.failed = true;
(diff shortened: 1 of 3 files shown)


sindresorhus:
Since technically changing error messages is a breaking change and we have some other potential breaking changes coming up now, I guess we can do a major version after these changes, so we can make breaking changes now.

sindresorhus:
> The fixed code avoids breaking backward compatibility by using the integer first if it's available.

We don't have to do this anymore. We just have to make sure we make this clear in the release notes (Note to self).

sindresorhus:
> This also fixes few issues with result.code:

Shouldn't these changes we done in a separate PR when fixing #179?

ehmicky:
The changes suggested in #179 can be done in a separate PR IMHO. Note that this PR makes adding those changes much easier.

The backward compatibility fix I mentioned is on `error.code`. According to #179 it seems like `error.code` might be removed in next major version anyway, so maybe it's ok to keep this as is in the PR? Or would rather me remove that? To be precise we're talking about [this line](https://github.com/sindresorhus/execa/blob/73ab4b8367cd268dcf465c90eab64fe69e0aca3f/index.js#L171).

sindresorhus:
> According to #179 it seems like error.code might be removed in next major version anyway, so maybe it's ok to keep this as is in the PR?

Sure

sindresorhus:
This looks good 👌

Would you be interested in joining the project as a maintainer? No worries if not. Just thought I'd ask as you've done some good PRs 😃

sindresorhus:
🦄

ehmicky:
Sure! :+1: 

My main motivation was that I have two projects closely related to `execa`:
  - a [Portable Node.js guide](https://github.com/ehmicky/portable-node-guide) which relies on `execa` for cross-platform commands
  - a `gulp-execa` project not released yet (but almost finished coding). There are about 15 Gulp plugins to trigger shell commands in Gulp, but none of them are cross-platform. All of them use `child_process.spawn()` as is, except `gulp-shell` [does some work](https://github.com/sun-zheng-an/gulp-shell/blob/master/index.ts#L37) for cross-platform support, but it's quite limited. Also none of them use `execa` under the hood which I think is the only really maintained Node.js library for its purpose.

sindresorhus:
@ehmicky Awesome. Welcome to the project!

sindresorhus:
Even if you have commit access, it's usually best to do a PR for changes, to get feedback.

ehmicky:
@sindresorhus sounds good. Is it ok to merge other developers PRs if the PR is a straightforward change, or do you prefer always giving it a thumbs up?

sindresorhus:
> sounds good. Is it ok to merge other developers PRs if the PR is a straightforward change, or do you prefer always giving it a thumbs up?

I prefer PRs to get multiple reviews. One reviewer might miss something another reviewer might see. I will wait for you to review PRs too as I'm sure you will catch things I might miss too.

sindresorhus:
I could also use your thoughts/comments on the existing issues :)

Issue #126, opened 2018-02-04
Use the `cross-spawn` shell option


sindresorhus:
See: https://github.com/sindresorhus/execa/pull/124

> - There's custom code that handles shell and shellSync but you could use `options.shell` from cross-spawn, which basically does what it's currently being done but is a bit more robust. See: https://github.com/moxystudio/node-cross-spawn/blob/master/lib/parse.js#L66
> - Related to the point above, any code handling `__winShell` could be removed

sindresorhus:
Fixed by #181

PR #181, opened 2019-03-01
Improve `execa.shell()` and remove dead code

ehmicky:
The logic in `execa.shell()` can be replaced by simply using the `shell: true` option.

I looked it through and I think the reason this logic was here in the first place was for older Node.js versions.

`handleShell()` does exactly what [the `shell: true` option](https://github.com/nodejs/node/blob/master/lib/child_process.js#L474) does in core Node.js, except the last one is a little better:
   - it supports Android (which uses a different location for `/bin/sh`)
   - it supports the `%COMSPEC%` environment variable being something else than `cmd.exe`. The current `execa` code adds `cmd.exe`-specific flags even when the chosen shell is not `cmd.exe`.
   - it adds the `/d` flag with `cmd.exe` which disables registry AutoRun commands

The `handleShell()` function also sets a `__winShell` flag which [bypasses](https://github.com/sindresorhus/execa/blob/master/index.js#L28) `cross-spawn._parse()`. But `cross-spawn._parse()` with `shell: true` is [effectively a noop with Node.js >= 6](https://github.com/moxystudio/node-cross-spawn/blob/master/lib/parse.js#L71) which is [what this library supports](https://github.com/sindresorhus/execa/blob/master/package.json#L12).

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,8 +14,6 @@ const stdio = require('./lib/stdio');
 const TEN_MEGABYTES = 1000 * 1000 * 10;
 
 function handleArgs(command, args, options) {
-	let parsed;
-
 	options = Object.assign({
 		extendEnv: true,
 		env: {}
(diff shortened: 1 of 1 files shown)


sindresorhus:
Yay! I love being able to remove code.

I guess this fixes https://github.com/sindresorhus/execa/issues/126?

ehmicky:
Yes #126 can be closed :+1:

PR #182, opened 2019-03-01
Support single string input without using the `shell` option

### LINKED ISSUE #176: [Feature proposal] Single string input without using the `shell`
I can submit a PR for this feature proposal if approved.

I think many users might use `execa.shell()` or the `shell: true` option as a way to pass arguments as a single string which feels more intuitive than using one string + one array:

```js
execa.shell('eslint *.js *.yml --ignore-path=.gitignore --fix --cache --format=codeframe --max-warnings=0')
execa('eslint', ['*.js', '*.yml', '--ignore-path=.gitignore', '--fix', '--cache', '--format=codeframe', '--max-warnings=0'])
```

However those two calls use very different mechanism to execute the command, as the first one relies on system calls and the second one uses a shell interpreter. Using a shell interpreter is:
  - less secure: it allows for [command injection](https://www.owasp.org/index.php/Command_Injection) by passing arguments like `$(rm -rf /)` or `&& rm -rf /`
  - less cross-platform: it encourages using [shell-specific features](https://twitter.com/ehmicky/status/1098980276787732480) such as globbing, single quote escaping or even semicolons which won't work on `cmd.exe`.
  - slower as it goes through an extra step (the shell interpreter)

Using the shell interpreter can almost always be avoided as almost all shell features can be emulated in Node.js:
  - globbing can be done with libraries like `minimatch`. Directory recursion can also be used.
  - piping, subcommands, chaining and streams redirection can be done with the `std*` and `input` options.
  - escaping/quoting becomes unnecessary.
  - variable expansion can be done with template strings. This includes `process.env`.
  - passing environment variables can be done with the `env` option.
  - built-in Bash commands logic can be performed in JavaScript instead.
  - background processes can be done with the `detached` option.

In order to encourage more secure, cross-platform and faster shell execution, I suggest the following: 
  - when the `shell` option is `false`, the first argument can be either `command` (current behavior) or `command ...args` (new behavior)
  - `command ...args` is a string delimited by spaces.
  - spaces can be escaped with backslashes: `command escaped\ space`. Anything else does not need escaping, just like the `args` array.

What do you think? Again if approved I can submit a PR.

ehmicky:
Fix #176 

This allows passing command+arguments as a single string with `execa()` without using the `shell` option:

```js
execa('eslint *.js *.yml --ignore-path=.gitignore --fix --cache --format=codeframe --max-warnings=0')
```

Instead of:

```js
execa('eslint', ['*.js', '*.yml', '--ignore-path=.gitignore', '--fix', '--cache', '--format=codeframe', '--max-warnings=0'])
```

This is not only for convenience, the main goal is actually to make `execa` more secure, cross-platform and faster. The reason is: many users might be tempted to use `execa.shell()` for the convenience of specifying command and arguments inside a single string, because it feels closer to typing in a terminal. However `execa.shell()` involves using a shell interpreter which is:

* less secure: it allows for [command injection](https://www.owasp.org/index.php/Command_Injection) by passing arguments like `$(rm -rf /)` or `&& rm -rf /`
* less cross-platform: it encourages using [shell-specific features](https://twitter.com/ehmicky/status/1098980276787732480) such as globbing, single quote escaping or even semicolons which won't work on `cmd.exe`.
* slower as it goes through an extra step (the shell interpreter)

Giving users the possibility to do this without `execa.shell()` prevents this issue.

Like `execa()` (and the underlying `child_process.spawn()`) arguments do not need any escaping/quoting. One exception: spaces can be escaped with a backslash if not meant as an arguments delimiter.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,17 @@ const stdio = require('./lib/stdio');
 
 const TEN_MEGABYTES = 1000 * 1000 * 10;
 
+const SPACES_REGEXP = / +/g;
+
+function parseCommand(command, args = []) {
+	const [file, ...extraArgs] = command
+		.trim()
+		.split(SPACES_REGEXP)
+		.reduce(handleEscaping, []);
+	const newArgs = [...extraArgs, ...args];
+	return [file, newArgs];
+}
+
 function handleArgs(command, args, options = {}) {
 	if (args && !Array.isArray(args)) {
 		options = args;
@@ -76,17 +87,6 @@ function handleArgs(command, args, options = {}) {
 
 	return {command, args, options, parsed};
 }
-
-function parseCommand(command, args = []) {
-	const [file, ...extraArgs] = command
-		.trim()
-		.split(SPACES_REGEXP)
-		.reduce(handleEscaping, []);
-	const newArgs = [...extraArgs, ...args];
-	return [file, newArgs];
-}
-
-const SPACES_REGEXP = / +/g;
 
 // Allow spaces to be escaped by a backslash if not meant as a delimiter
 const handleEscaping = function (tokens, token, index) {
--- a/index.js
+++ b/index.js
@@ -69,7 +69,7 @@ function handleArgs(command, args, options = {}) {
 		options.cleanup = false;
 	}
 
-	if (process.platform === 'win32' && path.basename(command, 'exe') === 'cmd') {
+	if (process.platform === 'win32' && path.basename(command, '.exe') === 'cmd') {
 		// #116
 		args.unshift('/q');
 	}
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ function handleArgs(command, args, options) {
 		buffer: true,
 		stripFinalNewline: true,
 		preferLocal: true,
-		localDir: parsed.options.cwd || process.cwd(),
+		localDir: options.cwd || process.cwd(),
 		encoding: 'utf8',
 		reject: true,
 		cleanup: true,
@@ -71,9 +71,9 @@ function handleArgs(command, args, options) {
 		options.cleanup = false;
 	}
 
-	if (process.platform === 'win32' && path.basename(parsed.command, 'exe') === 'cmd') {
+	if (process.platform === 'win32' && path.basename(command, 'exe') === 'cmd') {
 		// #116
-		parsed.args.unshift('/q');
+		args.unshift('/q');
 	}
 
 	return {command, args, options, parsed};
--- a/index.js
+++ b/index.js
@@ -92,10 +92,10 @@ const handleEscaping = function (tokens, token, index) {
 	const previousToken = tokens[index - 1];
 
 	if (!previousToken.endsWith('\\')) {
-		return tokens.concat(token);
+		return [...tokens, token];
 	}
 
-	return tokens.slice(0, index - 1).concat(`${previousToken.slice(0, -1)} ${token}`);
+	return [...tokens.slice(0, index - 1), `${previousToken.slice(0, -1)} ${token}`];
 };
 
 function handleInput(spawned, input) {
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 const SPACES_REGEXP = / +/g;
 
 // Allow spaces to be escaped by a backslash if not meant as a delimiter
-const handleEscaping = (tokens, token, index) => {
+function handleEscaping(tokens, token, index) {
 	if (index === 0) {
 		return [token];
 	}
@@ -30,7 +30,7 @@ const handleEscaping = (tokens, token, index) => {
 	}
 
 	return [...tokens.slice(0, index - 1), `${previousToken.slice(0, -1)} ${token}`];
-};
+}
 
 function parseCommand(command, args = []) {
 	if (args.length !== 0) {
--- a/index.js
+++ b/index.js
@@ -16,6 +16,10 @@ const stdio = require('./lib/stdio');
 const TEN_MEGABYTES = 1000 * 1000 * 10;
 
 function handleArgs(command, args, options) {
+	if (!options.shell && command.includes(' ')) {
+		[command, args] = parseCommand(command, args);
+	}
+
 	const parsed = crossSpawn._parse(command, args, options);
 	command = parsed.command;
 	args = parsed.args;
@@ -67,6 +71,32 @@ function handleArgs(command, args, options) {
 
 	return {command, args, options, parsed};
 }
+
+function parseCommand(command, args = []) {
+	const [newCommand, ...extraArgs] = command
+		.trim()
+		.split(SPACES_REGEXP)
+		.reduce(handleEscaping, []);
+	const newArgs = [...extraArgs, ...args];
+	return [newCommand, newArgs];
+}
+
+const SPACES_REGEXP = / +/g;
+
+// Allow spaces to be escaped by a backslash if not meant as a delimiter
+const handleEscaping = function (tokens, token, index) {
+	if (index === 0) {
+		return [token];
+	}
+
+	const previousToken = tokens[index - 1];
+
+	if (!previousToken.endsWith('\\')) {
+		return tokens.concat(token);
+	}
+
+	return tokens.slice(0, index - 1).concat(`${previousToken.slice(0, -1)} ${token}`);
+};
 
 function handleInput(spawned, input) {
 	if (input === undefined) {
--- a/index.js
+++ b/index.js
@@ -18,12 +18,15 @@ const TEN_MEGABYTES = 1000 * 1000 * 10;
 const SPACES_REGEXP = / +/g;
 
 function parseCommand(command, args = []) {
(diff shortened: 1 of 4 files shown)


ehmicky:
Rebased to latest `master`

sindresorhus:
Sorry for the slow response. Too many PRs to review.

Are you sure there's no possibility of injection attacks using this? That's what's been holding me back from adding this in the past.

readme.md
```diff
@@ -114,23 +124,24 @@ It exposes an additional `.all` stream, with `stdout` and `stderr` interleaved.
...
-### execa.stderr(file, [arguments], [options])
+### execa.stderr(command, [arguments], [options])
 
 Same as `execa()`, but returns only `stderr`.
 
 ### execa.shell(command, [options])
 
-Execute a command through the system shell. Prefer `execa()` whenever possible, as it's both faster and safer.
+Execute a command through the system shell. Prefer `execa()` whenever possible,
+as it's faster, safer and more cross-platform.
```
sindresorhus: Don't hard-wrap

readme.md
```diff
@@ -102,10 +108,14 @@ try {
 
 ## API
 
-### execa(file, [arguments], [options])
+### execa(command, [arguments], [options])
```
sindresorhus: I don't think we should change this to command as it still support file, which is what I still would recommend to people.
```suggestion
### execa(file | command, [arguments], [options])
```

Happy to consider other options.

readme.md
```diff
@@ -102,10 +108,14 @@ try {
...
 
 Execute a file.
 
+Arguments can be specified either inside `command` (a string) or `arguments`
+(an array of strings). When specified inside `command`, spaces can be escaped
+with a backslash. Otherwise arguments need neither escaping nor quoting.
```
sindresorhus: Don't hard-wrap

sindresorhus:
Does this work on `execa.sync` too?

sindresorhus:
Maybe we should add a note in `execa.shell` about the downsides of using it (like you've mention in the PR description) and that users should prefer the non-shell methods.

ehmicky:
Yes this works with `execa.sync()` as well. I have just added a test for it.

I think for the injection attacks, that's the opposite: it can happen with `shell: true` but not with `shell: false`. So if a warning should be added to the documentation, it would be for `execa.shell()` instead, as you suggesting.

There is already a note in the `README` mentioning `Prefer `execa()` whenever possible, as it's faster, safer and more cross-platform.`. Should we make any of those three words more descriptive? 

With `shell: true` (including `execa.shell()`), the command is interpreted by a shell (usually Bash or `cmd.exe`) which allows several type of injections:
  - environment variables substitution: 
```js
const userInput = '$SOME_SECRET_VAR'; 
execa.shell(`echo ${userInput}`);
```
  - spawning commands through sub-shells: 
```js
const userInput = '$( rm -rf / )'; 
execa.shell(`echo ${userInput}`);
```

With `shell: false` (which this PR uses as default for `execa()`), the command is not interpreted by a shell and instead passed directly as an OS syscall. This means no injection can be performed, as the arguments will always be passed as is. No substitution/interpration/quoting is performed, they are just string literals.

Actually this PR translates `execa('echo foo bar $var')` into `execa('echo', ['foo', 'bar', '$var'])`, i.e. it has the same characteristics (including security) as the current `execa()`.

I just checked the Node source code and arguments passed to the underlying `childProcess.spawn()` with `shell: false` ends up [here](https://github.com/nodejs/node/blob/8137668c531935c58347cd8e78147d3a9d3cc339/src/process_wrap.cc#L141) then [here on Unix](https://github.com/libuv/libuv/blob/619937c783a05b51fba95cc9a62543deeffe5fa7/src/unix/process.c#L411) and [here on Windows](https://github.com/libuv/libuv/blob/c5593b51dc98715f7f32a919301b5801ebf1a8ce/src/win/process.c#L937). Everything looks good security-wise.

readme.md
```diff
@@ -52,6 +53,21 @@ const execa = require('execa');
...
 
+	// Cancelling a spawned process
+	const subprocess = execa('node');
+	setTimeout(() => { spawned.cancel() }, 1000);
+	try {
+		await subprocess;
+	} catch (error) {
+		console.log(subprocess.killed); // true
+		console.log(error.isCanceled); // true
+	}
```
sindresorhus: This one is now defined twice.

index.js
```diff
@@ -68,6 +77,32 @@ function handleArgs(command, args, options) {
...
+		.split(SPACES_REGEXP)
+		.reduce(handleEscaping, []);
+	const newArgs = [...extraArgs, ...args];
+	return [file, newArgs];
+}
+
+const SPACES_REGEXP = / +/g;
+
+// Allow spaces to be escaped by a backslash if not meant as a delimiter
+const handleEscaping = function (tokens, token, index) {
```
```suggestion
const handleEscaping = (tokens, token, index) => {
```

readme.md
```diff
@@ -111,10 +127,12 @@ try {
 
 ## API
 
-### execa(file, [arguments], [options])
+### execa(file | command, [arguments], [options])
```
sindresorhus: I think it would be clearer to define this as overloads:

```js
### execa(file, [arguments], [options])
### execa(command, [options])
```

readme.md
```diff
@@ -111,10 +127,12 @@ try {
...
 
 Execute a file.
 
+Arguments can be specified either inside `command` (a string) or `arguments` (an array of strings). When specified inside `command`, spaces can be escaped with a backslash. Otherwise arguments need neither escaping nor quoting.
```
sindresorhus: The first sentence here could be clearer and I would mention `arguments` first as I still think that should be the preferred way to use `execa`.

readme.md
sindresorhus: Maybe we should throw a useful error if `arguments` is given and `command` has spaces?

readme.md
sindresorhus: I think we should also make clear that this is not a shell and you can't do stuff like `execa('echo $PATH')`.

index.js
```diff
@@ -68,6 +77,32 @@ function handleArgs(command, args, options) {
...
+function parseCommand(command, args = []) {
+	const [file, ...extraArgs] = command
+		.trim()
+		.split(SPACES_REGEXP)
+		.reduce(handleEscaping, []);
+	const newArgs = [...extraArgs, ...args];
+	return [file, newArgs];
+}
+
+const SPACES_REGEXP = / +/g;
```
sindresorhus: This should be defined before it's used for readability reasons.

ehmicky:
Thanks for the review! This is all fixed now.

sindresorhus:
This is looking great now. Thanks for spending so much time on perfecting it 👌

ashokaditya:
@sindresorhus It would be super nice if you were to publish these changes to npm. I've been using Snyk for checking security vulnerabilities in my projects and the latest Snyk is using execa@1.0.0 and execa@0.7.0 which still allows Arbitrary Command Injection.

More on this https://app.snyk.io/vuln/SNYK-JS-EXECA-174564

ehmicky:
I think Snyk is getting this wrong.

Node.js [`child_process.spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) (and related methods) has a `shell` option which passes input to the shell (typicall `/bin/sh`, `/bin/bash` or `cmd.exe`). If improperly used, this option can lead to command injection. As the Node.js documentation says:

> If the shell option is enabled, do not pass unsanitized user input to this function. Any input containing shell metacharacters may be used to trigger arbitrary command execution.

`execa()` is a wrapper around `child_process.spawn()`. As such it inherits that `shell` option. `execa.shell()` is just a shortcut for `execa(..., { shell: true })` (similar to `child_process.exec()`). I.e. `execa` does not introduce any vulnerabilities already present in Node.js itself. 

What this PR does is to encourage users not to use the `shell` option. It does it by allowing the command and its arguments to be specified as a single string, which is convenient and similar to the `shell` option, but without the command execution risk, because everything is escaped.

```js
const argumentsArray = ['echo', ['rm', '-r', '/some/dir']]
const argumentsString = 'echo rm -r /some/dir'

// Those commands won't execute `rm`
child_process.spawn(...argumentsArray)
execa(...argumentsArray)

// Those commands will execute `rm`
child_process.spawn(argumentsString, { shell: true }) 
child_process.exec(argumentsString)
execa(argumentsString, { shell: true })
execa.shell(argumentsString)

// Added by this PR. It won't execute `rm`
execa(argumentsString)
```

Note: @sindresorhus has npm publish access.

robcresswell:
@ehmicky Hey, I'm one of the engineers at Snyk. Thanks for the update. We're pulling it from the vuln db while we re-evaluate. Sorry for the trouble.

ehmicky:
@robcresswell thanks for the quick response!

sindresorhus:
@robcresswell This is not ok. Snyk cannot go around publishing false information like this. None of us maintainers were even contacted before the report was published, which would have prevented this mishap. I urge you to improve your processes.

ehmicky:
Snyk vulnerability is now completely removed. See https://support.snyk.io/hc/en-us/requests/374

PR #184, opened 2019-03-06
Fix `extendEnv` option not working with `shell` option

### LINKED ISSUE #183: `extendEnv` does not work when combined with `shell: true`
The `extendEnv` option does not work as intended when combined with the `shell: true` option:

```js
process.env.TEST = 'test'
const result = execa.sync('echo $TEST', { shell: true, env: {}, extendEnv: true })
assert.equal(result.stdout, 'test')
```

`stdout` is an empty string but should be `'test'`

I will work on a PR to fix this.

ehmicky:
Fixes #183 

Options parsing was not correct for the `extendEnv` option which made it not work when combined with `shell: true` option.

### DIFF
--- a/test.js
+++ b/test.js
@@ -487,11 +487,11 @@ test('do not extend environment with `extendEnv: false` option', async t => {
 	]);
 });
 
-test('use extend environment with `extendEnv: true` option', async t => {
+test('use extend environment with `extendEnv: true` option and `shell: true`', async t => {
 	process.env.TEST = 'test';
 	const command = process.platform === 'win32' ? 'echo %TEST%' : 'echo $TEST';
-	const result = await m.shellSync(command, {env: {}, extendEnv: true});
-	t.is(result.stdout, 'test');
+	const stdout = await m.stdout(command, {shell: true, env: {}, extendEnv: true});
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -487,6 +487,14 @@ test('do not extend environment with `extendEnv` option', async t => {
 	]);
 });
 
+test('use extend environment with `extendEnv: true` option', async t => {
```
sindresorhus: The title should mention the purpose is to test it with the `shell` option.

test.js
```diff
@@ -487,6 +487,14 @@ test('do not extend environment with `extendEnv` option', async t => {
...
+	process.env.TEST = 'test';
+	const command = process.platform === 'win32' ? 'echo %TEST%' : 'echo $TEST';
+	const result = await m.shellSync(command, {env: {}, extendEnv: true});
```
sindresorhus: You're awaiting a synchronous method. Should be:

```suggestion
	const result = await m.shell(command, {env: {}, extendEnv: true});
```

The synchronous methods are slow, so we should prefer to use async methods in tests whenever possible.

ehmicky:
Fixed.

PR #188, opened 2019-03-07
Add TypeScript definition

BendingBender:
~~Waiting for https://github.com/SamVerschueren/tsd-check/pull/13 until this can be merged.~~


index.d.ts
```diff
@@ -0,0 +1,365 @@
+/// <reference types="node"/>
+import {ChildProcess} from 'child_process';
+import {Stream, Readable} from 'stream';
```
```suggestion
import {Stream, Readable as ReadableStream} from 'stream';
```

index.d.ts
```diff
@@ -0,0 +1,365 @@
...
+
+export type StdIOOption =
+	| 'pipe'
+	| 'ipc'
+	| 'ignore'
+	| 'inherit'
+	| Stream
+	| number
+	| null
+	| undefined;
```
sindresorhus: Can you use the Node.js types instead?

index.d.ts
```diff
@@ -0,0 +1,365 @@
...
+
+export interface SyncOptions<EncodingType = string>
+	extends CommonOptions<EncodingType> {
+	/**
+	 * Write some input to the `stdin` of your binary.
+	 */
+	readonly input?: string | Buffer;
+}
+
+export interface ExecaReturns<StdOutErrType = string> {
```
```suggestion
export interface ExecaReturnValue<StdoutErrorType = string> {
```

sindresorhus:
Let me know if you have any feedback on how the `execa` API could be improved, especially regarding making TS usage easier. We're planning to do a major version, so we could do breaking changes to improve the UX.

sindresorhus:
> Waiting for SamVerschueren/tsd-check#13 until this can be merged.

Make sure you check out the latest changes in master when you get back to the types here. There are some API changes.

index.d.ts
BendingBender: There is actually no type in node types that fits the type required here. `StdioOptions` fro node's `child_process` is not really a match here because of the convenience options `std{in,out,err}`, that split up this type so that we can't use it any more directly from node's types.

BendingBender:
> Let me know if you have any feedback on how the `execa` API could be improved, especially regarding making TS usage easier. We're planning to do a major version, so we could do breaking changes to improve the UX.

It's actually hard to say for me, I don't really use this module.

My only problem with the current API is the usage of `encoding: null` to switch to `Buffer` encoding. I know that this is how node APIs work, I still find it quite unnatural. However, for the sake of compatibility with node, I'd leave it like this.

index.d.ts
```diff
@@ -0,0 +1,398 @@
+/// <reference types="node"/>
+import {ChildProcess} from 'child_process';
+import {Stream, Readable as ReadableStream} from 'stream';
+
+export type StdIOOption =
```
```suggestion
export type StdioOption =
```

index.d.ts
```diff
@@ -0,0 +1,398 @@
...
+
+	/**
+	 * Child's [stdio](https://nodejs.org/api/child_process.html#child_process_options_stdio) configuration.
+	 *
+	 * @default 'pipe'
+	 */
+	readonly stdio?: 'pipe' | 'ignore' | 'inherit' | ReadonlyArray<StdIOOption>;
+
+	/**
+	 * Prepare child to run independently of its parent process. Specific behavior [depends on the platform](https://nodejs.org/api/child_process.html#child_process_options_detached).
```
```suggestion
	 * Prepare child to run independently of its parent process. Specific behavior [depends on the platform](https://nodejs.org/api/child_process.html#child_process_options_detached).
	 *
	 *  @default false
```

index.d.ts
```diff
@@ -0,0 +1,398 @@
...
+
+export interface SyncOptions<EncodingType = string>
+	extends CommonOptions<EncodingType> {
+	/**
+	 * Write some input to the `stdin` of your binary.
+	 */
+	readonly input?: string | Buffer;
+}
+
+export interface ExecaReturnBase<StdOutErrType> {
```
```suggestion
export interface ExecaReturnBase<StdoutErrorType> {
```

index.d.ts
```diff
@@ -0,0 +1,398 @@
...
+	extends ExecaSyncError<StdOutErrType> {
+	/**
+	 * The output of the process with `stdout` and `stderr` interleaved.
+	 */
+	all: StdOutErrType;
+}
+
+export interface ExecaChildPromise<StdOutErrType> {
+	catch<ResultType = never>(
+		onrejected?:
```
```suggestion
		onRejected?:
```

index.d.ts
sindresorhus: Hmm, actually, it's just just the `stdout` type, but also `stderr`. The naming might make it confusing for users.

ehmicky:
I am not very familiar with TypeScript yet (it's on my short-term to-do list), so I will pass on the code review.

index.d.ts
```diff
@@ -0,0 +1,400 @@
...
+	 * @default process.cwd()
+	 */
+	readonly cwd?: string;
+
+	/**
+	 * Environment key-value pairs. Extends automatically from `process.env`. Set `extendEnv` to `false` if you don't want this.
+	 *
+	 * @default process.env
+	 */
+	readonly env?: NodeJS.ProcessEnv;
```
sindresorhus: @BendingBender Do you know how we have to convince to reduce globals for the Node.js type definition and make stuff like this importable instead?

BendingBender: @sindresorhus I suppose that making a PR on DefinitelyTyped is the easiest way to request feedback and start a discussion.

Just out of curiosity, is is just aesthetical or are there technical reasons why you want to migrate away from global usage?

sindresorhus: I don't want to put any work into it before I know whether it's an acceptable change. I guess I'll just open an issue about it.

> Just out of curiosity, is it just aesthetical or are there technical reasons why you want to migrate away from global usage?

I hate globals :P Node.js is all about explicitly importing things instead of dependending on globals (like the browser). Everything on `NodeJS` could just be imported instead.

sindresorhus:
@BendingBender Can you update for https://github.com/sindresorhus/execa/pull/193 ? I'm trying to remove most of the `null` usage.

BendingBender:
Yeah, it's on my radar 😜

sindresorhus:
And also https://github.com/sindresorhus/execa/commit/79356cc27d52b231fee097a11bcf4abc691ff02c

BendingBender:
Waiting for https://github.com/SamVerschueren/tsd-check/pull/19

PR #189, opened 2019-03-10
Make the spawned process cancelable

### LINKED ISSUE #113: Make it cancelable
Using: https://github.com/sindresorhus/p-cancelable

It's possible to call `cp.kill()`, but would be nicer if it was cancelable.

ammarbinfaisal:
see if this is okay.
I was wondering whether there should we a check if oncancel property in options object passed by someone is actually a function. is this required?

Fixes #113

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,7 +143,7 @@ function getStream(process, stream, {encoding, buffer, maxBuffer}) {
 function makeError(result, options) {
 	const {stdout, stderr, code, signal} = result;
 	let {error} = result;
-	const {joinedCommand, timedOut, canceled, parsed: {options: {timeout}}} = options;
+	const {joinedCommand, timedOut, isCanceled, parsed: {options: {timeout}}} = options;
 
 	const [exitCodeName, exitCode] = getCode(result, code);
 
@@ -152,7 +152,7 @@ function makeError(result, options) {
 		error = new Error(message);
 	}
 
-	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCode, exitCodeName, canceled});
+	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCode, exitCodeName, isCanceled});
 	error.message = `Command ${prefix}: ${error.message}`;
 
 	error.code = exitCode || exitCodeName;
@@ -164,7 +164,7 @@ function makeError(result, options) {
 	error.signal = signal || null;
 	error.cmd = joinedCommand;
 	error.timedOut = Boolean(timedOut);
-	error.canceled = canceled;
+	error.isCanceled = isCanceled;
 
 	if ('all' in result) {
 		error.all = result.all;
@@ -185,12 +185,12 @@ function getCode({error = {}}, code) {
 	return [];
 }
 
-function getErrorPrefix({timedOut, timeout, signal, exitCodeName, exitCode, canceled}) {
+function getErrorPrefix({timedOut, timeout, signal, exitCodeName, exitCode, isCanceled}) {
 	if (timedOut) {
 		return `timed out after ${timeout} milliseconds`;
 	}
 
-	if (canceled) {
+	if (isCanceled) {
 		return 'was canceled';
 	}
 
@@ -244,7 +244,7 @@ module.exports = (command, args, options) => {
 
 	let timeoutId = null;
 	let timedOut = false;
-	let canceled = false;
+	let isCanceled = false;
 
 	const cleanup = () => {
 		if (timeoutId) {
@@ -310,12 +310,12 @@ module.exports = (command, args, options) => {
 		result.stderr = results[2];
 		result.all = results[3];
 
-		if (result.error || result.code !== 0 || result.signal !== null || canceled) {
+		if (result.error || result.code !== 0 || result.signal !== null || isCanceled) {
 			const error = makeError(result, {
 				joinedCommand,
 				parsed,
 				timedOut,
-				canceled
+				isCanceled
 			});
 
 			// TODO: missing some timeout logic for killed
@@ -342,7 +342,7 @@ module.exports = (command, args, options) => {
 			signal: null,
 			cmd: joinedCommand,
 			timedOut: false,
-			canceled: false
+			isCanceled: false
 		};
 	}), destroy);
 
@@ -355,13 +355,13 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = function () {
+	spawned.cancel = () => {
 		if (spawned.killed) {
 			return;
 		}
 
 		if (spawned.kill()) {
-			canceled = true;
+			isCanceled = true;
 		}
 	};
 
--- a/index.js
+++ b/index.js
@@ -152,7 +152,7 @@ function makeError(result, options) {
 		error = new Error(message);
 	}
 
-	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCode, exitCodeName, isCanceled});
+	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCodeName, exitCode, isCanceled});
 	error.message = `Command ${prefix}: ${error.message}`;
 
 	error.code = exitCode || exitCodeName;
--- a/index.js
+++ b/index.js
@@ -8,7 +8,6 @@ const npmRunPath = require('npm-run-path');
 const isStream = require('is-stream');
 const _getStream = require('get-stream');
 const pFinally = require('p-finally');
-const PCancelable = require('p-cancelable');
 const onExit = require('signal-exit');
 const errname = require('./lib/errname');
 const stdio = require('./lib/stdio');
@@ -215,6 +214,7 @@ module.exports = (command, args, options) => {
 
 	let timeoutId = null;
 	let timedOut = false;
+	let canceled = false;
 
 	const cleanup = () => {
 		if (timeoutId) {
@@ -235,7 +235,7 @@ module.exports = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
-	const processDone = new PCancelable((resolve, reject, onCancel) => {
+	const processDone = new Promise(resolve => {
 		spawned.on('exit', (code, signal) => {
 			cleanup();
 			resolve({code, signal});
@@ -252,11 +252,16 @@ module.exports = (command, args, options) => {
 				resolve({error});
 			});
 		}
-
-		onCancel(() => {
-			spawned.kill();
-		});
 	});
+
+	processDone.cancel = function () {
+		if (canceled) {
+			return;
+		}
+
+		canceled = true;
+		spawned.kill();
+	};
 
 	function destroy() {
 		if (spawned.stdout) {
@@ -305,7 +310,8 @@ module.exports = (command, args, options) => {
 			killed: false,
 			signal: null,
 			cmd: joinedCommand,
-			timedOut: false
+			timedOut: false,
+			canceled
 		};
 	}), destroy);
 
@@ -316,7 +322,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = (reason = 'command was canceled') => processDone.cancel(reason);
+	spawned.cancel = () => processDone.cancel();
 
 	// TOOD: Remove the `if`-guard when targeting Node.js 10
 	if (Promise.prototype.finally) {
--- a/index.js
+++ b/index.js
@@ -235,7 +235,7 @@ module.exports = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
-	const processDone = new PCancelable((resolve, reject, oncancel) => {
+	const processDone = new PCancelable((resolve, reject, onCancel) => {
 		spawned.on('exit', (code, signal) => {
 			cleanup();
 			resolve({code, signal});
@@ -253,7 +253,7 @@ module.exports = (command, args, options) => {
 			});
 		}
 
-		oncancel(() => {
+		onCancel(() => {
 			spawned.kill();
 		});
 	});
@@ -316,7 +316,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = () => processDone.cancel();
+	spawned.cancel = reason => processDone.cancel(reason);
 
 	// TOOD: Remove the `if`-guard when targeting Node.js 10
 	if (Promise.prototype.finally) {
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const npmRunPath = require('npm-run-path');
 const isStream = require('is-stream');
 const _getStream = require('get-stream');
 const pFinally = require('p-finally');
+const PCancelable = require('p-cancelable');
 const onExit = require('signal-exit');
 const errname = require('./lib/errname');
 const stdio = require('./lib/stdio');
@@ -234,7 +235,7 @@ module.exports = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
-	const processDone = new Promise(resolve => {
+	const processDone = new PCancelable((resolve, reject, oncancel) => {
 		spawned.on('exit', (code, signal) => {
 			cleanup();
 			resolve({code, signal});
@@ -251,6 +252,10 @@ module.exports = (command, args, options) => {
 				resolve({error});
 			});
 		}
+
+		oncancel(() => {
+			spawned.kill();
+		});
 	});
 
 	function destroy() {
@@ -311,6 +316,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
+	spawned.cancel = () => processDone.cancel();
 
 	// TOOD: Remove the `if`-guard when targeting Node.js 10
 	if (Promise.prototype.finally) {
--- a/index.js
+++ b/index.js
@@ -356,9 +356,12 @@ module.exports = (command, args, options) => {
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
 	spawned.cancel = function () {
-		if (!spawned.killed) {
+		if (spawned.killed) {
+			return;
+		}
+
+		if (spawned.kill()) {
 			canceled = true;
-			spawned.kill();
 		}
 	};
 
--- a/index.js
+++ b/index.js
@@ -124,7 +124,7 @@ function getStream(process, stream, {encoding, buffer, maxBuffer}) {
 function makeError(result, options) {
 	const {stdout, stderr, code, signal} = result;
 	let {error} = result;
-	const {joinedCommand, timedOut, parsed: {options: {timeout}}} = options;
+	const {joinedCommand, timedOut, canceled, parsed: {options: {timeout}}} = options;
 
 	const [codeString, codeNumber] = getCode(result, code);
 
@@ -133,9 +133,8 @@ function makeError(result, options) {
 		error = new Error(message);
 	}
 
-	const prefix = getErrorPrefix({timedOut, timeout, signal, codeString, codeNumber});
+	const prefix = getErrorPrefix({timedOut, timeout, signal, codeString, codeNumber, canceled});
 	error.message = `Command ${prefix}: ${error.message}`;
-
 	error.code = codeNumber || codeString;
 	error.stdout = stdout;
 	error.stderr = stderr;
@@ -143,6 +142,7 @@ function makeError(result, options) {
 	error.signal = signal || null;
 	error.cmd = joinedCommand;
 	error.timedOut = Boolean(timedOut);
+	error.canceled = canceled;
 
 	return error;
 }
@@ -159,9 +159,13 @@ function getCode({error = {}}, code) {
 	return [];
 }
 
-function getErrorPrefix({timedOut, timeout, signal, codeString, codeNumber}) {
+function getErrorPrefix({timedOut, timeout, signal, codeString, codeNumber, canceled}) {
 	if (timedOut) {
 		return `timed out after ${timeout} milliseconds`;
+	}
+
+	if (canceled) {
+		return 'was canceled';
 	}
 
 	if (signal) {
@@ -254,15 +258,6 @@ module.exports = (command, args, options) => {
 		}
 	});
 
-	processDone.cancel = function () {
-		if (canceled) {
-			return;
-		}
-
-		canceled = true;
-		spawned.kill();
-	};
-
 	function destroy() {
 		if (spawned.stdout) {
 			spawned.stdout.destroy();
@@ -283,11 +278,12 @@ module.exports = (command, args, options) => {
 		result.stdout = results[1];
 		result.stderr = results[2];
 
-		if (result.error || result.code !== 0 || result.signal !== null) {
+		if (result.error || result.code !== 0 || result.signal !== null || canceled) {
 			const error = makeError(result, {
 				joinedCommand,
 				parsed,
-				timedOut
+				timedOut,
+				canceled
 			});
 
 			// TODO: missing some timeout logic for killed
@@ -311,7 +307,7 @@ module.exports = (command, args, options) => {
 			signal: null,
 			cmd: joinedCommand,
 			timedOut: false,
-			canceled
+			canceled: false
 		};
 	}), destroy);
 
@@ -322,7 +318,12 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = () => processDone.cancel();
+	spawned.cancel = function () {
+		if (!spawned.killed) {
+			canceled = true;
+			spawned.kill();
+		}
+	};
 
 	// TOOD: Remove the `if`-guard when targeting Node.js 10
 	if (Promise.prototype.finally) {
--- a/index.js
+++ b/index.js
@@ -316,7 +316,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = reason => processDone.cancel(reason);
+	spawned.cancel = (reason = 'command was canceled') => processDone.cancel(reason);
 
 	// TOOD: Remove the `if`-guard when targeting Node.js 10
 	if (Promise.prototype.finally) {
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
 		"get-stream": "^4.0.0",
 		"is-stream": "^1.1.0",
 		"npm-run-path": "^2.0.0",
+		"p-cancelable": "^1.1.0",
 		"p-finally": "^1.0.0",
 		"signal-exit": "^3.0.0",
 		"strip-final-newline": "^2.0.0"
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
 		"get-stream": "^4.0.0",
 		"is-stream": "^1.1.0",
 		"npm-run-path": "^2.0.0",
-		"p-cancelable": "^1.1.0",
 		"p-finally": "^1.0.0",
 		"signal-exit": "^3.0.0",
 		"strip-final-newline": "^2.0.0"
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -251,6 +252,8 @@ module.exports = (command, args, options) => {
 				resolve({error});
 			});
 		}
+
+		oncancel(parsed.options.oncancel);
```
sindresorhus: This should kill the process, not execute a function from an option.

index.js
```diff
@@ -234,7 +235,7 @@ module.exports = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
-	const processDone = new Promise(resolve => {
+	const processDone = new PCancelable((resolve, reject, oncancel) => {
```
```suggestion
	const processDone = new PCancelable((resolve, reject, onCancel) => {
```

sindresorhus:
You need to add tests and docs too.

index.js
ammarbinfaisal: done

ammarbinfaisal:
There is CancelError being thrown. In test.js and readme, I've passed an empty function `() => {}` to spawned.catch to ignore it. should I use a try catch block inside the main code to do this?

sindresorhus:
No, you're missing the point of CancelError. Please read the `p-cancelable` readme and understand how it works.

index.js
```diff
@@ -311,6 +316,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
+	spawned.cancel = reason => processDone.cancel(reason);
```
ehmicky: Maybe you should provide a default `reason` if none is passed, like `The command was canceled`

ehmicky:
I think it would be nice to add properties similar to how `timedOut`, `failed` and `killed` and currently handled:
  - add `result.canceled`. Always defined (`true` or `false`) whether on success or failure. Exception: not defined with `sync()` since synchronous calls cannot be canceled.
  - enhance `result.message` (`getErrorPrefix()` function) to add `was canceled: ${reason}`. `: ${reason}` should not be appended in it's default `reason`.

readme.md
```diff
@@ -51,6 +51,13 @@ const execa = require('execa');
 	const {stdout} = await execa.shell('echo unicorns');
 	//=> 'unicorns'
 
+	// Cancelling a spawned process
+	const spawned = execa("echo unicorns");
+	spawned.catch( error => {
+		// This would be equal to the reason provided for canceling
```
ehmicky: This comment should probably be removed because `error.message` is actually changed by `execa` so `error.message` would not be exactly `"reason for canceling"`.

readme.md
```diff
@@ -51,6 +51,13 @@ const execa = require('execa');
 	const {stdout} = await execa.shell('echo unicorns');
 	//=> 'unicorns'
 
+	// Cancelling a spawned process
+	const spawned = execa("echo unicorns");
+	spawned.catch( error => {
```
ehmicky: Minor coding style: no space after parenthesis here.

readme.md
```diff
@@ -51,6 +51,13 @@ const execa = require('execa');
...
+		// This would be equal to the reason provided for canceling
+		console.log(error.message);
+	}); // To handle the CancelError
```
ehmicky: I think this comment will be a little obscure to readers, so could be left out.

readme.md
```diff
@@ -51,6 +51,13 @@ const execa = require('execa');
...
+	spawned.cancel("reason for canceling");
```
ehmicky: To simplify the example I would personally call `cancel()` without arguments here, but it's up to you if you think it's better.

test.js
```diff
@@ -569,3 +569,16 @@ if (Promise.prototype.finally) {
 		t.is(result.message, 'called');
 	});
 }
+
+test('the spawned process is cancelable', t => {
+	const spawned = execa('ls');
```
ehmicky: `ls` is not available on Windows. You should use one of the test fixtures e.g. `node fixtures/forever`

test.js
```diff
@@ -569,3 +569,16 @@ if (Promise.prototype.finally) {
...
+	spawned.catch(error => {
+		if (error.message === 'reason' && error.isCanceled) {
+			return;
+		}
+
+		throw error;
+	}); // To handle the CancelError
+	spawned.cancel('reason');
+	t.true(spawned.killed);
```
ehmicky: I think `canceled` should be tested here instead of `killed`. See my other comment about adding that property.

test.js
```diff
@@ -569,3 +569,16 @@ if (Promise.prototype.finally) {
 		t.is(result.message, 'called');
 	});
 }
+
+test('the spawned process is cancelable', t => {
+	const spawned = execa('ls');
+	spawned.catch(error => {
```
ehmicky: I think you should instead use `await t.throwsAsync(spawned, ...)`

test.js
```diff
@@ -569,3 +569,16 @@ if (Promise.prototype.finally) {
...
+		if (error.message === 'reason' && error.isCanceled) {
+			return;
+		}
+
+		throw error;
+	}); // To handle the CancelError
+	spawned.cancel('reason');
```
ehmicky: I think this test should be split into several tests each testing different parts of the PR:
  - test that after `cancel()` `result.canceled` is `true`
  - test that without `cancel()` `result.canceled` is `false`
  - test that after `cancel()` `promise is rejected with the right error
  - test that the message passed as a `reason` appears somewhere inside `result.message`

readme.md
ammarbinfaisal: what you are inferring seems better

test.js
ammarbinfaisal: I'd have to explore ava's docs, I never written tests before :sweat_smile:

ehmicky: If you've never written tests before, I think this might a great opportunity for you here. [Ava](https://github.com/avajs/ava) is a great project and it's beginner-friendly.

For me the main thing is:
  - keeping tests simple: each test does one thing, then check if the expected result is correct.
  - splitting tests into small tests that each test one thing.
  - keeping tests independent from each other. Ava runs tests in parallel by default so it enforces avoiding using any global state.
  - using the tools the test runner provides (here Ava), e.g. `t.throwsAsync()` assertions, etc.
  - making sure you're covering all the features you're adding and things that could break.
  - remembering target platform: Windows/Mac/Linux, Node.js `>= 8`

Please ask any questions you might have while learning Ava and writing those tests, we're here to help :)

ammarbinfaisal: thanks a lot! actually I am a bit busy right now. I'll do all this tonight (5-6 hours from now)

ehmicky: There is no rush :+1:

ammarbinfaisal:
> I think it would be nice to add properties similar to how timedOut, failed and killed and currently handled:

add result.canceled. Always defined (true or false) whether on success or failure. Exception: not defined with sync() since synchronous calls cannot be canceled.
enhance result.message (getErrorPrefix() function) to add was canceled: ${reason}. : ${reason} should not be appended in it's default reason.

@ehmicky, I can't figure out how to do this. On cancelling p-cancelable itself throws CancelError and this is not notified in the handlePromise function - I mean in the if block which checks whether there is an error and and creates a proper execa error using makeError.

test.js
```diff
@@ -569,3 +570,30 @@ if (Promise.prototype.finally) {
...
+test('invoking cancel method throws CancelError', async t => {
+	const spawned = execa('ls');
+	spawned.cancel();
+	await t.throwsAsync(spawned, CancelError);
+});
+
+test('default CancelError message is "command was canceled"', async t => {
+	const spawned = execa('ls');
+	spawned.cancel();
+	const error = await t.throwsAsync(spawned);
```
ehmicky: You can do those two lines in a single one: `await t.throwsAsync(spawned, { message: 'command was canceled' })`

Same thing for the error tests.

See https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expected-message for more information.

ehmicky:
Something like this might work if I understand your problem correctly:

```js
	const processDone = new PCancelable((resolve, reject, onCancel) => {
        let canceled = false;
			cleanup();
			resolve({code, signal, canceled});
		});

		/* more code */

		onCancel(() => {
            canceled = true;
			spawned.kill();
		});
	});
```

ammarbinfaisal:
> Something like this might work if I understand your problem correctly:

@ehmicky No, it won't. I also thought of taking this variable in handlePromise function. As soon as we cancel the promise p-cancelable itself throws the error as you can see [here](https://github.com/sindresorhus/p-cancelable/blob/master/index.js).

ehmicky:
This might work if you put the `canceled` variable one level up:

```js
let canceled = false;
const processDone = new PCancelable((resolve, reject, onCancel) => {
  /* more code */
  onCancel(() => {
    canceled = true;
    spawned.kill();
  });
});

const handlePromise = () => pFinally(Promise.all([
		processDone,
		getStream(spawned, 'stdout', {encoding, buffer, maxBuffer}),
		getStream(spawned, 'stderr', {encoding, buffer, maxBuffer})
	]).then(results => { // eslint-disable-line promise/prefer-await-to-then
		const result = results[0];
		result.stdout = results[1];
		result.stderr = results[2];

		if (result.error || result.code !== 0 || result.signal !== null || canceled) {
			const error = makeError(result, {
				joinedCommand,
				parsed,
				timedOut,
                canceled
			});

			// TODO: missing some timeout logic for killed
			// https://github.com/nodejs/node/blob/master/lib/child_process.js#L203
			// error.killed = spawned.killed || killed;
			error.killed = error.killed || spawned.killed;

			if (!parsed.options.reject) {
				return error;
			}

			throw error;
		}

		return {
			stdout: handleOutput(parsed.options, result.stdout),
			stderr: handleOutput(parsed.options, result.stderr),
			code: 0,
			failed: false,
			killed: false,
			signal: null,
			cmd: joinedCommand,
			timedOut: false,
			canceled: false,
		};
	}), destroy);
```

Notice the `|| canceled` condition: if the promise is rejected because of cancellation, exit code might still be `0` with no errors nor signals (in case child process killing happens on the next tick), but we want to make sure cancellation is taken into account.

ehmicky:
Ok I just realized from my above comment that the `processDone` promise will be rejected with a `CancelError`, which won't work with the rest of the code which assumes otherwise.

However when we cancel the promise, the child process is killed. This will lead to the `exit` event to be called, which will resolve the promise. Also we do want to wait on the child process to be killed before returning the result. 

So I suggest two solutions. First using [`onCancel.shouldReject = false`](https://github.com/sindresorhus/p-cancelable/blob/master/readme.md#new-pcancelableexecutor).

Second not using `p-cancelable` and simply doing:

```js
	const processDone = new Promise((resolve, reject) => {
			cleanup();
			resolve({code, signal});
		});

		spawned.on('error', error => {
			cleanup();
			resolve({error});
		});

		if (spawned.stdin) {
			spawned.stdin.on('error', error => {
				cleanup();
				resolve({error});
			});
		}
	});

    let canceled = false;
    processDone.cancel = function() {
      if (canceled) { return; }

      canceled = true;
      spawned.kill();
    }
```

ammarbinfaisal:
> This might work if you put the canceled variable one level up

I actually tried doing all this.

`
	// eslint-disable-next-line promise/prefer-await-to-then
	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
	spawned.catch = onRejected => handlePromise().catch(onRejected);
	spawned.cancel = (reason = 'command was canceled') => processDone.cancel(reason);  // the problem is with this line

`

Since, calling spawned.cancel class processDone.cancel, CancelError is thrown and proccess simply terminates. The callback would be executed inside onCancel function and canceled would become true however that if block won't be  executed so we can't make a proper execa error and do the rest of thing you are thinking of. atleast, i can't figue out a way. We can prevent cancelError from being thrown([https://github.com/sindresorhus/p-cancelable#new-pcancelableexecutor](see this)) and then make a new error on our own if canceled is `true`. Will that be okay?

ammarbinfaisal:
@ehmicky  I can't understand why is this happening but if I use `noop`, `node forever` instead of ls in the latest added test, it fails

```
  1 test failed

  test › calling cancel method makes result.cancel true


  Rejected promise returned by test. Reason:

  Error {
    cmd: 'node fixtures',
    code: 2,
    errno: 'ENOENT',
    failed: true,
    killed: false,
    path: 'node fixtures',
    signal: null,
    spawnargs: [],
    stderr: '',
    stdout: '',
    syscall: 'spawn node fixtures',
    timedOut: false,
    message: 'Command failed with exit code 2 (ENOENT): spawn node fixtures ENOENT',
  }

```

ehmicky:
Could you post the code of that test?

ammarbinfaisal:
this is the failing test

```
test('calling cancel method makes result.cancel true', async t => {
	const spawned = execa('ls');
	spawned.cancel();
	const result = await spawned;
	t.true(result.canceled);
});
```

I think this depends on the time taken by the process like even with ls it failed one time on my PC and ran the next time
Screenshot from 2019-03-13 19-49-10

ammarbinfaisal:
I think using p-cancelable and allowing CancelError to be thrown is the best think which can be done; probably @sindresorhus asked to use p-cancelable because of that and why add a canceled property in result when the person using this can just check for CancelError(i mean use catch) :thinking:

ehmicky:
You're showing a different test that the initial failure. The initial failure you mentioned was about `Command failed with exit code 2 (ENOENT): spawn node fixtures ENOENT`. The one you are now posting uses a different command and is a separate problem. Let's try not to fork too many different discussions, otherwise it's hard to keep track.

What you are trying is also dependent on how you will implement this feature. If there is a timing issue there, the issue is not the test but the implementation.

About letting the `CancelError` exception propagate: the issue there is that:
  1) while `process.kill()` has been called, the child process might not be called right away (synchronously) by Node.js, i.e. `execa()` might resolve while the child process is still being killed. This is probably more proper to only resolve `execa()` once the child process is killed.
  2) because of 1), the result will be missing the termination `signal` and the `code`
  3) it adds a new value to which `processDone()` can resolve, which leads to more polymorphic code, i.e. more complexity.

Either of the two solutions above would solve those issues (or any third solution you might think of).

ammarbinfaisal:
Sorry, I didn't see that was ENOENT.
I think I have done what you asked for or am I missing something?

index.js
```diff
@@ -311,6 +322,7 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
+	spawned.cancel = () => processDone.cancel();
```
ehmicky: This could be merged with `processDone.cancel()`. I.e. if `processDone.cancel = function {` was `spawned.cancel = function () {` instead, then this line of code could be removed.

ammarbinfaisal: You mean it should be like this 
```
spawned.cancel = processDone.cancel;
```

ehmicky:
Ah ok indeed the code was changed. I thought otherwise because of your comment:

> I think using p-cancelable and allowing CancelError to be thrown is the best think which can be done

For the test, let's break it down into two different problems:
  a) You had an `ENOENT` error. Please let me know if the fix I suggested above is working.
  b) You had a timing / race condition issue. I think the new implementation might have fixed that problem.

index.js
ehmicky: No:
  - removing the `spawned.cancel = () => processDone.cancel()'` line
  - replacing `processDone.cancel = function {` by `spawned.cancel = function () {`

package.json
```diff
@@ -41,6 +41,7 @@
 		"get-stream": "^4.0.0",
 		"is-stream": "^1.1.0",
 		"npm-run-path": "^2.0.0",
+		"p-cancelable": "^1.1.0",
```
ehmicky: This can be removed if it's not used.

ammarbinfaisal:
> You had an ENOENT error. Please let me know if the fix I suggested above is working.

that ENOENT problem isn't there anymore using `noop`; still, this happens

```
1 test failed

  test › calling cancel method makes result.cancel true

  /home/ammar/osource/execa/index.js:5

   4: const childProcess = require('child_process');           
   5: const crossSpawn = require('cross-spawn');               
   6: const stripFinalNewline = require('strip-final-newline');

  Rejected promise returned by test. Reason:

  Error {
    cmd: 'noop',
    code: undefined,
    failed: true,
    killed: true,
    signal: 'SIGTERM',
    stderr: '',
    stdout: '',
    timedOut: false,
    message: 'Command was killed with SIGTERM: noop',
  }

  makeError (index.js:5:1171)
  Promise.all.then.results (index.js:7:376)
```

<br>

**One thing more**

```
const execa = require("../execa");
const spawned = execa("ls");

spawned.cancel();
spawned.then(result => {
    console.log(result.canceled, spawned.killed)
}).catch(error => console.log(error));
```
this logs
`true true`
but if I replace `ls` with `tree` then
```
{ Error: Command was killed with SIGTERM: tree
    at makeError (/home/ammar/osource/execa/index.js:133:11)
    at Promise.all.then.results (/home/ammar/osource/execa/index.js:278:18)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  code: undefined,
  stdout: '',
  stderr: '',
  failed: true,
  signal: 'SIGTERM',
  cmd: 'tree',
  timedOut: false,
  killed: true }
```

index.js
```diff
@@ -253,6 +254,15 @@ module.exports = (command, args, options) => {
 		}
 	});
 
+	processDone.cancel = function () {
+		if (canceled) {
+			return;
+		}
+
+		canceled = true;
+		spawned.kill();
```
ehmicky: Probably this should actually be:

```js
if (spawned.kill()) {
  canceled = true;
}
```

Otherwise trying to `cancel()` an already killed process will set it `canceled` although `cancel()` did nothing.

See Node source code: https://github.com/nodejs/node/blob/master/lib/internal/child_process.js#L435

ehmicky:
Both code samples seem to be wrong:
  - in the first one, `spawned` should be rejected (since it was canceled), i.e. `catch()` callback should be called, not `then()` callback
  - the second is almost correct except the return value should include `canceled: true`

index.js
```diff
@@ -322,7 +318,12 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
-	spawned.cancel = () => processDone.cancel();
+	spawned.cancel = function () {
+		if (!spawned.killed) {
```
ammarbinfaisal: I did `!spawned.killed` instead of `spawn.kill()` because doing that would give another error(prcess was killed with SIGTERM) and not the one with message "Command was canceled" and property canceled equal to `true`(I think this is the required thing)

ehmicky: This sounds good but I think you should still check the return value of `spawned.kill()`. For example if the child process exited successfully (not killed) `spawned.kill()` will return `false`. In that case it was not canceled, and `canceled` should not be `true`.

test.js
```diff
@@ -586,3 +586,42 @@ if (Promise.prototype.finally) {
...
+	});
+	t.regex(error.message, /Command\swas\scanceled/);
+});
+
+test('error.canceled is true when cancel method is used', async t => {
+	const spawned = execa('noop');
+	spawned.cancel();
+	const error = await t.throwsAsync(async () => {
+		await spawned;
+	});
```
ehmicky: You can directly use `t.throwsAsync(spawned)` (same thing for the other tests).

test.js
```diff
@@ -586,3 +586,42 @@ if (Promise.prototype.finally) {
...
+	t.false(result.canceled);
+});
+
+test('calling cancel method throws an error with message "Command was canceled"', async t => {
+	const spawned = execa('noop');
+	spawned.cancel();
+	const error = await t.throwsAsync(async () => {
+		await spawned;
+	});
+	t.regex(error.message, /Command\swas\scanceled/);
```
ehmicky: You can directly use `t.throwsAsync(spawned, { message: /Command was canceled/ })`

test.js
```diff
@@ -586,3 +586,42 @@ if (Promise.prototype.finally) {
...
+});
```
ehmicky: I think the regExp can use spaces instead of `\s` because we know for sure those will just be single spaces.

test.js
```diff
@@ -586,3 +586,42 @@ if (Promise.prototype.finally) {
...
+
+test('cancel method kills the spawned process', t => {
+	const spawned = execa('node');
+	spawned.cancel();
+	t.true(spawned.killed);
+});
+
+test('result.canceled is false when spawned.cancel isn\'t called', async t => {
+	const spawned = execa('noop');
+	const result = await spawned;
```
ehmicky: You can directly do `const result = await execa('noop')`

readme.md
ehmicky: Actually both comments can probably be left out above.

test.js
```diff
@@ -586,3 +586,42 @@ if (Promise.prototype.finally) {
...
+test('error.canceled is true when cancel method is used', async t => {
+	const spawned = execa('noop');
+	spawned.cancel();
+	const error = await t.throwsAsync(async () => {
+		await spawned;
+	});
+	t.true(error.canceled);
+});
+
+test('error.canceled is false when kill method is used', async t => {
```
ehmicky: Except for comments above, tests look good now!

Just three extra tests that you might want to add: 
  - calling `spawned.cancel()` twice should have same behavior as calling it once.
  - calling `spawned.cancel()` on a child process that has already successfully exited
  - calling `spawned.cancel()` on a child process that has already been killed

index.js
```diff
@@ -358,7 +358,9 @@ module.exports = (command, args, options) => {
 	spawned.cancel = function () {
 		if (!spawned.killed) {
 			canceled = true;
-			spawned.kill();
+			if (!spawned.kill()) {
+				canceled = false;
+			}
 		}
```
ehmicky: If you call twice `cancel()`, the second time `canceled` will be set back to `false`.

This could be fixed by doing instead:

```js
spawned.cancel = function() {
  if (spawned.killed) { return; }

  if (spawned.kill()) {
    canceled = true;
  }
};
```

test.js
```diff
@@ -586,3 +586,56 @@ if (Promise.prototype.finally) {
...
+test('result.canceled is false when spawned.cancel isn\'t called', async t => {
+	const result = await execa('noop');
+	t.false(result.canceled);
+});
+
+test('calling cancel method throws an error with message "Command was canceled"', async t => {
+	const spawned = execa('noop');
+	spawned.cancel();
+	const error = await t.throwsAsync(spawned);
+	t.regex(error.message, /Command was canceled/);
```
ehmicky: You can directly use t.throwsAsync(spawned, { message: /Command was canceled/ })

ammarbinfaisal: yeah, sorry for forgetting this

index.js
```diff
@@ -356,7 +356,7 @@ module.exports = (command, args, options) => {
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
 	spawned.cancel = function () {
-		if (!spawned.killed) {
+		if (!spawned.killed && spawned.exitCode !== 0) {
```
ehmicky: In which situation can `spawned` be `killed` but with an `exitCode` of `0`?

ammarbinfaisal: oh sorry! I should check for something else...but that won't work

ammarbinfaisal: I am not wrong then exitcode is 0 when the process was successful. this would just check whether the proccess hasn't successfully exited and can be no longer killed

ehmicky: As far as I understand the only thing we need to check is whether the process has already terminated. It should not matter whether it was successful or not, and whether it was done through a signal (`killed`) or not. 

This is already given by Node.js with the return value of `spawned.kill()` so wouldn't this work instead?

```js
if (canceled) { return }

if (process.kill()) {
  canceled = true;
}
```

readme.md
ammarbinfaisal: done

index.js
ammarbinfaisal: This would work but then what would be the use of `canceled` at all. We won't be able to use it. On killing the process, processDone promise would get resolved

```
                // some code
		spawned.on('error', error => {
			cleanup();
			resolve({error});  // because of this
		});
                // more code
	});
```

and then in handlePromise function this condition check would pass with canceled still being `false`.

```
if (result.error || result.code !== 0 || result.signal !== null || canceled) {
			const error = makeError(result, {
				joinedCommand,
				parsed,
				timedOut,
				canceled
			});
```
 Thus, the error made by makeError won't be as desired - having canceled == true and a proper message

ammarbinfaisal: all what is mean is that we can;t execute `kill()` before making `canceled` `true`

ehmicky: Isn't `handlePromise` only resolved on the next tick?  I.e. the order would be: 
  1) `spawned.kill()` is fired
  2) `canceled = true` is executed
  3) `handlePromise` callback is executed

Correct me if I'm wrong.

ammarbinfaisal: oh! this does work. not sure why but I tried this in morning(kill before making canceled true) and this threw that error.

ammarbinfaisal:
sorry for troubling you so much over this PR.

index.js
ehmicky: You actually made me double check, I was not sure either whether `exit` would be fired on the next tick or not.

ehmicky:
You are helping us add a new feature, there's nothing to be sorry about! 

Code review is part of the process and I enjoy it, so no worries.

Hopefully this is a good experience for you as well, in case you wanted to contribute again after this PR :)

readme.md
```diff
@@ -53,8 +53,14 @@ const execa = require('execa');
 	//=> 'unicorns'
 
 	// Cancelling a spawned process
-	const spawned = execa("echo unicorns");
+	const spawned = execa("node");
 	spawned.cancel();
```
ehmicky: Should we put this in a `setTimeout(() => { spawned.cancel() }, 1000)` block? So that it's clear the process can be canceled after being `await`ed?

readme.md
```diff
@@ -149,6 +155,13 @@ Execute a command synchronously through the system shell.
 
 Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
 
+### spawned.cancel()
+
+Cancel a process spawned using execa. Calling this method kills it.
+
+Throws an error with `error.canceled` equal to `true`, provided that the process gets canceled.
+Process would not get canceled if it has already exited or has been killed by `spawned.kill()`.
```
ehmicky: I think you can just stop at `already exited` since if it has been killed, then it has exited.

ehmicky:
@sindresorhus the PR looks good to me, but awaiting on your own code review before merging.

index.js
```diff
@@ -334,7 +341,8 @@ module.exports = (command, args, options) => {
 			killed: false,
 			signal: null,
 			cmd: joinedCommand,
-			timedOut: false
+			timedOut: false,
+			canceled: false
```
sindresorhus: I would prefer to name it `isCanceled`. I realize it's inconsistent with `killed`, but I want to rename that one at some point too. Use the `isCanceled` naming for all the `canceled` variables.

index.js
```diff
@@ -347,6 +355,15 @@ module.exports = (command, args, options) => {
 	// eslint-disable-next-line promise/prefer-await-to-then
 	spawned.then = (onFulfilled, onRejected) => handlePromise().then(onFulfilled, onRejected);
 	spawned.catch = onRejected => handlePromise().catch(onRejected);
+	spawned.cancel = function () {
```
sindresorhus: Use arrow function

readme.md
```diff
@@ -52,6 +52,15 @@ const execa = require('execa');
 	const {stdout} = await execa.shell('echo unicorns');
 	//=> 'unicorns'
 
+	// Cancelling a spawned process
+	const spawned = execa("node");
```
sindresorhus: Single-quotes

sindresorhus: I think `process` would be a better name than `spawned`.

readme.md
```diff
@@ -146,6 +155,13 @@ Execute a command synchronously through the system shell.
 
 Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
 
+### spawned.cancel()
```
sindresorhus: It's not clear enough what `spawned` is. And it's weird to place the docs here. I think it would be better to document it in the `execa()` section.

test.js
```diff
@@ -586,3 +586,57 @@ if (Promise.prototype.finally) {
...
+	t.true(spawned.killed);
+});
+
+test('result.canceled is false when spawned.cancel isn\'t called', async t => {
+	const result = await execa('noop');
+	t.false(result.canceled);
+});
+
+test('calling cancel method throws an error with message "Command was canceled"', async t => {
+	const spawned = execa('noop');
```
```suggestion
	const process = execa('noop');
```

readme.md
ehmicky: Would this confuse readers about the global variable named `process`? Child processes returned by `spawn()` and global `process` have different methods/properties (although some are shared).

sindresorhus: Ah yeah, `subprocess` is better.

readme.md
```diff
@@ -121,6 +121,9 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
 
 It exposes an additional `.all` stream, with `stdout` and `stderr` interleaved.
 
+It can be canceled with `.cancel` method which throws an error with `error.canceled` equal to `true`, provided that the process gets canceled.
```
ehmicky: `canceled` -> `isCanceled()`

index.js
```diff
@@ -152,7 +152,7 @@ function makeError(result, options) {
 		error = new Error(message);
 	}
 
-	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCodeName, exitCode});
+	const prefix = getErrorPrefix({timedOut, timeout, signal, exitCode, exitCodeName, isCanceled});
```
sindresorhus: Don't do unrelated changes

sindresorhus:
Thank you for contributing, @ammarbinfaisal :)

index.js
ammarbinfaisal: I might have done this while fixing some merge conflict; that's the only time I touched those two variables `exitCode` and `exitCodeName`

PR #192, opened 2019-03-11
Add `execa.all()`

ehmicky:
This is a follow up on #171 by @tomsotte.

`options.all` was added but we forgot to also add `execa.all()`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -368,6 +368,11 @@ module.exports.stderr = async (...args) => {
 	return stderr;
 };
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
No, it was intentional: https://github.com/sindresorhus/execa/pull/171#discussion_r264036409

ehmicky:
Oh sorry did not see that one before. 

Should `execa.stdout()` and `execa.stderr()` be deprecated?

PR #197, opened 2019-03-30
Update dependencies, refactor TypeScript definition to CommonJS compatible export

BendingBender:
@sindresorhus This should be part of v2 milestone.

See [this issue](https://github.com/sindresorhus/mem/issues/31) for details.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,304 +2,307 @@
 import {ChildProcess} from 'child_process';
 import {Stream, Readable as ReadableStream} from 'stream';
 
-export type StdioOption =
-	| 'pipe'
-	| 'ipc'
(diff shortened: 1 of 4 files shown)


ehmicky:
The dependencies update looks good to me. I cannot judge for TypeScript.

sindresorhus:
> Also there is one test that fails on CI.

It's not new. I have to rerun the build one in a while. Seems the test is flaky on Node.js 8 on Windows. We should look into making it not flaky.

ehmicky:
Yes I noticed it as well before and was re-running the test on CI. However now I tried re-running several times and it kept failing. But it might just be luck. We should definitely look into that one test.

PR #199, opened 2019-04-03
Fix timeout option not leaving early when on `sleep` mode

### LINKED ISSUE #157: timeout option is not considered if streams are attached
Hello there,

I was fiddling with the timeout option. It seems like it doesn't quite work as I'd expect (any maybe you as well!)

## The setup
I have a bash script that sleeps for 6 seconds (sample.sh) and a nodejs file that is supposed to timeout after 200ms.

## What's happening
The nodejs file rejects as expected but not after 200ms but after 6seconds. This is weird.
I investigated a bit and the internal timeout to kill the process is reached. The process is killed as well.

The issue seems to be [this line](https://github.com/sindresorhus/execa/blob/master/index.js#L275); It looks like we are waiting for the stream of the killed process. It seems unusual since it will never talk again.
Removing the selected line (and the one after it) makes the process timeout nicely. The cleanup doesn't seem to be correct.

I've tried to work around by passing the options `stdout` and `stderr` and set each to null but that didn't really help.

Attached you'll find the sample project. Please let me know if something is missing. I might try to find a solution this evening.. ;)
[execa_timeout_broken.tar.gz](https://github.com/sindresorhus/execa/files/2575712/execa_timeout_broken.tar.gz)

GMartigny:
Fix #157
Add regression test

### DIFF
--- a/index.js
+++ b/index.js
@@ -267,9 +267,25 @@ const execa = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
+	const resolvable = (() => {
+		let extracted;
+		const promise = new Promise(resolve => {
+			extracted = resolve;
+		});
+		promise.resolve = extracted;
+		return promise;
+	})();
+
 	const processDone = new Promise(resolve => {
 		spawned.on('exit', (code, signal) => {
 			cleanup();
+
+			if (timedOut) {
+				resolvable.resolve([
+					{code, signal}, '', '', ''
+				]);
+			}
+
 			resolve({code, signal});
 		});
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -267,9 +267,25 @@ const execa = (command, args, options) => {
...
 		spawned.on('exit', (code, signal) => {
 			cleanup();
+
+			if (timedOut) {
+				resolvable.resolve([
+					{code, signal}, '', '', ''
+				]);
+			}
+
 			resolve({code, signal});
```
ehmicky: I am wondering whether the following would be simpler (did not try, this is just an example):

```js
	const timeoutPromise = new Promise(resolve => {
       if (parsed.options.timeout > 0) {
		    timeoutId = setTimeout(() => {
			    timeoutId = undefined;
			    timedOut = true;
                resolve([{ code: '1', signal: parsed.options.killSignal }, '', '', '']);
		        spawned.kill(parsed.options.killSignal);
		    }, parsed.options.timeout);
        }
	});

    // No changes in the code there
			cleanup();
			resolve({code, signal});
		});

		spawned.on('error', error => {
			cleanup();
			resolve({error});
		});

		if (spawned.stdin) {
			spawned.stdin.on('error', error => {
				cleanup();
				resolve({error});
			});
		}
	});

	const handlePromise = () => pFinally(Promise.race([
        timeoutPromise,
        Promise.all([
		    processDone,
	   	    getStream(spawned, 'stdout', {encoding, buffer, maxBuffer}),
		    getStream(spawned, 'stderr', {encoding, buffer, maxBuffer}),
		    getStream(spawned, 'all', {encoding, buffer, maxBuffer: maxBuffer * 2})
	    ]).then(results => { // eslint-disable-line promise/prefer-await-to-then
```

ehmicky:
Thanks a lot for the PR Guillaume! I've just reviewed it. Looking forward to merge this fix.

index.js
ehmicky: Thinking about it, it's probably a bad idea as the main promise will resolve before the child process has exited.

GMartigny: Yes, this works and look cleaner to me. One thing tho, I'm not a fan of never resolving promises, but I might be overly protective.

ehmicky: The bigger issue with my suggestion was that the main promise would be resolved before the child process has exited, so you can discard it :)

GMartigny:
As for the tests, they're working on local. I could remove the "[leaving early" test](https://github.com/sindresorhus/execa/pull/199/files#diff-1dd241c4cd3fd1dd89c570cee98b79ddR405). I thought that checking 2000ms over 500ms expected was safe enough. I'll try to extends the margin first, but I might have to remove this test (which remove the regression test).

ehmicky:
You can try to use `test.serial()` as well to ensure no other tests are running concurrently. This will ensure more predictable timing.

index.js
GMartigny: I can make a mix of both world with:

```js
// TODO: Use native "finally" syntax when targeting Node.js 10
const handlePromise = () => {
	const timedOutPromise = new Promise(resolve => {
		if (parsed.options.timeout > 0) {
			timeoutId = setTimeout(() => {
				timeoutId = undefined;
				timedOut = true;
				const {killSignal} = parsed.options;
				resolve([
					{code: 1, signal: killSignal},
					'',
					'',
					''
				]);
				spawned.kill(killSignal);
			}, parsed.options.timeout);
		}
	});

	return pFinally(Promise.race([
		timedOutPromise,
		Promise.all([
			processDone,
			getStream(spawned, 'stdout', {encoding, buffer, maxBuffer}),
			getStream(spawned, 'stderr', {encoding, buffer, maxBuffer}),
			getStream(spawned, 'all', {encoding, buffer, maxBuffer: maxBuffer * 2})
		])
	]).then(results => { // eslint-disable-line promise/prefer-await-to-then
    // ...
```

ehmicky: No the problem there is that `resolve()` gets called before the `exit` event gets emitted. I.e. `execa()` will resolve before the child process exited.

ehmicky:
This looks good to me. Just waiting for @sindresorhus review before merging it :)

Thanks! :clap:

index.js
```diff
@@ -301,51 +317,62 @@ const execa = (command, args, options) => {
...
+		if (timeoutId) {
+			processComplete = Promise.race([
+				processComplete,
+				resolvable
+			]);
+		}
 
-			if (!parsed.options.reject) {
-				return error;
+		return pFinally(processComplete.then(results => { // eslint-disable-line promise/prefer-await-to-then
```
sindresorhus: I would prefer using async/await here. If ESLint complains, it's usually a good reason for it. You could extract everything inside the `.then()` here into an async function that awaits `processComplete` first. Then you could just call it here: `pFinally(finalize(), destroy)`

index.js
```diff
@@ -267,9 +267,25 @@ const execa = (command, args, options) => {
 		}, parsed.options.timeout);
 	}
 
+	const resolvable = (() => {
```
sindresorhus: This is better known as a deferred, and is usually considered a bad practice: https://github.com/petkaantonov/bluebird/wiki/Promise-Anti-patterns#the-deferred-anti-pattern

I haven't looked closely into the changes here, but maybe there's a way to solve it without using a deferred? If not, that's fine too. Sometimes it's needed. But usually when someone grabs for a deferred, there's usually a better way.

index.js
GMartigny: FWIW, I didn't write this code, just moved it around. Your suggestion is clearer, so I've implemented it.

index.js
GMartigny: I think we fall into the "it's needed" case.
> You might have to use a deferred object when wrapping a callback API that doesn't follow the standard convention. Like setTimeout.

The goal is to resolve as soon as the process complete or the `setTimeout` ends. I could use some kind of cancellation, but I don't see any better solution right now.

sindresorhus: Yeah, I don't see a better way either. I was just trying my luck in case you could think of a better way.

sindresorhus:
This is looking good. Thanks for contribution, @GMartigny 👌

PR #200, opened 2019-04-03
Add `.node()` method

### LINKED ISSUE #65: Add `.fork()` method

GMartigny:
Fix #65


GMartigny:
This is a working draft to check if I'm going the right way.

As per https://github.com/sindresorhus/execa/issues/65#issuecomment-474761212, the fork method is just doing a spawn of a `node` process with some options.

ehmicky:
Thanks this new PR @GMartigny!

This is going the right direction except:
  - you should `execa(execPath, ...)` not `execa('node', ..., { execPath })`
  - changing `stdio.js` like this won't work because it's going to add this fourth channel to other methods as well, not only to `fork()`.
  - there are still few things missing (please check the issue, I think I've listed them). Please let me know once you've got those covered as well!

GMartigny:
Ok thanks @ehmicky for the help.

>changing stdio.js like this won't work because it's going to add this fourth channel to other methods as well

Is that a problem if that forth channel is only ever set to `undefined` on other cases ?

> use `process.execPath` ...

`process.execPath` should be the running file, execa main file or something else ?

> use `process.execArgv` ...

Should we redirect **all** node arguments to this option ?

ehmicky:
> Is that a problem if that forth channel is only ever set to undefined on other cases ?

Unless I am mistaken, this seems like at the moment, `stdio: "inherit"` would translate to `stdio: ["inherit", "inherit", "inherit", "ipc"]` to both `fork()` (which is good) and normal `execa()` (which is not wanted).

This could be solved by doing the `stdio` normalization logic before calling `execa()` inside `fork()`.

> process.execPath should be the running file, execa main file or something else ?

Yes `process.execPath` is the currently run `node` binary. I.e. it is the main file spawned. 

> Should we redirect all node arguments to this option ?

We need to pass `process.execArgv` down to the child process. Now the core Node.js `fork()` method also allows overidding with `opts.execArgv`. Users might expect the same options as the core Node.js method, i.e. it might be a good idea to add this.

GMartigny:
From what I read on the node's source code, you can call `fork` with one `'ipc'` in any channel of `stdio`. Should we comply with this behavior ? This would add more complexity to options.

Also, there are (IMO) arbitrary choices over options hierarchy. For example, an `array` `stdio` ignores the `silent` option.

Here's what make sens to me, but I need confirmation:

```
options => effective result

1 {}                      => ['inherit', 'inherit', 'inherit', 'ipc']    // default to 'inherit'
2 {stdio: 'pipe'}         => ['pipe', 'pipe', 'pipe', 'ipc']             // Can be defined
3 {stdio: 'whatever'}     => ['whatever', 'whatever', 'whatever', 'ipc'] // Let user fail
4 {stdio: [0, 1, 2]}      => [0, 1, 2, 'ipc']                            // Keep stdio arrays as is
4b {stdio: [0, 1, 2, 3]}  => [0, 1, 2, 'ipc']                            // But ignores overflow
5 {stdio: [0, 'ipc', 2]}  => [0, 'ipc', 2]                               // According to node's code

6 {stdout: 'any'} => ['inherit', 'any', 'inherit', 'ipc'] // Can defined individual channel
7 {stdout: 'ipc'} => ['inherit', 'ipc', 'inherit']        // According to node's code

8 {stdio: 'any', silent: true}     => ['pipe', 'pipe', 'pipe', 'ipc'] // Silent take precedence
9 {stdio: [0, 1, 2], silent: true} => ['pipe', 'pipe', 'pipe', 'ipc'] // Silent take precedence
10 {stdout: 'any', silent: true}   => ['pipe', 'pipe', 'pipe', 'ipc'] // Silent take precedence
```

Lines 3, 4, 4b and 9 don't follow what I read from Node's doc or code.

ehmicky:
All that logic is already handled by both `execa` and Node.js `child_process` module. The only thing that's needed here is IMHO, inside `fork()`, before calling main `execa()` method:
  1) if only one string was passed, expand to an array of three repeated strings (already done by `stdio.js`, just need to call it and maybe tweak it)
  2) if an array was passed, set `array[3]` to `ipc`

This is in essence what the Node.js code for `fork()` seems to do there as well.

GMartigny:
> if an array was passed, set array[3] to ipc
We don't have to restrict the `ipc` channel to the 3rd one ? I don't know why, but why not let the user define it's `stdio` and pushing `ipc` to the last available position.

If I push Unit Tests for that new function, could you review them ? I feel it would be better going TDD style.

ehmicky:
Sure anyway works better for you :)

Yes you're right about `ipc`. For reference, this is [the Node.js core code for this](https://github.com/nodejs/node/blob/master/lib/child_process.js#L100).

test/stdio.js
```diff
@@ -4,47 +4,67 @@ import stdio from '../lib/stdio';
...
+
+test(stdioMacro, {stdio: {foo: 'bar'}}, new TypeError('Expected `stdio` to be of type `string` or `Array`, got `object`'));
+
+test(stdioMacro, {stdin: 'inherit', stdio: 'pipe'}, new Error('It\'s not possible to provide `stdio` in combination with one of `stdin`, `stdout`, `stderr`'));
+test(stdioMacro, {stdin: 'inherit', stdio: ['pipe']}, new Error('It\'s not possible to provide `stdio` in combination with one of `stdin`, `stdout`, `stderr`'));
+test(stdioMacro, {stdin: 'inherit', stdio: [undefined, 'pipe']}, new Error('It\'s not possible to provide `stdio` in combination with one of `stdin`, `stdout`, `stderr`'));
+
+const forkMacro = createMacro(stdio.fork);
+
+test(forkMacro, undefined, ['pipe', 'pipe', 'pipe', 'ipc']);
```
ehmicky: `fork()` `stdio` should default to `inherit` when the `silent` option is `false`, and to `pipe` when it is `true`.

The `silent` option should default to `false. 

See https://github.com/nodejs/node/blob/master/lib/child_process.js#L102

test/stdio.js
```diff
@@ -4,47 +4,67 @@ import stdio from '../lib/stdio';
...
+test(forkMacro, {stdio: 'ignore'}, ['ignore', 'ignore', 'ignore', 'ipc']);
+test(forkMacro, {stdio: [0, 1, 2]}, [0, 1, 2, 'ipc']);
+test(forkMacro, {stdio: [0, 1, 2, 3]}, [0, 1, 2, 3, 'ipc']);
```
ehmicky: Those two tests should throw. See https://github.com/nodejs/node/blob/master/lib/child_process.js#L107

test/stdio.js
```diff
@@ -4,47 +4,67 @@ import stdio from '../lib/stdio';
...
+
+test(forkMacro, {stdout: 'ignore'}, ['pipe', 'ignore', 'pipe', 'ipc']);
```
ehmicky: I think you must have mistyped that test, because the input is the same as one of the tests above, but the expected value is different.

test/stdio.js
```diff
@@ -4,47 +4,67 @@ import stdio from '../lib/stdio';
...
+test(forkMacro, {stdout: 'ignore', stderr: 'ignore'}, ['pipe', 'ignore', 'ignore', 'ipc']);
+
+test(forkMacro, {silent: true, stdio: 'ignore'}, ['pipe', 'pipe', 'pipe', 'ipc']);
+test(forkMacro, {silent: true, stdio: [0, 1, 2, 42, 5]}, ['pipe', 'pipe', 'pipe', 'ipc']);
```
ehmicky: I think `silent: true` should only have an effect when the `stdio` option is `undefined`. This follows Node.js core code: https://github.com/nodejs/node/blob/master/lib/child_process.js#L100

ehmicky:
The tests looks good :+1:

lib/stdio.js
```diff
@@ -39,3 +39,17 @@ module.exports = opts => {
 
 	return result;
 };
+
+module.exports = stdio;
+
+module.exports.fork = opts => {
```
GMartigny: This might need a better name.

index.js
```diff
@@ -427,3 +427,19 @@ module.exports.sync = (command, args, options) => {
 };
 
 module.exports.shellSync = (command, options) => handleShell(execa.sync, command, options);
+
+module.exports.fork = (filePath, args, options) => {
+	const stdioOption = stdio.fork(options);
+	delete options.stdin;
+	delete options.stdout;
+	delete options.stderr;
```
ehmicky: I think we should avoid mutating caller's arguments. This is a side-effect the caller might not expect.

For example, let's say you're calling `fork()` twice in a row using the same `options` object. The second call will get different `options` with the current code.

index.js
```diff
@@ -427,3 +427,19 @@ module.exports.sync = (command, args, options) => {
...
+
+	return execa(filePath, args, {
+		...options,
+		code: 0,
```
ehmicky: There is no `code` option in execa.

GMartigny: This was a test I commit by mistake, it's removed on my local directory.

fixtures/plain.js
```diff
@@ -0,0 +1 @@
+console.log('Javascript');
```
ehmicky: You could also use the fixture `noop` that already exists.

GMartigny: Yes, but I wanted to try that fork use the same process as the parent, ie `node`. `noop` is a bash script.

ehmicky: `noop` is not a Bash script, it's a Node.js file with a shabang. Maybe we're talking about a different file: https://github.com/sindresorhus/execa/blob/master/fixtures/noop

Fork will never use the same process as the parent, it will always run in a child process.

GMartigny:
Hi, tests are failing on Windows because:
```js
// This will be run on a fork of Node, which can't work
execa.fork("fixtures/hello.cmd"); // => node fixtures/hello.cmd

// Will spawn a new cmd that don't exit itself
execa.fork("fixtures\\hello.cmd", [], {execPath: "cmd.exe"}); // => cmd.exe fixtures\hello.cmd

// Same as
execa("cmd.exe", ["fixtures\\hello.cmd"]);
```

I saw that `node-cross-spawn` add [a bunch of options](https://github.com/moxystudio/node-cross-spawn/blob/48bee1dd5cf06f8162cbbce812b5c6e01b1c439d/lib/parse.js#L62) to `cmd`. But only on certain cases that we don't fulfill here.

Should we add the options ourselves or am I missing something ?

ehmicky:
I don't think the `fork()` method is meant to fire shell scripts (neither `cmd.exe` nor Bash) but only Node.js files, since it uses `node` as the main binary file.

sindresorhus:
@GMartigny Would you be able to finish this? :)

GMartigny:
Hi @sindresorhus, sorry I kind of forget about this.
I stop here because I don't know how to deal with using `fork` with another process than node using the `execPath` option.

> the two first commands above should throw an exception

Ok, but I got no idea how to correctly detect that `whatever.exe` is not the node executable. As usual, we can let the user fail or remove the `execPath` option (so it's always `process.execPath`).

ehmicky:
@GMartigny Sorry I think I misunderstood your last comments. 

The approach you took to test `execPath` looks good to me: passing either `cmd.exe` or `bash` as `options.execPath` then firing a shell script. `fork()` is not meant to be used with shell scripts, but this is good for the purpose of testing `options.execPath` works.

sindresorhus:
You also need to update https://github.com/sindresorhus/execa/blob/master/index.d.ts and https://github.com/sindresorhus/execa/blob/master/index.test-d.ts

GMartigny:
I don't know TypeScript (yet), so I hope copy/pasting was enough for this case.

index.js
```diff
@@ -473,3 +473,24 @@ module.exports.sync = (command, args, options) => {
...
 	};
 };
+
+module.exports.fork = (filePath, args = [], options = {}) => {
+	const stdioOption = stdio.fork(options);
+
+	return execa(
+		options.execPath || process.execPath,
+		[
+			path.normalize(filePath),
```
ehmicky: Is `path.normalize()` needed here? Core Node.js does not do it, so I am assuming Node.js `child_process` methods will handle the path correctly even if not normalized. Same thing with `execa()`.

index.js
```diff
@@ -95,7 +95,7 @@ function handleArgs(command, args, options = {}) {
 
 	if (process.platform === 'win32' && path.basename(command, '.exe') === 'cmd') {
 		// #116
-		args.unshift('/q');
+		args = [...new Set(['/q', '/d', '/s', '/c', ...args])];
```
ehmicky: Why is this needed: does `cmd.exe` complain on duplicate CLI arguments?

Also this probably should be in a separate PR.

GMartigny: `node-cross-spawn` add those options when in `shell` mode and when the command file is not a ".exe". Or fork method runs `cmd.exe` in non-shell mode. Without those options, the cmd script will never quit and pipe its output.
The order is important, that's why I added a non-duplicate safety here.

ehmicky: I think you might have used the French "Or" there :)

You're speaking about [those lines](https://github.com/moxystudio/node-cross-spawn/blob/48bee1dd5cf06f8162cbbce812b5c6e01b1c439d/lib/parse.js#L45), right? Also just for reference those lines just emulate what [Node.js does with `shell: true`](https://github.com/nodejs/node/blob/6f2bd9dcc3a9ed6cf509cdcd381c1c456f0bd3cb/lib/child_process.js#L483).

I might be wrong here, but it seems to me that when `shell: false` (like for `fork()`):
  - if the command is `*.exe` or `*.cmd`, `node-cross-spawn` does not touch it.
  - otherwise, it calls the command with `cmd.exe /d /s /c` (like `shell: true` would).

Here the `fork` method will run `node` (not `cmd.exe`) in non-shell mode, so `node-cross-spawn` should not touch it. Or am I missing something?

GMartigny: > I think you might have used the French "Or" there :)

I'm loosing my mind apparently -__-

> You're speaking about those lines

Yes. When `fork` runs `cmd.exe` through the `execPath` option, it should adds those 3 options too. I'm still not sure if `fork` is intended to run any other executable than `node`.

ehmicky: As far as I understand it, `fork` is not meant to run `cmd.exe` through the `execPath` option.

Now would the problem you're describing still happen when doing `execa('cmd.exe', ['file.exe'], { shell: false })` (which is what `fork` would eventually call)?

GMartigny: I just tried and yes, the example you provide fail on Windows. The script never quit itself and block the thread.

ehmicky: This would be a bug in the current code, not related to this PR (since it can be reproduced without `fork()`). Do you think you could open a separate issue about this, with a reproducible example? I could take a stab at fixing it.

GMartigny: https://github.com/sindresorhus/execa/issues/232

ehmicky:
CI tests do not complete on Windows. However I've fixed few things there, so if you rebase on `master`, this might solve that problem.

GMartigny:
Even though tests succeed, I had a couple of issues with tests on local that another user or contributer might encounter:
 - Major: My Node.js is installed under `C:/Program Files/node/node.exe`.
Note the space inside `Program Files`, this throw a `Arguments cannot be inside command when also specified as an array of strings` error in execa when using `fork` because `process.execPath` (the default value) is the full node path.
 - Minor: My system is in French, so the error message thrown when using a wrong command is different `execa('wrong command')`.

ehmicky:
I've opened a PR #261 addressing your first issue with `C:\Program Files\node\node.exe`. The problem is that `execa` now allows the command and its arguments to specified as a single string, without the `shell` option: `execa('hello world')`. In your case, `execa` thought that `C:\Program` was the command and `Files\node\node.exe` the first argument. With #261, it will now correctly identify `C:\Program Files\node\node.exe` as the full command, providing the `args` array is defined (which is the case in your implementation of `fork()`.

lib/stdio.js
```diff
@@ -39,3 +39,23 @@ module.exports = opts => {
...
+
+	let stdioOption = stdio(opts || {stdio: defaultOption});
+
+	if (typeof stdioOption === 'string') {
+		stdioOption = [...new Array(3)].fill(stdioOption);
+	} else if (Array.isArray(stdioOption)) {
+		stdioOption = stdioOption.map(channel => channel === undefined ? defaultOption : channel);
+	}
+
+	if (!stdioOption.includes('ipc')) {
```
ehmicky: [Node.js code](https://github.com/nodejs/node/blob/0ca9297c3945f52325fcd08afe3fb9dfe3adf123/lib/internal/child_process.js#L228) pushes `ipc` as the last element, even if it's already present.

lib/stdio.js
```diff
@@ -39,3 +39,23 @@ module.exports = opts => {
...
+
+module.exports.fork = opts => {
+	const defaultOption = (opts && opts.silent && !opts.stdio) ? 'pipe' : 'inherit';
+
+	let stdioOption = stdio(opts || {stdio: defaultOption});
+
+	if (typeof stdioOption === 'string') {
+		stdioOption = [...new Array(3)].fill(stdioOption);
+	} else if (Array.isArray(stdioOption)) {
+		stdioOption = stdioOption.map(channel => channel === undefined ? defaultOption : channel);
```
ehmicky:
```suggestion
		stdioOption = stdioOption.map((channel = defaultOption) => channel);
```

readme.md
```diff
@@ -146,6 +146,15 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
...
 Stream combining/interleaving [`stdout`](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout) and [`stderr`](https://nodejs.org/api/child_process.html#child_process_subprocess_stderr).
 
+### execa.fork(file, [arguments], [options])
+
+Run a file through a forked process. The default sub-process is the current (node).
+
+It introduce a communication channel (IPC) that allows messages to be passed back and forth between the parent and child.<br>
+Note that the `shell` option will be ignored.
+
+Returns the same object as [execa](#execafile-arguments-options).
```
ehmicky: I find the explanation a little confusing:
  - For example this sentence: `The default sub-process is the current (node).`
  - the `IPC` channel is available to `execa()` as well, not just by default.

What about the following?

```
Same as `execa('node', [file, ...arguments], options)` except (like [`child_process#fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options)):
  - the `execPath`, `execArgv` and `silent` options can be used
  - the [`shell`](#shell) option cannot be used 
  - the [`stdio`](#stdio)-related options defaults to `inherit` instead of `pipe`
```

readme.md
```diff
@@ -405,6 +414,26 @@ Default: `false`
 
 If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
+#### execPath (for `fork` only)
+
+Type: `string`<br>
+Default: `process.execPath`
+
+Define the sub-process executable binary.
```
ehmicky: Let's re-use the Node.js documentation: `Executable used to create the child process.`

readme.md
```diff
@@ -405,6 +414,26 @@ Default: `false`
...
+
+#### execArgv (for `fork` only)
+
+Type: `string[]`<br>
+Default: `process.execArgv`
+
+Define the sub-process arguments.
```
ehmicky: Let's re-use the Node.js documentation: `List of string arguments passed to the executable.`

readme.md
```diff
@@ -405,6 +414,26 @@ Default: `false`
...
+
+#### silent (for `fork` only)
+
+Type: `boolean`<br>
+Default: `false`
+
+If `true`, set all stdio channel to `'pipe'`.
```
ehmicky:
```suggestion
If `true`, set all stdio channels to `'pipe'`.
```

test.js
```diff
@@ -706,3 +706,49 @@ test('calling cancel method on a process which has been killed does not make err
 	const {isCanceled} = await t.throwsAsync(subprocess);
 	t.false(isCanceled);
 });
+
+test.cb('fork()', t => {
+	const fork = execa.fork('fixtures/noop');
+
+	fork.on('close', code => {
+		t.is(code, 0);
```
ehmicky: We removed the `code` property.

test.js
```diff
@@ -706,3 +706,49 @@ test('calling cancel method on a process which has been killed does not make err
 	const {isCanceled} = await t.throwsAsync(subprocess);
 	t.false(isCanceled);
 });
+
+test.cb('fork()', t => {
+	const fork = execa.fork('fixtures/noop');
```
ehmicky: Would the following work instead?

```js
test('fork()', async t => {
  const {exitCode} = await execa.fork('fixtures/noop');
  t.is(exitCode, 0); 
})
```

test.js
```diff
@@ -706,3 +706,49 @@ test('calling cancel method on a process which has been killed does not make err
...
+test('fork pipe stdout', async t => {
+	const {stdout} = await execa.fork('fixtures/noop', ['foo'], {
+		stdout: 'pipe'
+	});
+
+	t.is(stdout, 'foo');
+});
+
+test('fork correctly use execPath', async t => {
+	const result = await execa.fork(process.platform === 'win32' ? '/c hello.cmd' : 'hello.sh', [], {
```
ehmicky: Please use destructuring: `const {stdout} = ...`

Same for the other test below.

ehmicky: I think `/c` should be added as `execArgv` instead.

index.d.ts
```diff
@@ -185,14 +185,35 @@ declare namespace execa {
...
+		/**
+		Define the sub-process arguments
+
+		@default process.execArgv
+		 */
+		readonly execArgv?: string[];
+
+		/**
+		If `true`, set all stdio channel to `'pipe'`
+		 */
```
ehmicky: You should add the default value `false`.

lib/stdio.js
GMartigny: True only if [`options.stdio` is not an array](https://github.com/nodejs/node/blob/master/lib/child_process.js#L97)

ehmicky: No it's not, check the lines above:

```js
  if (typeof options.stdio === 'string') {
    options.stdio = stdioStringToArray(options.stdio, 'ipc');
```

GMartigny: In my case, when the user define a string, `stdio.fork` append an `ipc` channel at the end of length 3 array. Check [tests](https://github.com/sindresorhus/execa/pull/200/files#diff-ddb2f51fa9f852e91d150fdec4404ba8R60) for more example.

readme.md
GMartigny: It was copied from the [node API documentation](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options), but your version looks good too.

ehmicky:
Could you please keep the comments unresolved until new code is being pushed? This helps me go through my former comments, and resolve them one by one.

readme.md
ehmicky: Just for clarification, only the IPC sentence was.

test.js
GMartigny: `code` was removed from the result object return by `execa`, but listening to event still return the appropriate code.
Anyway, both solutions works, but I prefer yours since it match with what's already done.

ehmicky: See #250. We removed it in tests too.

readme.md
```diff
@@ -150,8 +150,10 @@ Stream combining/interleaving [`stdout`](https://nodejs.org/api/child_process.ht
 
 Run a file through a forked process. The default sub-process is the current (node).
 
-It introduce a communication channel (IPC) that allows messages to be passed back and forth between the parent and child.<br>
-Note that the `shell` option will be ignored.
+Same as `execa('node', [file, ...arguments], options)` except (like [`child_process#fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options)):
+  - the `execPath`, `execArgv` and `silent` options can be used
+  - the [`shell`](#shell) option cannot be used 
+  - the [`stdio`](#stdio)-related options defaults to `inherit` instead of `pipe`
```
ehmicky: We should also add:

```
- an extra channel [`ipc`](https://nodejs.org/api/child_process.html#child_process_options_stdio) is passed to [`stdio`](#stdio)
```

Also the first two sentences (`Run a file... current (node)`) and the last sentence (`Returns the same object`) can be removed as they are implied by those new sentences.

index.d.ts
```diff
@@ -185,14 +185,37 @@ declare namespace execa {
...
 		/**
 		Write some input to the `stdin` of your binary.
 		*/
 		readonly input?: string | Buffer;
 	}
 
+	interface ForkOptions<EncodingType = string> extends CommonOptions<EncodingType> {
+
+		/**
+	 	Define the sub-process executable binary
```
ehmicky: Those comments should be the same as the `README`

index.d.ts
```diff
@@ -368,6 +391,26 @@ declare const execa: {
 		file: string,
 		options?: execa.SyncOptions<null>
 	): execa.ExecaSyncReturnValue<Buffer>;
+
+	/**
+	 Run a file through a fork of the current process.
```
ehmicky: This comment should be the same as in the `README`.

readme.md
GMartigny: For the last sentence I agree, but I think it's nice to have a little "prosaic" description of the function as introduction.

ehmicky: Sure about the first sentence. However I find the second sentence `The default sub-process is the current (node).` a little confusing. Also the next sentence `Same as execa('node', ...)` should already indicate that the default child process is `node`.

GMartigny: Alright !

ehmicky:
Adding a comment here because GitHub does not allow commenting on binary files: the `media/logo.sketch` binary file was changed accidentally.

ehmicky:
The only thing left would be [this comment](https://github.com/sindresorhus/execa/pull/200#discussion_r286007660), but otherwise looks good to me!

GMartigny:
> the `media/logo.sketch` binary file was changed by accident.

You're right my mistake, but it's some weird git shenanigan. I'm unable to revert it and git said there's no change in the file. Even cloning the repo again doesn't help. 🤔 

> We should allow `execa.fork(filePath, options)`

Missed that, done.

ehmicky:
Except for the issue with `media/logo.sketch`, this looks ready for another code review from @sindresorhus  :tada: :tada: 

Thanks for your work on this @GMartigny!

sindresorhus:
> I'm unable to revert it and git said there's no change in the file.

Did you try to properly revert the file change from the command-line? I don't think Git would prevent that. https://www.google.com/search?q=git+revert+file+change&oq=git+revert+file+change&aqs=chrome..69i57j0j69i60j0l3.3699j0j1&sourceid=chrome&ie=UTF-8

readme.md
```diff
@@ -405,6 +415,26 @@ Default: `false`
...
+
+Type: `string[]`<br>
+Default: `process.execArgv`
+
+List of string arguments passed to the executable.
+
+#### silent (for `fork` only)
+
+Type: `boolean`<br>
+Default: `false`
```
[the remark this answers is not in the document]
ehmicky: Yes defaulting to `pipe` is better. I.e. default should be `true` even though it's different from Node.js

readme.md
```diff
@@ -146,6 +146,16 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
 
 Stream combining/interleaving [`stdout`](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout) and [`stderr`](https://nodejs.org/api/child_process.html#child_process_subprocess_stderr).
 
+### execa.fork(file, [arguments], [options])
+
+Run a file through a forked process.
```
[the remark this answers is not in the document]
GMartigny: Could you or @ehmicky give a better description ?

readme.md
```diff
@@ -432,6 +441,19 @@ Default: `false`
...
+Default: [`process.execPath`](https://nodejs.org/api/process.html#process_process_execpath)
+
+Node.js executable used to create the child process.
+
+#### execArgv *(for `.node()` only)*
+
+Type: `string[]`<br>
+Default: [`process.execArgv`](https://nodejs.org/api/process.html#process_process_execargv)
+
+List of string arguments passed to the node.js executable.
```
ehmicky:
```suggestion
List of string arguments passed to the Node.js executable.
```

test/fixtures/send.js
```diff
@@ -0,0 +1,7 @@
+process.on('message', message => {
```
ehmicky: This should have a shebang and `use strict`

test/node.js
```diff
@@ -0,0 +1,46 @@
+import test from 'ava';
+import pEvent from 'p-event';
+import path from 'path';
+import execa from '..';
+
+process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
+
+test('node()', async t => {
+	const {exitCode} = await execa.node('test/fixtures/noop');
```
ehmicky: You should not need `test/fixtures/` because of the `PATH` fix above?

GMartigny: I would agree with you, but apparently with the modified path running `noop` works (instead of `./test/fixtures/noop`) but running `node noop` doesn't work (while `node ./test/fixtures/noop` does work).
I don't know how Node resolves file execution as parameter, there might be another way.

ehmicky: Oh yes you're right, `PATH` is only used for the executable file, not its arguments, and here `noop` is actually the argument.

readme.md
```diff
@@ -180,6 +180,15 @@ Same as [`execa.command()`](#execacommand-command-options) but synchronous.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### execa.node(file, [arguments], [options])
+
+Run a file through a forked process.
+
+Same as `execa('node', [file, ...arguments], options)` except (like [`child_process#fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options)):
+  - the `execPath` and `execArgv` options can be used
```
ehmicky: Each option should link to its definition below.

readme.md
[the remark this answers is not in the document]
ehmicky: `Execute a Node.js file as a child process`?

index.d.ts
```diff
@@ -185,14 +185,29 @@ declare namespace execa {
...
+		 @default process.execPath
+		 */
+		readonly execPath?: string;
+
+		/**
+		 List of string arguments passed to the executable.
+
+		 @default process.execArgv
+		 */
+		readonly execArgv?: string[];
```
```suggestion
		readonly nodeArguments?: string[];
```

index.d.ts
```diff
@@ -185,14 +185,29 @@ declare namespace execa {
...
+	interface NodeOptions<EncodingType = string> extends CommonOptions<EncodingType> {
+		/**
+		 Executable used to create the child process.
+
+		 @default process.execPath
+		 */
+		readonly execPath?: string;
+
+		/**
+		 List of string arguments passed to the executable.
```
```suggestion
		 List of string arguments passed to the Node.js executable.
```

test/fixtures/detach
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+'use strict';
+
+const execa = require('../../');
```
sindresorhus:
```suggestion
const execa = require('../..');
```

test/node.js
```diff
@@ -0,0 +1,46 @@
...
+
+test('node pipe stdout', async t => {
+	const {stdout} = await execa.node('test/fixtures/noop', ['foo'], {
+		stdout: 'pipe'
+	});
+
+	t.is(stdout, 'foo');
+});
+
+test('node correctly use execPath', async t => {
```
sindresorhus: This is not correct use of those options. They should only be used to set a different Node.js executable. And `exec.node(..)` should only execute Node.js compatible scripts.

readme.md
```diff
@@ -180,6 +180,15 @@ Same as [`execa.command()`](#execacommand-command-options) but synchronous.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### execa.node(file, [arguments], [options])
+
+Run a file through a forked process.
```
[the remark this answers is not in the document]
ehmicky: Related to https://github.com/sindresorhus/execa/pull/200#discussion_r291545470

test/node.js
GMartigny: Is there a way to check that the `execPath` is a valid Node.js executable ? There's so much difference between Node.js version/fork that I doubt we can find any common ground.
If the user bend `.node()` to run something else than Node.js it doesn't do any harm ?
In any case, I don't see any way to have a different Node.js executable path on CI.

index.d.ts
```diff
@@ -185,14 +185,29 @@ declare namespace execa {
...
 		readonly input?: string | Buffer;
 	}
 
+	interface NodeOptions<EncodingType = string> extends CommonOptions<EncodingType> {
+		/**
+		 Executable used to create the child process.
+
+		 @default process.execPath
+		 */
+		readonly execPath?: string;
```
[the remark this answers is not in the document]
GMartigny: Hi @ehmicky, does it still make sense to [link to the Node.js docs](https://github.com/sindresorhus/execa/pull/200#discussion_r291544074) if we change the name of this param (same for `execArgv` => `nodeArguments` below) ? Would this be just confusing ?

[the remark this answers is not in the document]
ehmicky: Yes you're right, we should remove the link.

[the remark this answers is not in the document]
GMartigny: After all, removing the links felt worse ... my bad.
You can check [the result live](https://github.com/GMartigny/execa/tree/addForkMethod#execanodefile-arguments-options).

index.d.ts
```diff
@@ -437,12 +437,12 @@ declare const execa: {
 	commandSync(command: string, options?: execa.SyncOptions<null>): execa.ExecaSyncReturnValue<Buffer>;
 
 	/**
-	 Run a file through a forked process.
+	Execute a Node.js file as a child process.
```
sindresorhus: `file` => `script` for consistency with the argument.

readme.md
```diff
@@ -180,6 +180,15 @@ Same as [`execa.command()`](#execacommand-command-options) but synchronous.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### execa.node(file, [arguments], [options])
```
sindresorhus: This should be in sync with the TS definition.
```suggestion
### execa.node(scriptPath, [arguments], [options])
```

sindresorhus:
Only a few minor tweaks and this looks good to me :)

You also need to fix the merge conflict.

test/node.js
sindresorhus: I guess it's fine to keep it like this for testing, as long as we don't document or recommend using it like this.

sindresorhus:
Yay! Thanks for working on this :)

PR #201, opened 2019-04-04
Add a logo

ehmicky:
This adds a logo I just made:

execa_white

The green parts are representing the `>_` symbols often used to represent Unix shell prompts.

Green-on-black is a common color scheme for terminals. That shade of green is also one of the Node.js official shades of green.

The font is [ISL Fade to Blak](https://www.dafont.com/isl-fade-to-blak.font) and is in the public domain.

### DIFF
--- a/logo.svg
+++ b/logo.svg
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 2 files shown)


sindresorhus:
The logo looks really good! 🙌❤️

ehmicky:
Thanks! I am not a designer but creating logos is fun :)

ehmicky:
By the way I think if some reason we need an icon (as opposed to a full logo) the "x" letter can be used as it contains the `>` shell and is recognizable on its own.

sindresorhus:
> I am not a designer but creating logos is fun :)

👍 Ditto :)

PR #203, opened 2019-04-08
Add `/q` option to `cmd` as well as `cmd.exe` on Windows

GMartigny:
The `/q` option should be added whether the running command is `cmd.exe` or `cmd`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,7 +60,8 @@ function handleArgs(command, args, options) {
 		options.cleanup = false;
 	}
 
-	if (process.platform === 'win32' && path.basename(command).startsWith('cmd')) {
+	const basename = path.basename(command);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -60,7 +60,8 @@ function handleArgs(command, args, options) {
 		options.cleanup = false;
 	}
 
-	if (process.platform === 'win32' && path.basename(command) === 'cmd.exe') {
+	const basename = path.basename(command);
```
ehmicky:
```suggestion
	const basename = path.basename(command, '.exe');
```

Then the next line does not need to be changed.

sindresorhus:
Can you give the PR a more descriptive title?

ehmicky:
This looks good to me. Thanks for this PR!

PR #208, opened 2019-04-28
Add feature to retry process killing after specified duration

### LINKED ISSUE #105: When killing a process, ensure it was killed
When a user does `child.kill()` (NOT when `child.kill('SIGKILL')`), after some timeout (How long?), we should check if the process was actually killed with `process.kill(0)` and do `process.kill('SIGKILL')` if not.

zokker13:
Hi,

this PR fixes #105.
I'm wondering if the approach is good. If so, I'll add tests.

Personally, I'm not sure about the options parameter. It seems odd to be forced pass a parameter for the timeout but at the same time it's fine since there's a default value in the code.


index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
...
+	spawned.kill = (signal, options) => {
+		if (signal === undefined || signal === 'SIGTERM') {
+			const retryAfter = options !== undefined &&
+				['number', 'string'].includes(typeof options.retryAfter) ?
+				Number(options.retryAfter) :
+				2000;
+
+			setTimeout(() => {
+				try {
+					originalKill(0);
```
ehmicky: This is not needed, as `process.kill('SIGKILL')` is a noop if process is already killed.
See https://github.com/nodejs/node/blob/f5e2ad784ff8d51db275a7268660fc72cf58289d/lib/internal/child_process.js#L435

index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options) => {
+		if (signal === undefined || signal === 'SIGTERM') {
```
ehmicky: `signal` should be case-insensitive.

index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
...
+			const retryAfter = options !== undefined &&
```
ehmicky: Instead of an `options` object, wouldn't passing the argument directly as a number/string be simpler?

index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
...
+				['number', 'string'].includes(typeof options.retryAfter) ?
+				Number(options.retryAfter) :
+				2000;
```
ehmicky: 2 seconds might be quite fast for some users. Would 5 seconds be better?

index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
...
+
+			setTimeout(() => {
```
ehmicky: `setTimeout()` should only be called after `originalKill()` has returned `true`. This is in case `signal` is erroneous (which will throw an error) or there is some permissions error (which will emit an `error` event). In both cases, we don't want to repeat `originalKill()`.

index.js
```diff
@@ -237,6 +237,28 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options) => {
+		if (signal === undefined || signal === 'SIGTERM') {
+			const retryAfter = options !== undefined &&
+				['number', 'string'].includes(typeof options.retryAfter) ?
```
ehmicky: There should be a way for users to disable this timeout. At the moment, this could be done by passing `Infinity`, but I think passing `false` might be more proper.

index.js
ehmicky: I think `const timeout = setTimeout(...); timeout.unref()` should be called:
  - if the process was already killed, we don't want to hold the process from exiting.
  - if it was not killed, the child process will hold the process from exiting, so no need for the timeout to do it as well.

index.js
ehmicky: `signal` can also be an integer. That integer is OS-specific (`os.constants.signals.SIGTERM`).

ehmicky: There are other signals that signal termination, not only `SIGKILL` (the default value of `process.kill()`). You can get them from `signalExit.signals()` (`signal-exit` is already a dependency).

See https://github.com/tapjs/signal-exit/blob/master/signals.js

index.js
zokker13: Yes, good point!
46ee043aa907ef050c0d5eda4d0dc8ff31600cca

index.js
zokker13: Yup, done: 8f590cb48d61837c1ba4572f743a92decdfa4c01

zokker13: I added this possibility now: 
8f590cb48d61837c1ba4572f743a92decdfa4c01

zokker13: Mh... I wonder if we should also consider those. The thing is that the default implementation for kill [defaults](https://github.com/nodejs/node/blob/f5e2ad784ff8d51db275a7268660fc72cf58289d/lib/internal/child_process.js#L438) to SIGTERM. So to keep the API clean the thought was that people that don't care about the _killing_ usage gain benefit while people that perform weird signal passing don't kill their process by accident.
Having only a condition on SIGTERM ensures that the purpose of calling `kill` was to destroy the existence of the process in a nice way.

index.js
zokker13: Easier for sure. But I'd also say more confusing. The API for the normal [kill](https://nodejs.org/dist/latest-v10.x/docs/api/process.html#process_process_kill_pid_signal) has the signal and an integer as well - just reversed. So people might be thinking it's the same behavior.
I'm also a huge fan of options parameters in general since it let's you update your API without getting hated by your userbase.

index.js
zokker13: You might be right, I'm not really sure. It's a kill command after all - 2 seconds seems pretty long for that to finish imo. But I changed it: af15667e3d995374b2f22a8b8fc8bbf7d8418152

index.js
zokker13: I added another flag `retry` which can be set to true of false: ab00810a95cb9ac78855babcebd1f8c6febd4af0

index.js
zokker13: Ah, nice. Didn't know that existed. Added: b39acfdb29d37f064b448e54d50d5dad3610f6e8

zokker13: Good idea, thanks: b39acfdb29d37f064b448e54d50d5dad3610f6e8

index.js
ehmicky: Closing servers sometimes require network round-trips, i.e. can easily take few seconds.

zokker13:
@ehmicky Thanks or the review. I implemented some changes, feel free to check them out again.
The CI seems to hate me - sometimes it works, sometimes not. I'll add docs/tests if you like the changes and deem them worthy.

index.js
ehmicky: I see what you mean.

On one hand, if the signal was coming from the user, we should definitely take into account those extra signals. For example, `SIGINT` is done with `CTRL-C`, and also `SIGQUIT` is done with `CTRL-\` on Windows `cmd.exe`.

But here it's programmatic. Except for the case where signals would be forwarded from parent to child (which should be uncommon), your points are valid.

Also only `SIGTERM`, `SIGKILL` and `SIGTERM` [work on Windows](https://github.com/ehmicky/cross-platform-node-guide/blob/master/docs/networking_ipc/signals.md#cross-platform-signals).

So I agree with you, we should stick to `SIGTERM` only. @sindresorhus, what do you think?

index.js
ehmicky: Yes I agree with you. Myself I usually prefer named parameters and options objects, because it's easier to extend as you mention. However, for parameters that are not expected to be extended too much (which is arguably hard to predict), simple types might be more developer-friendly I believe.

Either way works for me:
  1) the current solution: `options` object with `retry` boolean and `retryAfter` string/number
  2) `options` boolean/string/number

I would like to know what @sindresorhus thinks about it?

index.js
```diff
@@ -237,6 +237,37 @@ const execa = (command, args, options) => {
...
+				Number(options.retryAfter) :
+				5000;
+
+			const retry = !(options !== undefined && options.retry === false);
+
+			const killResult = originalKill(signal);
+
+			if (retry && killResult) {
+				const safetyTimeout = setTimeout(() => {
```
ehmicky: This is a tiny optimization, but `retryAfter` can now be declared within this block, so it's not evaluated if this block is not evaluated.

index.js
```diff
@@ -237,6 +237,37 @@ const execa = (command, args, options) => {
...
+					try {
+						originalKill('SIGKILL');
+					} catch (_) {
+						// All good - process killed
```
ehmicky: This comment should be removed. If an error is thrown there, it's most likely that the process is not killed actually. `process.kill()` does not throw when process is already killed or does not exist.

I checked the Node.js source and it's actually quite unlikely the code will reach this line. You should try to test cover it to see if that's even possible. However it's better to be safe than sorry, consider it would otherwise throw an uncaught exception, which could call `process.exit()`.

index.js
```diff
@@ -237,6 +237,37 @@ const execa = (command, args, options) => {
...
+					}
+				}, retryAfter);
+				safetyTimeout.unref();
+			}
+
+			return killResult;
+		}
+
+		return originalKill(signal);
```
ehmicky: Some style suggestion: could this line be merged with the `originalKill()` from above? This is possible by moving the `if` blocks around.

ehmicky:
Looks good to me! We need a review from @sindresorhus as well.

I added two minor suggestions. It does not look like CI has failed, so all good there :)

Thanks for this PR again.

index.js
zokker13: Yes, I mixed this one up with process.kill even though we talk child_process.kill here. My bad. I removed the comment: b8e4bce76a7565f1bfac4c717ccd7fc2a5739108

index.js
zokker13: Oh yes, thanks: c86bb8ab10b0e4a483678130eec92df4f525c6c0

zokker13:
@ehmicky Implemented some more changes and added documentation. I have problems with the tests though. Maybe it's too early for me. Could have take a look if you have time?
8d06a6c4f4f8fc25e433b99c874319f532d11a59

readme.md
```diff
@@ -126,6 +134,8 @@ It exposes an additional `.all` stream, with `stdout` and `stderr` interleaved.
 
 The spawned process can be canceled with the `.cancel()` method on the promise, which causes the promise to reject an error with a `.isCanceled = true` property, provided the process gets canceled. The process will not be canceled if it has already exited.
 
+The spawned process can be killed with `.kill([signal], [options])` which allows you to retry the process killing at a later time (default: after 5 seconds)
```
ehmicky: Maybe this should be rephrased to emphasize:
  - the retry is automatically done
  - it does not "try" to kill, it forces it

readme.md
```diff
@@ -329,6 +339,24 @@ Default: `false`
...
+
+You may kill the child process with this function. It is slightly different that the original function.
+This one allows you to kill the process with `SIGTERM` and after a specified timeout (5 seconds) a `SIGKILL` is invoked to ensure killing the process.
+
+#### options.retry
+
+Type: `boolean`<br>
+Default: true
+
+Tells the kill to start a second attemp to kill.
```
ehmicky: `Tells the kill to start` is a little confusing to me. Also "attempt" has a typo. 

Maybe instead: If the first signal does not terminate the child process, tries again using `SIGKILL`.

ehmicky: Tells the kill to start is a little confusing to me. Also "attempt" has a typo.

```suggestion
If the first signal does not terminate the child process, tries again using SIGKILL.
```

readme.md
```diff
@@ -329,6 +339,24 @@ Default: `false`
...
+
+#### options.retryAfter
+
+Type: `string | number`<br>
+Default: 5000
+
+Specifies a timeout in ms. This one defines the delay until you try to kill the pocess again.
```
ehmicky: "pocess" has a typo. "This one" might not be needed. Also "you try" implies the user needs to retry, but the retry is automatic.

```suggestion
How long to wait for the child process to terminate before sending `SIGKILL`.
```

readme.md
```diff
@@ -329,6 +339,24 @@ Default: `false`
 
 If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
+### child_process.kill([signal], [options])
+
+You may kill the child process with this function. It is slightly different that the original function.
+This one allows you to kill the process with `SIGTERM` and after a specified timeout (5 seconds) a `SIGKILL` is invoked to ensure killing the process.
```
ehmicky: I think this could be simplified.

```suggestion
Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
```

readme.md
```diff
@@ -109,6 +109,14 @@ try {
 	}
 	*/
 }
+
+// Killing a spawned process gently, then with more force
```
ehmicky: I think the three comments could be summarized in just one, such as:

```suggestion
// If the following spawned process is not terminated by `SIGTERM`, `SIGKILL` will automatically be sent after 2 seconds.
```

readme.md
```diff
@@ -109,6 +109,14 @@ try {
...
+const subprocess = execa('node'); // Lets assume node doesn't react to SIGTERM
+setTimeout(() => {
+	subprocess.kill('SIGKILL', {
```
ehmicky:
```suggestion
subprocess.kill('SIGTERM', {
```

test.js
```diff
@@ -134,6 +134,50 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
 	t.throws(() => execa.shellSync('node fixtures/error-message.js'), {message: STDERR_STDOUT_REGEXP, code: 1});
 });
 
+test.cb('execa() with .kill after it with SIGKILL should kill cleanly', t => {
+	const proc = execa('node', ['fixtures/sub-process-no-killable']);
+	t.is(typeof proc.pid, 'number');
```
ehmicky: I don't think `proc.pid` needs to be tested.

test.js
```diff
@@ -134,6 +134,50 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
 	t.throws(() => execa.shellSync('node fixtures/error-message.js'), {message: STDERR_STDOUT_REGEXP, code: 1});
 });
 
+test.cb('execa() with .kill after it with SIGKILL should kill cleanly', t => {
```
ehmicky: Tests should use the same style as the other tests:
  - `test.cb()` + `test.end()` should use `test()` + `async`/`await` instead. You can promisify `setTimeout` with `util.promisify(setTimeout)`
  - `proc` should be named `subprocess` as the other tests.

fixtures/sub-process-no-killable
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
```
ehmicky: We don't need to test that `SIGTERM` is received. We explicitly call `kill()` in the tests so we can safely assume child process receive it.

I.e. this fixture is not needed, you could use `fixtures/forever` instead.

test.js
```diff
@@ -134,6 +134,50 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
...
+	const proc = execa('node', ['fixtures/sub-process-no-killable']);
+	t.is(typeof proc.pid, 'number');
+	setTimeout(() => {
+		proc.kill('SIGKILL');
+	}, 100);
+
+	setTimeout(() => {
+		t.false(isRunning(proc.pid));
+		t.end();
+	}, 2000);
```
ehmicky: Instead of using both `setTimeout()` you could maybe just:
  - call `proc.kill('SIGKILL') synchronously
  - wait for `await proc` rejection in a try/catch block (and asserts it is rejected)
  - asserts that `error.signal` is `SIGKILL`

fixtures/sub-process-no-killable
```diff
@@ -0,0 +1,10 @@
...
+'use strict';
+
+process.on('SIGTERM', () => {
+  console.log('got sigterm but we ignore it');
+});
+
+setInterval(() => {
+  // run forever
+}, 5000);
```
ehmicky: You might want to increase that timeout. When run in parallel, tests can have quite some lag. See `fixtures/forever`

test.js
```diff
@@ -134,6 +134,50 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
...
+});
+
+test.cb('execa() with .kill after it with SIGTERM should not kill (no retry)', t => {
```
ehmicky: The ideas behind the two tests look good, but with the same comments as the first test.

readme.md
zokker13: Fixed: 2fb0ae0a7664c4c240d3ae9f764b6b0976549827

readme.md
zokker13: Fixed: 2fb0ae0a7664c4c240d3ae9f764b6b0976549827

readme.md
zokker13: Fixed: 2fb0ae0a7664c4c240d3ae9f764b6b0976549827

readme.md
zokker13: Fixed: 
* 2fb0ae0a7664c4c240d3ae9f764b6b0976549827
* 233e222679fcbce4b1547370bffcd5bffe32c7c4

readme.md
zokker13: Fixed: 
* 2fb0ae0a7664c4c240d3ae9f764b6b0976549827
* 233e222679fcbce4b1547370bffcd5bffe32c7c4

readme.md
zokker13: Oh, this totally slipped. :s

Fixed: 
* 2fb0ae0a7664c4c240d3ae9f764b6b0976549827
* 233e222679fcbce4b1547370bffcd5bffe32c7c4

test.js
zokker13: Done: bb5b280c806e049600d3de4db876a7d5d6a17888

test.js
zokker13: bb5b280c806e049600d3de4db876a7d5d6a17888

fixtures/sub-process-no-killable
zokker13: Done: bb5b280c806e049600d3de4db876a7d5d6a17888

test.js
zokker13: Done it but the second test is not upposed to die but dies anyway. This is a bit weird to me. You have an idea?

test.js
zokker13: I agree 
Done: bb5b280c806e049600d3de4db876a7d5d6a17888

readme.md
```diff
@@ -329,6 +339,23 @@ Default: `false`
...
 If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
+### child_process.kill([signal], [options])
+
+Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
+
+#### options.retry
+
+Type: `boolean`<br>
+Default: true
```
ehmicky:
```suggestion
Default: `true`
```

readme.md
```diff
@@ -329,6 +339,23 @@ Default: `false`
...
+
+If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
+
+#### options.retryAfter
+
+Type: `string | number`<br>
+Default: 5000
```
ehmicky:
```suggestion
5000
```

test.js
```diff
@@ -134,6 +134,52 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
...
+test('execa() with .kill after it with SIGKILL should kill cleanly', async t => {
+	const subprocess = execa('node', ['fixtures/sub-process-no-killable']);
+
+	subprocess.kill('SIGKILL');
+
+	try {
+		await subprocess;
+		t.fail('Didnt expect success');
+	} catch (error) {
+		t.deepEqual(error.signal, 'SIGKILL');
```
ehmicky:
```suggestion
		t.is(error.signal, 'SIGKILL');
```

Does a shallow comparison instead of deep (faster).

test.js
```diff
@@ -134,6 +134,52 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
...
 });
 
+test('execa() with .kill after it with SIGKILL should kill cleanly', async t => {
+	const subprocess = execa('node', ['fixtures/sub-process-no-killable']);
+
+	subprocess.kill('SIGKILL');
+
+	try {
+		await subprocess;
+		t.fail('Didnt expect success');
```
ehmicky: You can also use:

```js
await t.notThrowsAsync(subprocess)
```

test.js
ehmicky: I just tried it by cloning your work, and for me the child process does not die. The test passes. Do you get different results?

ehmicky:
@zokker13 thanks for this new batches of updates! Hopefully we can merge this soon! :tada: :tada:

readme.md
zokker13: Done: 772cfd2e28e70ed3b902e10599bdc38a809939e5

readme.md
zokker13: Done: 772cfd2e28e70ed3b902e10599bdc38a809939e5

test.js
zokker13: Fixed: 6e9424db6017cc6268ae0201aacca520c2074c53

test.js
zokker13: Fixed: 6e9424db6017cc6268ae0201aacca520c2074c53

test.js
zokker13: Are you sure it's running? When I add `t.true(isRunning(subprocess.pid));` I get not passing tests. Since the fixture eats all `SIGTERM`s, the process should not finish.

zokker13:
@ehmicky  Ready for another attempt ^^

test.js
ehmicky: Yes you're right it dies but should not.

I figured out the problem: it's a race condition. The fixture file starts running after  `subprocess.kill()` is executed. I.e. when it receives the `SIGTERM` signal, it has not setup any signal handler yet. I.e. it is killed.

So you need to ensure the `SIGTERM` signal is fired after the fixture file has loaded. Possible solutions:
  - `await promisify(setTimeout)(1e3)` before sending `SIGTERM`. Now that one feels a little hacky.
  - use `ipc` and make the fixture file call `process.send()` after setting up `SIGTERM` handler to communicate "I am ready, please send me your signal".
  - is there another way to know that a child process has finished loading? I am not sure.

You would actually need to do that in all tests, not just that test.

How I debugged this problem: I added a `assert.fail()` at the beginning of the fixture file. Since `await subprocess` was throwing with `SIGTERM`, it meant it did not even reach that `assert.fail()`.

fixtures/sub-process-no-killable
ehmicky: You should run `xo` on this file. I am not sure why `npm test` does not do it automatically. But `xo` seems to complain about the whitespace before `console.log()`, and the lack of an uppercase letter at the start of the comment `run forever`.

index.js
```diff
@@ -237,6 +237,32 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options) => {
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
```
ehmicky: Could you please add a test for when `signal` is `undefined`? I.e. when using `kill()` with no arguments.

This would give you 100% test coverage on this PR: https://coveralls.io/builds/23196199/source?filename=index.js#L281

ehmicky:
Thanks! I think I figured out the problem with your test, see my comment. Also I added a suggestion for another test. And there are some linting issue with the fixture file. Thanks for holding on :)

index.js
zokker13: Of course, added: c0c3527d06777617a30b6e4391442f5a5eef3244

test.js
zokker13: Amazing, thanks for bearing with me and give me so much guidance.
I added an ipc listener, now all tests work as expected: c0c3527d06777617a30b6e4391442f5a5eef3244

fixtures/sub-process-no-killable
zokker13: Mh... running xo directly on this odesn't seem to help (no output).
I tried to fix it though: c0c3527d06777617a30b6e4391442f5a5eef3244

Perhaps xo excludes fixtures by default? (Dropping the fixtures  patch in the package.json didnt help).

zokker13:
Thanks for finding that detail! I'm ready for anothr go :)

test.js
```diff
@@ -134,6 +142,64 @@ test('execa.shellSync() includes stdout and stderr in errors for improved debugg
...
+
+	subprocess.kill('SIGTERM', {
+		retryAfter: 50
+	});
+
+	const {signal} = await t.throwsAsync(subprocess);
+	t.is(signal, 'SIGKILL');
+});
+
+test('execa() with .kill after it with nothing (undefined) should kill after 50 ms with SIGKILL', async t => {
```
ehmicky: This test is identical to the one above, i.e. I think you meant to do `kill()` with no arguments

test.js
```diff
@@ -16,6 +16,14 @@ const NO_NEWLINES_REGEXP = /^[^\n]*$/;
 const STDERR_STDOUT_REGEXP = /stderr[^]*stdout/;
 const TIMEOUT_REGEXP = /timed out after/;
 
+const waitForProcessLaunch = subprocess => {
+	return new Promise(resolve => {
+		subprocess.once('message', _ => {
```
ehmicky: This could be simplified to:

```js
await promisify(subprocess.once.bind(subprocess))('message')
```

test.js
zokker13: Argh. Fixed: 79c3411a8c42db38af115ebf46c791ce78f1361b

test.js
zokker13: Oh yes. I didn't think this would work. Was expecting a cb like behavior for `util.promisify` to work.
I changed it in the `waitForProcessLaunch` function and not where it's used (might looks scary for some):
79c3411a8c42db38af115ebf46c791ce78f1361b

ehmicky: Yes I agree it looks a little scary. Hopefully we get the bind operator in JavaScript soon so this can become:

```js
await promisify(subprocess::once)('message')
```

ehmicky:
The tests are failing on CI for Windows. Maybe it's due to my suggestion of using `promisify()`? Or is it something else?

zokker13:
I think the ipc change is just the bad messenger since the tests never truly reported what's _really_ there. I'll make a quick cross check with normal stdout as trigger tomorrow evening.

ehmicky:
Also please note the tests do work on Linux, so there's some OS-specific difference at play here (I'm not sure what).

zokker13:
I took a quick look. Seems like we came across an issue with `SIGTERM`.
It seems like Windows thinks that a `SIGTERM` is supposed to kill the process even though you have a listener installed on it. Exactly the opposite of what the [docs](https://nodejs.org/dist/latest-v10.x/docs/api/process.html#process_signal_events) say:

> 'SIGTERM' is not supported on Windows, it can be listened on.

You can check this with normal process behavior on Windows; Just create a script that listens on `SIGTERM` and try to kill it from another shell with it. The process is killed. Obviously, this kills the purpose of this PR. :D

I'll check whatever the node binding sends to the child when on Windows and see if we can make it work on Windows that way.

ehmicky:
I am going to look at the Node bindings as well, but as far as I remember, while Windows does not any `SIGTERM`, Node emulates it by terminating it using the Windows API. I think the Node.js documentation is misleading here.

I just tried doing the following on Windows:
  - created an `index.js` with `setTimeout(() => {}, 1e6); console.log(process.pid)`
  - run it with `node index.js`
  - in another `cmd.exe` terminal, run node then `process.kill(PID, 'SIGTERM')`
  - it works: process is killed

zokker13:
Yes, nodejs (or libuv) emulates `SIGTERM` by sending `SIGKILL`.
It's good behavior if you want to kill things quickly. But no so great if you want to give the process appropriate time.

I added some conditionals in the tests so that they are green. I'm not a fan of it though. I wonder if the signals are just translated to the "obvious" Windows counterparts.

ehmicky:
I don't think Windows has natively (without using some projects like Cygwin) a concept of `SIGKILL`. 

On Windows `SIGTERM` is translated into [`TerminateProcess`](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-terminateprocess) (from the Windows API). Windows only supports the [following signals](https://github.com/ehmicky/cross-platform-node-guide/blob/master/docs/networking_ipc/signals.md#cross-platform-signals): `SIGKILL`, `SIGTERM`, `SIGINT` and the special signal `0`. Those three signals are handled the exact same way.

You can see it by checking the following lines of code: Node [calls `uv_process_kill()`](https://github.com/nodejs/node/blob/153c101a12d8dc6e149a57f0d4b94d5a310add16/src/process_wrap.cc#L282) which in turns calls [uv__kill()](https://github.com/libuv/libuv/blob/c5593b51dc98715f7f32a919301b5801ebf1a8ce/src/win/process.c#L1190), which calls `TerminateProcess`.

From the MSDN description, it seems that `TerminateProcess` behaves similarly to `SIGKILL` though, in that it does not allow a graceful exit. Precisely CPU execution is stopped, but Windows wait for I/O buffers to be emptied (which can take few seconds).

Now is this an issue with the test (and the feature as whole)? Even if Windows kills the process right away with `SIGTERM`, shouldn't the tests still pass?

ehmicky:
I've done some tests on Windows 10 and this is what I get:
  1) `process.kill()` with `SIGTERM`, `SIGINT` or `SIGKILL` terminates the process right away.
  2) The `signal` in `child_process.on('exit', (code, signal) => {})` event handler correctly matches the signal name. E.g. if `childProcess.kill('SIGTERM')` was used, `signal` will be `SIGTERM`. 
  3) `process.on()` works with `SIGINT` but only when doing CTRL-C in the terminal, not when doing `process.kill()`
  4) `process.on()` does not work `SIGTERM` nor `SIGKILL`.

Because of this, using `process.on('SIGTERM')` to prevent the child process termination in the tests does not work on Windows. Would there be a different method to prevent child process termination on Windows, so we can make sure the `retry` feature works on that OS?

ehmicky:
I digged through Node.js code a bit, and considering the points above, i cannot think of circumstances where `childProcess.kill()` would fail first, but succeed when trying again.

The main reasons why `childProcess.kill()` would fail seem to be:
  - process already killed. Retry won't help.
  - process access rights / permissions issues. Retry won't help.
  - signal was handled but did not exit the process fast enough. This seems to be the main use case for this feature. Retry helps there. But it can't happen on Windows because `SIGTERM` cannot be handled.

Based on that, it seems like this feature is a noop on Windows. I think this is ok. However we should probably skip the tests on Windows.

What do you think @zokker13 and @sindresorhus?

zokker13:
I agree, with all your points basically.
A shame but maybe we need a few more Windows interations to have proper POSIX behavior ;)

Do you think the current implementation is good enough or do you want a pass-through in the code itself? Currently it would be working as you've alredy said. If we can't kill it with SIGTERM (SIGKILL) we will never kill it.

There are different methods to "poison" a process in Windows though. I've had some situations in the past when a process would not be closable through the task manager. Process Hacker had a [terminator](http://i.imgur.com/nnP0TzZ.png) feature which would be creative in killing the processes. Of course nothing you'd want to do in node or as default fallback behavior.

ehmicky:
I think the current implementation works. We could only put the test definitions inside a `if (process.platform !== 'win32`) {}` so that Windows does not attempt to run those tests.

Personally I don't think having some way to have a "stronger" `SIGKILL` on Windows is within the scope of `execa` (and this PR), but I would be curious to know what @sindresorhus thinks.

sindresorhus:
> Personally I don't think having some way to have a "stronger" SIGKILL on Windows is within the scope of execa (and this PR), but I would be curious to know what @sindresorhus thinks.

Agreed

test.js
```diff
@@ -16,6 +17,10 @@ const NO_NEWLINES_REGEXP = /^[^\n]*$/;
 const STDERR_STDOUT_REGEXP = /stderr[^]*stdout/;
 const TIMEOUT_REGEXP = /timed out after/;
 
+const waitForProcessLaunch = subprocess => {
+	return promisify(subprocess.once.bind(subprocess))('message');
```
ehmicky:
```suggestion
    return pEvent(subprocess, 'message')
```

I have checked it and my suggestion of using `promisify` actually was wrong. Luckily tests are already using `p-event`, which is even less verbose.

ehmicky:
@zokker13 Do you need some extra help with this PR? This seems to me that the only things left are:
  - rebasing to latest `master`.
  - skipping the test definitions on Windows using `if (process.platform !== 'win32') {}`
  - I've added a comment about a suggestion I made which actually does not work.

zokker13:
@ehmicky Hi, thanks for asking, I was intimidated by doing this rebase since it screwed me hard. I was somehow able to do it (hopefully) somewhat well. I'll take care tomorrow or Thursday. I'm on it :)

fixtures/sub-process-no-killable
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
```
tomsotte: Sorry to bother, but I'm working on #170 and I might need to name a fixture exactly like this one; I will need to implement a similar retry as yours, but for killing the tree of the descendants of the spawned process. Also the prefix `sub-process` is used already for the fixtures that spawn themselves a sub-process.

Could you rename this to something without `sub-process` prefix? If I might suggest `unkillable` sounds good as a single name, or `no-killable` as well; your choice, I'll follow on my PR.

zokker13: Hi, thanks for letting me know. It's called `no-killable` now.

zokker13:
@ehmicky I think I'm ready for another go.
I think I screwed upp the rebase in the readme somehow and I'm not quite sure what you mean with the restructure there. It looks kinda the same for me (but I'm also horrible with READMEs).

test.js
```diff
@@ -130,16 +134,82 @@ test('stripFinalNewline: false', async t => {
...
+	await waitForProcessLaunch(subprocess);
+
+	subprocess.kill('SIGKILL');
+
+	const {signal} = await t.throwsAsync(subprocess);
+	t.is(signal, 'SIGKILL');
 });
 
+// Windows doesn't support sending signals. No re-emulates them down to SIGKILL
+if (process.platform !== 'win32') {
```
ehmicky: Instead of repeating this comment and condition several times, could you please regroup them?

```js
// Windows doesn't support sending signals. No re-emulates them down to SIGKILL
if (process.platform !== 'win32') {
  test(...)

  test(...)

  test(...)
}
```

test.js
```diff
@@ -130,16 +134,82 @@ test('stripFinalNewline: false', async t => {
...
+		t.is(signal, 'SIGKILL');
+	});
+}
+
 test('stripFinalNewline in sync mode', t => {
 	const {stdout} = execa.sync('noop', ['foo'], {stripFinalNewline: true});
 	t.is(stdout, 'foo');
 });
 
+test('stripFinalNewline on failure', async t => {
```
ehmicky: After the rebase, the tests are now in the middle of the `stripFinalNewline` tests. Could you put it just before them?

test.js
```diff
@@ -130,16 +134,82 @@ test('stripFinalNewline: false', async t => {
...
 
-test('stripFinalNewline on failure', async t => {
-	const {stderr} = await t.throwsAsync(execa('noop-throw', ['foo'], {stripFinalNewline: true}));
-	t.is(stderr, 'foo');
+test('execa() with .kill after it with SIGKILL should kill cleanly', async t => {
+	const subprocess = execa('node', ['fixtures/no-killable'], {
+		stdio: ['ipc']
+	});
+
+	await waitForProcessLaunch(subprocess);
```
ehmicky: `waitForProcessLaunch()` is now simple enough that it could be inlined in the test function, i.e. directly:

```js
await pEvent(subprocess, 'message')
```

readme.md
```diff
@@ -401,6 +411,23 @@ Default: `false`
 
 If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
+### child_process.kill([signal], [options])
```
ehmicky:
```suggestion
### kill([signal], [options])
```

This should be moved just before `cancel()` in the `README`. That section is about the methods of the returned `child_process`, i.e. only the method name needs to be mentioned.

readme.md
```diff
@@ -137,7 +145,9 @@ Unless the [`shell`](#shell) option is used, no shell interpreter (Bash, `cmd.ex
 Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) which:
   - is also a `Promise` resolving or rejecting with a [`childProcessResult`](#childProcessResult).
   - exposes the following additional methods and properties.
+The spawned process can be canceled with the `.cancel()` method on the promise, which causes the promise to reject an error with a `.isCanceled = true` property, provided the process gets canceled. The process will not be canceled if it has already exited.
```
ehmicky: This was removed in previous commits. It probably appeared because of the rebase, but you can safely remove it.

readme.md
```diff
@@ -137,7 +145,9 @@ Unless the [`shell`](#shell) option is used, no shell interpreter (Bash, `cmd.ex
...
 
+The spawned process can be killed with `.kill([signal], [options])` which allows you to retry the process killing at a later time (default: after 5 seconds)
```
ehmicky: This can be removed as well, because the `kill()` method description should follow right away.

index.d.ts
```diff
@@ -288,13 +288,25 @@ declare namespace execa {
...
 	}
 
 	type ExecaChildProcess<StdoutErrorType = string> = ChildProcess &
 		ExecaChildPromise<StdoutErrorType> &
 		Promise<ExecaReturnValue<StdoutErrorType>>;
 }
 
+interface killOptions {
+	retry: boolean;
+	retryAfter: number;
```
ehmicky: Could you please add comments those the options as well? They should be identical to the `README`.

index.d.ts
```diff
@@ -288,13 +288,25 @@ declare namespace execa {
 		Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal). This is preferred when cancelling the child process execution as the error is more descriptive and [`childProcessResult.isCanceled`](#iscanceled) is set to `true`.
 		*/
 		cancel(): void;
+
+		/**
+		 * Kills the subprocess.
+		 *
+		 * Causes the promise to reject an error with a `.isKilled = true` property.
+		 */
+		kill(signal?: string | undefined, options?: killOptions | undefined): void;
```
ehmicky: Could you please move it before `cancel()`? The order of this file is now the same as the order in the `README`.

index.d.ts
```diff
@@ -288,13 +288,25 @@ declare namespace execa {
 		Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal). This is preferred when cancelling the child process execution as the error is more descriptive and [`childProcessResult.isCanceled`](#iscanceled) is set to `true`.
 		*/
 		cancel(): void;
+
+		/**
+		 * Kills the subprocess.
+		 *
+		 * Causes the promise to reject an error with a `.isKilled = true` property.
+		 */
```
ehmicky: This comment should be identical to the `README` description (including Markdown links).

test.js
zokker13: Yes, of course. Weird that I didn't do it like that right away: 00ad35962d5b2960fd3364737ee647eb1035211d

test.js
zokker13: Done: d793df20c3e1ca071510849b295f16257306ee1e

test.js
zokker13: Done: d793df20c3e1ca071510849b295f16257306ee1e

readme.md
zokker13: Done: e2d589df448152eefd4eac35c845dc2d4460af69

readme.md
zokker13: Done: e2d589df448152eefd4eac35c845dc2d4460af69

readme.md
zokker13: Done: e2d589df448152eefd4eac35c845dc2d4460af69

index.d.ts
zokker13: Done: 50e52ad1a3bb510754557ad96bc0150beacce2ec

index.d.ts
zokker13: Done: 50e52ad1a3bb510754557ad96bc0150beacce2ec

index.d.ts
zokker13: Done: 50e52ad1a3bb510754557ad96bc0150beacce2ec

zokker13:
@ehmicky I'd be ready for another attempt :)

readme.md
```diff
@@ -116,6 +116,14 @@ try {
 	}
 	*/
 }
+
+// Kill a  process with SIGTERM and after 2 seconds, kill it  with SIGKILL
```
ehmicky:
```suggestion
// Kill a process with SIGTERM and after 2 seconds, kill it with SIGKILL
```

ehmicky:
PR looks good to me! Just waiting for @sindresorhus review, and we can merge it :)

Thanks for this awesome feature! :tada: :tada: :tada:

readme.md
zokker13: Done: a210520fe3c1eb122dc700dcca67439cf2ce4c90

readme.md
```diff
@@ -401,7 +427,6 @@ Default: `false`
 
 If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
-
```
sindresorhus: Don't do unrelated changes.

test.js
```diff
@@ -130,16 +190,16 @@ test('stripFinalNewline: false', async t => {
...
-
 test('stripFinalNewline in sync mode', t => {
 	const {stdout} = execa.sync('noop', ['foo'], {stripFinalNewline: true});
 	t.is(stdout, 'foo');
 });
 
+test('stripFinalNewline on failure', async t => {
+	const {stderr} = await t.throwsAsync(execa('noop-throw', ['foo'], {stripFinalNewline: true}));
+	t.is(stderr, 'foo');
+});
```
sindresorhus: This test should be moved back in the same position as it were.

fixtures/no-killable
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
+'use strict';
+
+process.on('SIGTERM', () => {
+  console.log('got sigterm but we ignore it');
```
sindresorhus:
```suggestion
  console.log('Received SIGTERM, but we ignore it.');
```

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
 		Promise<ExecaReturnValue<StdoutErrorType>>;
 }
 
+interface killOptions {
```
sindresorhus: The interface should be defined before it's used and should be `KillOptions`, and should be inside the `execa` namespace.

index.d.ts
```diff
@@ -284,6 +284,11 @@ declare namespace execa {
 			onRejected?: (reason: ExecaError<StdoutErrorType>) => ResultType | PromiseLike<ResultType>
 		): Promise<ExecaReturnValue<StdoutErrorType> | ResultType>;
 
+		/**
+		Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
+		*/
+		kill(signal?: string | undefined, options?: killOptions | undefined): void;
```
```suggestion
		kill(signal?: string, options?: execa.KillOptions): void;
```

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
...
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(typeof signal === 'number' && signal === os.constants.signals.SIGTERM)) {
+			const retry = !(options !== undefined && options.retry === false);
+
+			if (retry && killResult) {
+				const retryAfter = options !== undefined &&
+				['number', 'string'].includes(typeof options.retryAfter) ?
```
sindresorhus: Why does it accept `string`? That makes no sense to me.

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options) => {
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(typeof signal === 'number' && signal === os.constants.signals.SIGTERM)) {
```
sindresorhus: This could be made more readable by extracting the condition into a `isSigterm` variable.

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options) => {
```
```suggestion
	spawned.kill = (signal, options = {}) => {
```

Then you can drop https://github.com/sindresorhus/execa/pull/208/files#diff-168726dbe96b3ce427e7fedce31bb0bcR263

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
...
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
```
sindresorhus: I don't think we should silently accept lowercase `sigterm`. If anything, we should throw an error that it's incorrect.

index.js
sindresorhus: Can `signal` actually be a number for `cp.kill()`? Not according to the Node.js docs, at least: https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
...
+
+			if (retry && killResult) {
+				const retryAfter = options !== undefined &&
+				['number', 'string'].includes(typeof options.retryAfter) ?
+					Number(options.retryAfter) :
+					5000;
+				const safetyTimeout = setTimeout(() => {
+					try {
+						originalKill('SIGKILL');
+					} catch (_) {
```
```suggestion
					} catch (_) {}
```

index.js
```diff
@@ -251,6 +251,32 @@ const execa = (command, args, options) => {
...
+					}
+				}, retryAfter);
```
```suggestion
				}, retryAfter).unref();
```

readme.md
```diff
@@ -116,6 +116,14 @@ try {
 	}
 	*/
 }
+
+// Kill a process with SIGTERM and after 2 seconds, kill it with SIGKILL
```
```suggestion
// Kill a process with SIGTERM, and after 2 seconds, kill it with SIGKILL
```

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
...
+Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
+
+##### options.retry
+
+Type: `boolean`<br>
+Default: `true`
+
+If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
+
+##### options.retryAfter
```
sindresorhus: I think `retryAfter` => `forceKillAfter` would be clearer.

test.js
```diff
@@ -120,6 +120,66 @@ test('skip throwing when using reject option in sync mode', t => {
...
+
+	await pEvent(subprocess, 'message');
+
+	subprocess.kill('SIGKILL');
+
+	const {signal} = await t.throwsAsync(subprocess);
+	t.is(signal, 'SIGKILL');
+});
+
+// Windows doesn't support sending signals. No re-emulates them down to SIGKILL
```
sindresorhus: > No re-emulates them down to SIGKILL

I don't understand what this is trying to say. It should also end in a `.`

test.js
```diff
@@ -120,6 +120,66 @@ test('skip throwing when using reject option in sync mode', t => {
 	t.is(exitCode, 2);
 });
 
+test('execa() with .kill after it with SIGKILL should kill cleanly', async t => {
```
```suggestion
test('execa() with .kill() after it with SIGKILL should kill cleanly', async t => {
```

index.d.ts
```diff
@@ -284,6 +284,11 @@ declare namespace execa {
 			onRejected?: (reason: ExecaError<StdoutErrorType>) => ResultType | PromiseLike<ResultType>
 		): Promise<ExecaReturnValue<StdoutErrorType> | ResultType>;
 
+		/**
+		Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
```
```suggestion
		Same as the original [`child_process#kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal), except if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
```

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
 		Promise<ExecaReturnValue<StdoutErrorType>>;
 }
 
+interface killOptions {
+	/**
+	If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
+	*/
+	retry: boolean;
+	/**
```
```suggestion

	/**
```

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
...
+	How long to wait for the child process to terminate before sending `SIGKILL`.
+	*/
+	retryAfter: number;
```
```suggestion
	retryAfter?: number;
```

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
 		Promise<ExecaReturnValue<StdoutErrorType>>;
 }
 
+interface killOptions {
+	/**
+	If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
+	*/
+	retry: boolean;
```
```suggestion
	retry?: boolean;
```

index.d.ts
sindresorhus: You need to add some type tests for this. https://github.com/sindresorhus/execa/blob/master/index.test-d.ts

readme.md
```diff
@@ -116,6 +116,14 @@ try {
 	}
 	*/
 }
+
+// Kill a process with SIGTERM and after 2 seconds, kill it with SIGKILL
+const subprocess = execa('node');
+setTimeout(() => {
+	subprocess.kill('SIGTERM', {
+		retryAfter: 2000,
```
```suggestion
		retryAfter: 2000
```

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
   - is also a `Promise` resolving or rejecting with a [`childProcessResult`](#childProcessResult).
   - exposes the following additional methods and properties.
 
+#### kill([signal], [options])
+
+Same as the original [`child_process.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) except: if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
```
```suggestion
Same as the original [`child_process#kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal), except if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`.
```

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
...
+
+##### options.retry
+
+Type: `boolean`<br>
+Default: `true`
+
+If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
```
```suggestion
If the first signal does not terminate the child after a specified timeout, a `SIGKILL` signal will be sent to the process.
```

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
...
+
+##### options.retryAfter
+
+Type: `string | number`<br>
```
```suggestion
Type: `number`<br>
```

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
...
+Default: `5000`
```
sindresorhus: What considerations were given for having 5000 as the default?

readme.md
```diff
@@ -138,6 +146,24 @@ Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#c
...
+
+How long to wait for the child process to terminate before sending `SIGKILL`.
```
```suggestion
Milliseconds to wait for the child process to terminate before sending a `SIGKILL` signal.
```

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
...
+	/**
+	How long to wait for the child process to terminate before sending `SIGKILL`.
```
```suggestion
	Milliseconds to wait for the child process to terminate before sending a `SIGKILL` signal.
```

sindresorhus: You also need to document the default values for these options, per https://github.com/sindresorhus/typescript-definition-style-guide

index.d.ts
```diff
@@ -295,6 +300,17 @@ declare namespace execa {
 		Promise<ExecaReturnValue<StdoutErrorType>>;
 }
 
+interface killOptions {
+	/**
+	If the first signal does not terminate the child after a specified timout, a `SIGKILL` will be sent to the process.
```
```suggestion
	If the first signal does not terminate the child after a specified timeout, a `SIGKILL` signal will be sent to the process.
```

fixtures/no-killable
```diff
@@ -0,0 +1,12 @@
...
+});
+
+process.send('');
+
+setInterval(() => {
+  // Run forever
+}, 20000);
```
sindresorhus: This is not forever. Should be much higher in case of very slow CI's.

index.js
ehmicky: `signal` can be a `number`: 
https://github.com/nodejs/node/blob/908292cf1f551c614a733d858528ffb13fb3a524/lib/internal/child_process.js#L435
https://github.com/nodejs/node/blob/908292cf1f551c614a733d858528ffb13fb3a524/lib/internal/util.js#L220

But it's undocumented (intentionally or not) it seems. I have [filled an issue in Node.js](https://github.com/nodejs/node/issues/27750)

readme.md
ehmicky: We should rename `retry` to `forceKill` as well. It actually is not really a "retry" since we are somewhat guaranteed to succeed.

readme.md
ehmicky: Originally @zokker13 had it as `2000`. I suggested `5000` because I feel like 2 seconds is not enough to do proper shutdown of some processes. E.g. some servers need to perform some network requests before properly closing.

But that's an edge case, so this sounds good if you want a much smaller default value. Which one do you have in mind?

test.js
ehmicky:
```suggestion
// `SIGTERM` cannot be caught on Windows, and it always also abort the process (like `SIGKILL` on Unix).
// Therefore, this feature and those tests do not make sense on Windows. 
```

index.js
ehmicky: That was because that's how Node.js `childProcess.kill()` [handles it](https://github.com/nodejs/node/blob/908292cf1f551c614a733d858528ffb13fb3a524/lib/internal/util.js#L225).

fixtures/no-killable
zokker13: The number value isn't important since we use an `setInterval` here and the program never finishes.

test.js
zokker13: The reason why the change is visible is because I screwed up with the rebase. I think I'll attempt a second rebase on the current master as there are newer changes anyway.

fixtures/no-killable
zokker13: Fixed: 9fa1e1bb3646fc11a23a93fd700882f1ac227b3b

index.d.ts
zokker13: Fixed: 465a4d1d9fc7b640c142a00aa6ed08dc0c8a7b15

index.d.ts
zokker13: Fixed: 465a4d1d9fc7b640c142a00aa6ed08dc0c8a7b15

index.js
zokker13: Fixed: a09b4a35609b4423ce14d5424539cd6e6c39f13f

index.js
zokker13: Fixed: a09b4a35609b4423ce14d5424539cd6e6c39f13f

index.js
zokker13: Fixed: a09b4a35609b4423ce14d5424539cd6e6c39f13f

index.js
zokker13: Fixed: a09b4a35609b4423ce14d5424539cd6e6c39f13f

index.d.ts
zokker13: Fixed: 2756b71cb0dd7021dac6cbcc164c26a027aba22a

index.d.ts
zokker13: Fixed: 2756b71cb0dd7021dac6cbcc164c26a027aba22a

index.d.ts
zokker13: I think this one disappeared through other changes :o

index.d.ts
zokker13: Fixed: 2756b71cb0dd7021dac6cbcc164c26a027aba22a

readme.md
zokker13: Done: 57a54749eda4a6cc4013f31dda7adcc34dcd1f80

test.js
zokker13: Done: 70cc9d08fb63835298bd2a1155e0c3959aa05554

test.js
zokker13: Thanks, fixed: bd37113e03cddea242b2cb084bc56ef0eca88221

index.d.ts
zokker13: Done.

index.d.ts
zokker13: Done.

readme.md
zokker13: Done.

readme.md
zokker13: Done.

readme.md
zokker13: Done: 4d249ddc25e495ce596a35af03fe406694ceefdf

readme.md
zokker13: Done: 4d249ddc25e495ce596a35af03fe406694ceefdf

readme.md
zokker13: Done: 4d249ddc25e495ce596a35af03fe406694ceefdf

readme.md
zokker13: Done: 4d249ddc25e495ce596a35af03fe406694ceefdf

index.d.ts
zokker13: I hope this is sufficient: 353cd0a203a6f2a00ea5482b9f96656da4bb9623

zokker13:
@ehmicky @sindresorhus Would be ready for another round

test.js
ehmicky: You don't need to. Just move the test and commit it, and it should remove that diff.

index.js
```diff
@@ -256,6 +256,29 @@ const execa = (command, args, options) => {
...
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(typeof signal === 'number' && signal === os.constants.signals.SIGTERM)) {
+			const forceKill = !(options !== undefined && options.forceKill === false);
+
+			if (forceKill && killResult) {
+				const forceKillAfter = ['number'].includes(typeof options.forceKillAfter) ?
+					Number(options.forceKillAfter) :
+					5000;
```
ehmicky:
```suggestion
const forceKillAfter = Number.isInteger(options.forceKillAfter) ? options.forceKillAfter : 5000;
```

index.js
```diff
@@ -256,6 +256,29 @@ const execa = (command, args, options) => {
...
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options = {}) => {
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(typeof signal === 'number' && signal === os.constants.signals.SIGTERM)) {
+			const forceKill = !(options !== undefined && options.forceKill === false);
+
+			if (forceKill && killResult) {
```
ehmicky:
```suggestion
if (options.forceKill !== false && killResult) {
```

Then you can remove the `const forceKill = ...` line above.

index.js
```diff
@@ -256,6 +256,29 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal, options = {}) => {
```
ehmicky:
```suggestion
	spawned.kill = (signal = 'SIGTERM', options = {}) => {
```

Then you can remove `if signal === undefined` below

index.js
```diff
@@ -256,6 +256,29 @@ const execa = (command, args, options) => {
...
+		const killResult = originalKill(signal);
+		if (signal === undefined ||
+			(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(typeof signal === 'number' && signal === os.constants.signals.SIGTERM)) {
```
ehmicky:
```suggestion
		signal === os.constants.signals.SIGTERM) {
```

The `typeof signal === 'number'` is not needed since you're comparing with a `number` afterwards.

index.d.ts
ehmicky: I think you forgot to add the default value in the TypeScript definition.

test.js
```diff
@@ -51,40 +51,6 @@ test('stdout/stderr/all available on errors', async t => {
 	t.is(typeof all, 'string');
 });
 
-test('stdout/stderr/all are undefined if ignored', async t => {
```
ehmicky: You accidentally removed all those tests. Just copy/paste them again and commit them, it should fix the diff.

index.test-d.ts
```diff
@@ -8,6 +8,38 @@ import {
 	ExecaSyncError
 } from '.';
 
+try {
```
ehmicky: All those properties are already tested, i.e. don't need to be tested again.

The only properties that need testing are the options of `kill()` (and those are not tested at the moment).

index.js
zokker13: Good point, fixed: 2987b9fbe8cf34d560942360a2c5e095f0b83108

index.js
zokker13: Fixed: 2987b9fbe8cf34d560942360a2c5e095f0b83108

index.js
zokker13: No, I still need it. Otherwise the behavior would be explicit. Now people need to specify `false` if they want to disable it. Your approach would require people to write `true` in the options.
Or we add default properties in the object but then we'd be screwed if the user decides to only specify the `forceKillAfter`.

index.js
zokker13: Fixed: 2987b9fbe8cf34d560942360a2c5e095f0b83108

test.js
zokker13: Oops, now they are back: c092016523a8c4d9f7ac19e95fc89fd0d633dc11

index.test-d.ts
zokker13: Like this?: b85310f5626281c7c9bbb2352281e104a824778b

I have to admit that I don't really understand this kind of testing (if that's what you want to see).
Is this only for semver checking basically?

index.js
ehmicky: Yes that makes sense (basically the default value of `forceKill` is `true`) if I wrote `!options.forceKill` but it was `options.forceKill !== false`. When no `forceKill` option was passed, this will evaluate to false.

```js
const forceKill = !(options.forceKill === false);
if (forceKill && killResult) {
```

is the same as:

```js
if (!(options.forceKill === false) && killResult) {
```

which in turn is the same as:

```js
if (options.forceKill !== false && killResult) {
```

index.js
```diff
@@ -256,6 +256,28 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal = 'SIGTERM', options = {}) => {
+		const killResult = originalKill(signal);
+		if ((typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			(signal === os.constants.signals.SIGTERM)) {
```
ehmicky:
```suggestion
			signal === os.constants.signals.SIGTERM) {
```

index.test-d.ts
ehmicky: You can find out more about this kind of TypeScript tests here: https://github.com/SamVerschueren/tsd

index.js
zokker13: Fixed: ec8f1dff80634da6c4e9546535999947f6b1c039

zokker13:
@ehmicky Ready for another go..

index.js
```diff
@@ -257,6 +257,26 @@ const execa = (command, args, options) => {
 		return Promise.reject(error);
 	}
 
+	const originalKill = spawned.kill.bind(spawned);
+	spawned.kill = (signal = 'SIGTERM', options = {}) => {
+		const killResult = originalKill(signal);
+		if ((typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM') ||
+			signal === os.constants.signals.SIGTERM) {
+			if (options.forceKill !== false && killResult) {
```
ehmicky: Should we go crazy and combine those two `if` statements into a single one? :)

Also as @sindresorhus suggested, it might be a good idea to extract that into a separate function that returns a boolean.

zokker13: I think we should. :)
Done: 451a743ef2db8784edd87580ce0741ae91bf8e56

index.d.ts
```diff
@@ -279,11 +279,27 @@ declare namespace execa {
 		isCanceled: boolean;
 	}
 
+	interface killOptions {
+		/**
+		If the first signal does not terminate the child after a specified timeout, a `SIGKILL` signal will be sent to the process.
+		*/
+		forceKill?: boolean;
```
ehmicky: You should add `@default` for the two options.

ehmicky:
The only things left for me are those three small comments: [here](https://github.com/sindresorhus/execa/pull/208/files#r286855054), [here](https://github.com/sindresorhus/execa/pull/208/files#r285229594) and [there](https://github.com/sindresorhus/execa/pull/208/files#r285229969).

Otherwise looks good :+1: 
Thanks again @zokker13!

index.d.ts
zokker13: Done: 99a6aaf7fc8efff0c4c85164ed89b54d4a5f3388

zokker13:
@ehmicky now it should be in a suitable state :)

ehmicky:
This looks good to me, thanks! Ready for another code review from @sindresorhus.

ehmicky:
There's a small conflict with the latest `master`.

zokker13:
@ehmicky wow, I was finally able to rebase master in here so you may check it out again.
`git rebase --preserve-merges` is truly a blessing. <3

ehmicky:
Rebase looks good :)

zokker13:
@ehmicky what else is needed to have it merged?

sindresorhus:
Looks good now :)

ehmicky:
Thanks a lot @zokker13! 

I hope the code review process was not too tedious for you!

zokker13:
It was okay. I didn't expect so much back and forth but that means people care for the package/feature so it's a good thing :)

ehmicky:
Yes `execa` is used directly or indirectly by [2 million GitHub repositories](https://github.com/sindresorhus/execa/network/dependents) so changes need extra care. I hope this was not too tedious, especially having to rebase.

zokker13:
Ya, the usage is pretty overwhelming. 
But rebasing is fun and good practice I think. Once you figure it out that is ;)

PR #212, opened 2019-05-06
Fix non-executable execution with `input` option

### LINKED ISSUE #166: Crash with input option and absolute path without execute permissions
To reproduce, create `/non-executable-file` and don't give it execute permissions.

```js
execa('/non-executable-file', { input: 'foo' });
```

It crashes here because `stdin` doesn't exist:
https://github.com/sindresorhus/execa/blob/bfc9a4e228b7f865c4cff113fb5c015ebc0e01cc/index.js#L92

stroncium:
`execa` throws internal error instead of rejecting with proper error when running non-executable with input.

fixes #166

Failing test: https://travis-ci.org/sindresorhus/execa/builds/528756511?utm_source=github_status&utm_medium=notification

Added test, fixed the problem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,7 +107,7 @@ function handleArgs(command, args, options = {}) {
 }
 
 function handleInput(spawned, input) {
-	if (input === undefined) {
+	if (input === undefined || spawned.stdin === undefined) {
 		return;
 	}
 
--- a/index.js
+++ b/index.js
@@ -107,6 +107,7 @@ function handleArgs(command, args, options = {}) {
 }
 
 function handleInput(spawned, input) {
+	// Checking for stdin is workaround for https://github.com/nodejs/node/issues/26852 on nodejs v10, v11, v12
 	if (input === undefined || spawned.stdin === undefined) {
 		return;
 	}
--- a/index.js
+++ b/index.js
@@ -107,7 +107,7 @@ function handleArgs(command, args, options = {}) {
 }
 
 function handleInput(spawned, input) {
-	// Checking for stdin is workaround for https://github.com/nodejs/node/issues/26852 on nodejs v10, v11, v12
+	// Checking for stdin is workaround for https://github.com/nodejs/node/issues/26852 on Node.js 10 and 12
 	if (input === undefined || spawned.stdin === undefined) {
 		return;
 	}
--- a/test.js
+++ b/test.js
@@ -336,6 +336,12 @@ if (process.platform !== 'win32') {
 		const cp = execa('non-executable');
 		await t.throwsAsync(cp);
 	});
+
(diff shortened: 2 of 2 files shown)


test.js
```diff
@@ -336,6 +336,12 @@ if (process.platform !== 'win32') {
 		const cp = execa('non-executable');
 		await t.throwsAsync(cp);
 	});
+
+	test('execa() rejects with correct error and doesn\'t throw if running non-executable with input', async t => {
+		await t.throwsAsync(() => {
```
ehmicky: You should be able to shorten it to:

```js
await t.throwsAsync(execa('non-executable', {input: 'Hey!'}), /EACCESS/);
```

stroncium: But it becomes unreadable then...

ehmicky: I don't think this is unreadable. But either way, this is more consistent with the other tests:
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L60
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L67
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L71
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L75
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L79
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L83
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L88
  - https://github.com/sindresorhus/execa/blob/075faf3d6885f8f91eb5945d9c8b2f78ad4f0b59/test.js#L198
  - and so on

ehmicky:
Thanks for this PR @stroncium! 

This issue was reported with #166. There is an ongoing fix within Node.js here: https://github.com/nodejs/node/issues/26852

However for current and older Node.js versions, we should definitely use that check, i.e. looks good to me.

Just need @sindresorhus review as well.

index.js
```diff
@@ -107,7 +107,7 @@ function handleArgs(command, args, options = {}) {
 }
 
 function handleInput(spawned, input) {
-	if (input === undefined) {
+	if (input === undefined || spawned.stdin === undefined) {
```
sindresorhus: Can you add a comment on it being a workaround for https://github.com/nodejs/node/issues/26852 ?

stroncium: done

PR #216, opened 2019-05-07
Improve test for `options.preferLocal`

ehmicky:
This improves the test for `options.preferLocal`:
  - makes it work on Windows
  - simplifies it
  - removes `cat-names` dependency
  - removes `test.serial()`, i.e. make tests faster


ehmicky:
CI tests are now passing.

PR #219, opened 2019-05-07
Remove `execa.shell()` and `execa.shellSync()`

### LINKED ISSUE #211: Deprecate `execa.shell()` in favor of `execa(..., { shell: true })`
Should we deprecate and/or remove `execa.shell()` in favor of `execa(..., { shell: true })`?

The reason being `shell` is a misused option, which creates problems, so we should not encourage its use by providing a shortcut method.

Some of the issues with the `shell` option:
  1) It is not cross-platform. While it does call `cmd.exe` on Windows and `/bin/sh` on Unix, those shells have different syntaxes. If the command uses shell-specific syntax, it won't be cross-platform. It it does not, `shell` is not needed. 
  2) It is slower, as it interprets the command through a shell interpreter.
  3) It is unsafe as it allows for command injection.
  4) It is often used only to be able to pass command and arguments as a string string. Which is not an issue anymore since #182 

There are some legitimate use cases for the `shell` option, but I believe 99% of the times its usage does not satisfy the following criteria:
  a) `shell` is only useful for shell-specific features not available in JavaScript and Node.js. There are really not many of those.
  b) `shell` is only useful when cross-platform is not a concern.

Note: I can submit a PR if approved.

ehmicky:
Fixes #211 

Since the next version is a major release, I thought this would be the perfect time to directly remove those methods instead of deprecating them.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -389,51 +389,6 @@ declare const execa: {
 	stderr(file: string, options?: execa.Options<null>): Promise<Buffer>;
 
 	/**
-	Execute a command through the system shell.
-
-	Prefer `execa()` whenever possible, as it's both faster and safer.
-
-	@param command - The command to execute.
-	@returns A [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess).
-
-	@example
-	```
-	import execa from 'execa';
-
-	(async => {
-		// Run a shell command
-		const {stdout} = await execa.shell('echo unicorns');
-		//=> 'unicorns'
-
-		// Catching an error
-		try {
-			await execa.shell('exit 3');
-		} catch (error) {
-			console.log(error);
-			//{
-			//	message: 'Command failed with exit code 3 (ESRCH): exit 3',
-			//	code: 3,
-			//	exitCode: 3,
-			//	exitCodeName: 'ESRCH',
-			//	stdout: '',
-			//	stderr: '',
-			//	all: '',
-			//	failed: true,
-			//	command: 'exit 3',
-			//	timedOut: false,
-			//	killed: false
-			//}
-		}
-	})();
-	```
-	*/
-	shell(command: string, options?: execa.Options): execa.ExecaChildProcess;
-	shell(
-		command: string,
-		options?: execa.Options<null>
-	): execa.ExecaChildProcess<Buffer>;
-
-	/**
 	Execute a file synchronously.
 
 	This method throws an `Error` if the command fails.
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -36,7 +36,7 @@ $ npm install execa
 const execa = require('execa');
 
 (async () => {
-	const {stdout} = await execa('echo', ['unicorns']);
+	const {stdout} = await execa('echo unicorns');
```
sindresorhus: This is unrelated to the actual change. And even though we support this syntax, I still prefer to show the array version first.

readme.md
```diff
@@ -236,6 +215,11 @@ Default: `false`
 
 If `true`, runs `command` inside of a shell. Uses `/bin/sh` on UNIX and `cmd.exe` on Windows. A different shell can be specified as a string. The shell should understand the `-c` switch on UNIX or `/d /s /c` on Windows.
 
+We recommend against using this option since it is:
+- not cross-platform, encouraging shell-specific syntax.
+- slower, because of the additional shell interpretation.
+- unsafe, allowing command injection.
```
```suggestion
- unsafe, potentially allowing command injection.
```

readme.md
ehmicky: Fixed!

ehmicky:
Fixed after code review.

readme.md
```diff
@@ -236,6 +215,11 @@ Default: `false`
 
 If `true`, runs `command` inside of a shell. Uses `/bin/sh` on UNIX and `cmd.exe` on Windows. A different shell can be specified as a string. The shell should understand the `-c` switch on UNIX or `/d /s /c` on Windows.
 
+We recommend against using this option since it is:
+- not cross-platform, encouraging shell-specific syntax.
+- slower, because of the additional shell interpretation.
+- unsafe, potentially allowing command injection.
```
sindresorhus: This should be added here too: https://github.com/sindresorhus/execa/blob/df50cda9f4b0bd07fc6a9e709e223427cd9c977b/index.d.ts#L68

ehmicky:
Fixed.

index.d.ts
```diff
@@ -65,6 +65,7 @@ declare namespace execa {
 
 		/**
 		If `true`, runs `command` inside of a shell. Uses `/bin/sh` on UNIX and `cmd.exe` on Windows. A different shell can be specified as a string. The shell should understand the `-c` switch on UNIX or `/d /s /c` on Windows.
+		We recommend against using this option since it is not cross-platform, slower and unsafe.
```
sindresorhus: The TS text should be exactly what's in the readme. Easier to keep them in sync that way.

sindresorhus:

ehmicky:
Fixed

kopax:
Too bad you are not showing any concret example of before/after of this breaking change.

ehmicky:
Hi @kopax,

You might get some extra information in this [Medium article](https://codeburst.io/execa-v2-20ffafeedfdf#1025).

Generally speaking: 
Before: `execa.shell(command)` and `execa.shellSync(command)` 
After: `execa(command, {shell: true})` and `execa.sync(command, {shell: true})`

kopax:
Thanks, I finally understood by myself, but that's always annoying to have to read the whole documentation to find how the breaking change is affecting our code. I recommend to always have a concrete migration example for all of them

ehmicky:
The [release notes](https://github.com/sindresorhus/execa/releases/tag/v2.0.0) show:

```md
### Breaking changes
- Remove `execa.shell()` and `execa.shellSync()`. The [`shell option`](https://github.com/sindresorhus/execa#shell) should be used instead.
```

The `shell option` includes a link to its documentation, I think it should be a quite clear migration path. I agree that an explicit before/after example would have been better, but the release notes weren't too cryptic either about that.

kopax:
I got through this but my advice is still valid, that's just general advice so people in general don't have to go through as I did.

PR #221, opened 2019-05-07
Remove compatibility layer for `util.getSystemErrorName()`

ehmicky:
Although the [Node.js documentation says that `util.getSystemErrorName()`](https://nodejs.org/api/util.html#util_util_getsystemerrorname_err) is added in [`9.7.0`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V9.md#2018-03-01-version-970-current-rvagg-prepared-by-addaleax), it's actually been back-ported 6 months later to [`8.12.0`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#2018-09-11-version-8120-carbon-lts-mylesborins). I tried it with `nvm` and confirm this function works in `8.12.0`.

We use a compatibility layer for this function, which can now be removed.

I am a little unclear whether we support `8.0.0` to `8.11.0`, or only the later `8.*.*` versions:
  - the `package.json` `engines.node` field shows `>=8`, which is translated by `npm` to `>=8.0.0` (using `semver` under the hood).
  - but the version tested on Travis is `8`, which is translated to the latest `8.16.0` (using `nvm` under the hood).

(I think those two should be in sync, but that's a separate issue).

As a side effect, this would also fix [test coverage](https://coveralls.io/builds/23233590/source?filename=lib/errname.js) there.


sindresorhus:
> I am a little unclear whether we support 8.0.0 to 8.11.0, or only the later 8.*.* versions:

Generally, it's expected that users upgrade patch versions as those contain important security fixes. We can continue this in https://github.com/sindresorhus/meta/issues/8.

sindresorhus:
Yay. Feels good getting that removed.

PR #222, opened 2019-05-07
Remove some unused code (`Math.abs()` on exit codes)

ehmicky:
`Math.abs()` is used on the `code` returned by `child_process.on('exit', (code, signal) => {})`.

However [`code` is always positive](https://github.com/nodejs/node/blob/master/lib/internal/child_process.js#L247).

### DIFF
--- a/index.js
+++ b/index.js
@@ -220,7 +220,7 @@ function getCode({error = {}}, code) {
 	}
 
 	if (Number.isInteger(code)) {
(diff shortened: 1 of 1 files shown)


ehmicky:
By the way, test coverage is now 100% with the latest PRs :)

sindresorhus:
🎉🙌🎉

PR #224, opened 2019-05-07
Fix test for `options.cleanup`

ehmicky:
The unit test for `options.cleanup` has several issues and wasn't checking this feature properly.

It first noticed that some lines of code of this feature were not [test covered](https://coveralls.io/builds/23233590/source?filename=index.js#L292). Then I saw some inconsistencies between the names of the tests and which signals were triggered:

```js
test('cleanup - SIGINT', spawnAndKill, 'SIGINT', true);
test('cleanup - SIGKILL', spawnAndKill, 'SIGTERM', true);
test('cleanup false - SIGINT', spawnAndKill, 'SIGTERM', false);
test('cleanup false - SIGKILL', spawnAndKill, 'SIGKILL', false);
```

Digging deeper, the unit test was not checking the feature properly. This fixes it and also removes any timeout, making it more predictable.


sindresorhus:
Good find 👌

PR #225, opened 2019-05-07
Improve `cleanup` option (code quality, tests, documentation)

ehmicky:
This improves the `cleanup` option, but does not change its behavior:
  - small code simplification
  - added tests for `options.detached: true` (before only `options.detached: false` was tested)
  - fix tests not properly cleaning up created processes (i.e. memory leak)
  - improve option description on `README`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -102,6 +102,7 @@ declare namespace execa {
 
 		/**
 		Keep track of the spawned process and `kill` it when the parent process exits.
+		Kill the spawned process when the parent process exits unless either the spawned process is detached, or the parent process is terminated abruptly.
 
 		@default true
 		*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -101,8 +101,9 @@ declare namespace execa {
 		readonly reject?: boolean;
 
 		/**
-		Keep track of the spawned process and `kill` it when the parent process exits.
-		Kill the spawned process when the parent process exits unless either the spawned process is detached, or the parent process is terminated abruptly.
+		Kill the spawned process when the parent process exits unless either:
+			- the spawned process is [`detached`](https://nodejs.org/api/child_process.html#child_process_options_detached)
+			- the parent process is terminated abruptly, for example, with `SIGKILL` as opposed to `SIGTERM` or a normal exit
 
 		@default true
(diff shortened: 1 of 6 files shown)


readme.md
```diff
@@ -277,7 +277,9 @@ Setting this to `false` resolves the promise with the error instead of rejecting
 Type: `boolean`<br>
 Default: `true`
 
-Keep track of the spawned process and `kill` it when the parent process exits.
+Kill the spawned process when the parent process exits unless either:
+  - the spawned process is [`detached`](https://nodejs.org/api/child_process.html#child_process_options_detached)
+  - the parent process is terminated abruptly, e.g. with `SIGKILL` (as opposed to `SIGTERM` or a normal exit)
```
sindresorhus: Don't forget to update the TypeScript definition docs too when doing readme changes.

```suggestion
  - the parent process is terminated abruptly, for example, with `SIGKILL` as opposed to `SIGTERM` or a normal exit
```

ehmicky:
Fixed.

I've also added some tests when the child process exists before the parent process, with the `cleanup` and `detached` options. This was not tested before.

index.d.ts
```diff
@@ -102,6 +102,7 @@ declare namespace execa {
 
 		/**
 		Keep track of the spawned process and `kill` it when the parent process exits.
+		Kill the spawned process when the parent process exits unless either the spawned process is detached, or the parent process is terminated abruptly.
```
sindresorhus: The TS text should be exactly what's in the readme.

index.d.ts
```diff
@@ -102,6 +102,7 @@ declare namespace execa {
 
 		/**
 		Keep track of the spawned process and `kill` it when the parent process exits.
```
sindresorhus: This was removed from the readme.

index.d.ts
ehmicky: Sure, I will copy it.
I was thinking TypeScript comments appear in VS Code "suggestions" as you type, i.e. are read in a different context than a `README` and might need to be more concise. 
But copying the `README` makes it simpler to maintain.

ehmicky:
Fixed

PR #226, opened 2019-05-09
Simplify `cancel()` method

ehmicky:
This is a follow up on #189 by @ammarbinfaisal. It simplifies the `cancel()` method by removing some code, without changing behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -358,7 +358,7 @@ const execa = (command, args, options) => {
 			result.stderr = results[2];
 			result.all = results[3];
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Are you sure these checks are not needed? Could be that we're just not having tests for their behavior.

ehmicky:
`isCanceled` is always `false` because if `cancel()` was called, then `result.signal` will be defined.

`spawned.kill()` is a fast noop (does not throw and return `false`) if the child process has been killed. I.e. checking `if (spawned.killed)` before is not needed.

PR #227, opened 2019-05-10
Improve `error.killed`

ehmicky:
This changes the conditions behind `error.killed`.

### a) `childProcess.kill()` was successfully called

`true` in core Node.js, `true` in `execa`. Unchanged.

### b) `process.kill(childProcess.pid)` was called

`false` in core Node.js, `false` in `execa`. Unchanged.

Ideally this should be `true`, but I am not sure that's possible. The issue is that Windows uses `exitCode: 1` and no `signal`, while Unix uses `exitCode: null` and `signal: 'SIGTERM'`. The difference is due to Node.js emulating termination signals on Windows. On Windows (with just `exitCode: 1`) it seems impossible to know whether the child process was killed externally.

One consequence of this is: while `process.kill(childProcess.pid)` will set `error.killed` to `false` on both Windows and Unix, the error message will be different: `was killed with SIGTERM` on Unix, `failed with exit code 1` on Windows.

### c) `timeout` was reached

`true` in core Node.js, `true` in `execa`. Changed to `false` with this PR. We have `error.timedOut` for this. This makes those boolean error properties orthogonal to each other.

### d) `maxBuffer` was reached

`true` in core Node.js, `false` in `execa`. Unchanged. We should add a property similar to `error.timedOut` but for `maxBuffer` instead. I can do it in a separate PR.

Note: this not does change `childProcess.killed`, which is a separate property and is `true` when `childProcess()` was successfully called (in both core Node.js and `execa`).

### DIFF
--- a/index.js
+++ b/index.js
@@ -173,7 +173,7 @@ function getStream(process, stream, {encoding, buffer, maxBuffer}) {
 function makeError(result, options) {
 	const {stdout, stderr, code, signal} = result;
 	let {error} = result;
-	const {joinedCommand, timedOut, isCanceled, parsed: {options: {timeout}}} = options;
+	const {joinedCommand, timedOut, isCanceled, killed, parsed: {options: {timeout}}} = options;
 
 	const [exitCodeName, exitCode] = getCode(result, code);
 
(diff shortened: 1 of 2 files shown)


sindresorhus:

sindresorhus:
> On Windows (with just exitCode: 1) it seems impossible to know whether the child process was killed externally.

Should we open an Node.js issue about that?

ehmicky:
Merge conflict fixed.

I have opened [an issue on Node](https://github.com/nodejs/node/issues/27642). However I think this PR could still be merged regardless: if that issue resolves, we can create a new PR so that `error.killed` is `true` on external signal termination.

PR #229, opened 2019-05-10
Make `error.killed` and `error.isCanceled` always boolean (not `undefined`)

ehmicky:
The `killed` and `isCanceled` properties of `error` should always be `true` or `false`, not `undefined`. This makes the interface simpler and more predictable.

This PR solves the use cases where it could have been `undefined`, e.g. when using `execa.sync()`.

It also adds related tests.


sindresorhus:

ehmicky:
Fixed.

Issue #228, opened 2019-05-10
Improve stream error handling


ehmicky:
When `maxBuffer` is reached, the rejected error has a different shape than usual:

```js
{
	"name": "MaxBufferError",
	"bufferedData": "...",
	"stream": "stdout"
}
```

Shouldn't it have the same shape? For example:

```js
// No `code`, `exitCode`, `exitCodeName`, `signal`
{
	// If available
	stdout,
	stderr, 
	all,

	// This indicates a stream error
	stream: "stdout",
	// This indicates a stream error due to `maxBuffer`
	maxBuffer: true|false,

	failed: true,
	timedOut: false,
	isCanceled: false,
	
	command: "...",
}
```

`stdout`, `stderr` and `all` could either:
  a) be present only if the whole stream finished before the error
  b) show the partially `bufferedData`

The error message could be made more descriptive as well.

I can submit a PR if you think that's a good idea.

sindresorhus:
Yes. Good catch. That's really just an oversight. PR welcome.

> b) show the partially bufferedData

👍 Just make sure to document this behavior in the readme/TS.

PR #230, opened 2019-05-10
Make error messages more consistent

ehmicky:
This makes `error.message` more consistent.

`execa` errors in the following cases:
  a) exit code is defined and is not `0`
  b) child process `error` event, or `stdin` `error` event
  c) signal termination (including `cancel()`)
  d) timeout
  e) stream error (separate issue: #228)

a) c) d) e) look consistent. However b):
  - prints `stdout` and `stderr`, as opposed to the others. This can also lead to unexpectedly (very) long error messages, which might create issues for some users. `error.stdout` and `error.stderr` can be used instead.
  - does not print the command, as opposed to the others.

This adjusts b) error message to the same format used by the other errors.

It turns out to cover this, I just had to remove unit tests but did not need to add new ones (could not think of one not already covered).

### DIFF
--- a/fixtures/error-message.js
+++ b/fixtures/error-message.js
@@ -1,5 +1,0 @@
-#!/usr/bin/env node
-'use strict';
-console.log('stdout');
-console.error('stderr');
-process.exit(1);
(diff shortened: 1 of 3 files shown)


ehmicky:
CI is now passing.

PR #239, opened 2019-05-12
Improve the API section of the readme

ehmicky:
This improves the `API` section:
  - the description of each property of the return/resolved value (or thrown/rejected error) was missing.
  - the return value of `execa.sync()` was wrong (it is not the same as the return value of `spawnSync()` although it has some common properties.
  - the extra methods and properties of the `child_process` instance returned by the main function are now separate.
  - the description of the main function has been improved.

To be more efficient, I will update the `index.d.ts` file after a first code review.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -122,9 +122,9 @@ Arguments should not be escaped nor quoted, except inside `command` where spaces
 
 Unless the [`shell`](#shell) option is used, no shell interpreter (Bash, `cmd.exe`, etc.) is used, so shell features such as variables substitution (`echo $PATH`) are not allowed.
 
-Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess). That instance:
+Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) which:
   - is also a `Promise` which resolves or rejects with a [`childProcessResult`](#childProcessResult).
-  - expose the following additional methods and properties.
+  - exposes the following additional methods and properties.
 
 #### cancel()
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -112,36 +112,106 @@ try {
...
 ### execa.sync(command, [options])
 
 Execute a file synchronously.
 
-Returns the same result object as [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).
+Returns or throws a [`childProcessResult`](#childProcessResult).
+
+### childProcessResult
+
+Type: `object|Error`
```
```suggestion
Type: `object`
```

The `Error` is thrown, not returned, so it's not the result.

readme.md
```diff
@@ -112,36 +112,106 @@ try {
...
+
+#### exitCodeName
+
+Type: `string`
+
+The textual exit code of the process that was run.
+
+#### stdout
+
+Type: `string|Buffer`
```
```suggestion
Type: `string | Buffer`
```

readme.md
```diff
@@ -112,36 +112,106 @@ try {
...
 
-Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) which is enhanced to be a `Promise`.
+Returns a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) which:
+  - is also a `Promise` resolving or rejecting with a [`childProcessResult`](#childProcessResult).
+  - exposes the following additional methods and properties.
+
+#### cancel()
 
-It exposes an additional `.all` stream, with `stdout` and `stderr` interleaved.
+Same as [`childProcess.kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) but with a different error message and with [`childProcessResult.isCanceled`](#iscanceled) set to `true`.
```
sindresorhus: We should tell users to prefer this instead of `.kill()`.

ehmicky:
Fixed! I've also fixed the `index.d.ts`.

sindresorhus:
Looks good :)

PR #243, opened 2019-05-14
Allow transformation of stream output

ammarbinfaisal:
I would update the tests and the docs; before that just wanted to ask whether there are anymore built-in transforming options needed?   
_I hope I understood the issue description correctly_

### DIFF
--- a/index.js
+++ b/index.js
@@ -114,6 +114,15 @@ function handleOutput(options, value) {
 		value = stripFinalNewline(value);
 	}
 
+	const {transform} = options;
+	if (value && transform) {
+		if (transform === 'line') {
+			value = value.split('\n');
+		} else if (typeof transform === 'function') {
+			value = transform(value);
+		}
+	}
+
 	return value;
 }
 


ehmicky:
Hi @ammarbinfaisal, thanks a lot for the time spent on a new PR since last one (#189)!

Now I am wondering, wouldn't it be more straightforward (and easier to compose) for users to just do the following?

```js
const { stdout } = execa('echo', ['hello', 'world'])
const value = myTransform(stdout) 
```

What do you think @ammarbinfaisal and @sindresorhus?

ammarbinfaisal:
I guess this would make more sense if the observable support mentioned in #20 is implemented. Unfortunately I've never used observables. btw you'd know better

ehmicky:
I might be understanding #20 but it seems like it's more about adding support for Observables. 

When it comes to transforming the `stdout` and/or `stderr`, in buffered mode, this can be done easily at the moment (see code above). In streaming mode, it can also be done easily:

```js
execa('echo', ['hello', 'world']).stdout.pipe(myStreamTransform)
```

ammarbinfaisal:
So this PR and that issue should be closed?🙄

ehmicky:
Unless there is something I am not understanding about Observables, I think adding this option might already be easily done. Based on that, I would close this PR.

However I would like to know @sindresorhus opinion on this. Also feel free to let me know where I might be wrong.

sindresorhus:
Yes, the original intention was for use with Observables.

PR #246, opened 2019-05-14
Make `stdout` and `stderr` properties `undefined` instead of `null`

ehmicky:
The `stdout` and `stderr` properties can be `undefined` with `execa()`, e.g. when the option `stdout: 'ignore'` was used.

However with `execa.sync()`, it is `null` instead. This change increases consistency by setting those to `undefined` instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,8 +110,12 @@ function handleInput(spawned, input) {
 }
 
 function handleOutput(options, value) {
-	if (value && options.stripFinalNewline) {
-		value = stripFinalNewline(value);
+	if (typeof value !== 'string' && !Buffer.isBuffer(value)) {
+		return;
+	}
+
+	if (options.stripFinalNewline) {
+		return stripFinalNewline(value);
 	}
 
 	return value;
--- a/index.js
+++ b/index.js
@@ -109,9 +109,10 @@ function handleInput(spawned, input) {
 	}
 }
 
-function handleOutput(options, value) {
+function handleOutput(options, value, error) {
 	if (typeof value !== 'string' && !Buffer.isBuffer(value)) {
-		return;
+		// When `execa.sync()` errors, we normalize it to '' to mimic `execa()`
+		return error === undefined ? undefined : '';
 	}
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -109,8 +109,6 @@ try {
 		command: 'wrong command',
 		exitCode: 2,
 		exitCodeName: 'ENOENT',
-		stdout: null,
-		stderr: null,
```
sindresorhus: Shouldn't these be an empty string instead, like the async version? https://github.com/sindresorhus/execa/pull/246/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R72

ehmicky: Fixed: https://github.com/sindresorhus/execa/pull/246/commits/ace8ef2d6b9d82a87c6b1c7ea173d69f4a8c8db0

index.js
```diff
@@ -109,9 +109,10 @@ function handleInput(spawned, input) {
 	}
 }
 
-function handleOutput(options, value) {
+function handleOutput(options, value, error) {
 	if (typeof value !== 'string' && !Buffer.isBuffer(value)) {
-		return;
+		// When `execa.sync()` errors, we normalize it to '' to mimic `execa()`
+		return error === undefined ? undefined : '';
```
sindresorhus: Can you add a test for this?

ehmicky: I've just added two tests. :+1:

ehmicky:
Fixed.

sindresorhus:
Tests are failing on Windows.

ehmicky:
The Windows tests are fixed on CI now.

The issue is that `exec.sync()` should be an empty string when the child process emits an `error` event (as opposed to throwing an exception right away). The only way I manage to trigger that event is by firing a command that does not exist. I even looked up the code in core Node.js and libuv to get ideas on how to trigger this, but I get no ideas. And firing a command that does not exist produces some output on Windows, but not on Unix. But I fixed the test accordingly.

PR #247, opened 2019-05-14
Make tests more consistent (code refactoring)

ehmicky:
This is a code refactoring PR that is only style-related. 
  - it makes the style of tests more consistent with each other.
  - it uses destructuring whenever possible.

Also:
  - it tests `error.exitCode` (new property) instead of `error.code` (deprecated property)
  - it splits one test into two (because it was two tests combined in one)

### DIFF
--- a/test.js
+++ b/test.js
@@ -206,7 +206,9 @@ test('opts.stdout:ignore - stdout will not collect data', async t => {
 
 test('helpful error trying to provide an input stream in sync mode', t => {
 	t.throws(
-		() => execa.sync('stdin', {input: new stream.PassThrough()}),
+		() => {
+			execa.sync('stdin', {input: new stream.PassThrough()});
+		},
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -204,17 +204,15 @@ test('opts.stdout:ignore - stdout will not collect data', async t => {
 
 test('helpful error trying to provide an input stream in sync mode', t => {
 	t.throws(
-		() => {
-			execa.sync('stdin', {input: new stream.PassThrough()});
-		},
+		() => execa.sync('stdin', {input: new stream.PassThrough()}),
```
sindresorhus: This was intentional. I don't like using inline arrow functions when the return value is not used as it makes the code less readable.

ehmicky: Yes you're right, this is stricter

sindresorhus:

PR #248, opened 2019-05-14
Make `error.killed` more consistent

ehmicky:
`error.killed` is always `true` or `false` with `execa()`.

But with `execa.sync()`, it can also be `undefined` sometimes, e.g. on `ENOENT`. This PR ensures `error.killed` is always a boolean. 

Details:
  - `spawn()` returns a `childProcess` with [`childProcess.killed`](https://github.com/nodejs/node/blob/master/lib/internal/child_process.js#L446)
  - `spawnSync()` does not return this, but we can emulate most of the cases by checking the `signal` property.

Tests are added for this as well.

### DIFF
--- a/test.js
+++ b/test.js
@@ -318,6 +318,16 @@ test('result.killed is false if not killed, in sync mode', t => {
 	t.false(result.killed);
 });
 
+test('result.killed is false on process error', async t => {
+	const {killed} = await t.throwsAsync(execa('wrong command'));
+	t.false(killed);
+});
+
+test('result.killed is false on process error, in sync mode', t => {
+	const {killed} = t.throws(() => execa.sync('wrong command'));
+	t.false(killed);
+});
+
 if (process.platform === 'darwin') {
 	test.cb('sanity check: child_process.exec also has killed.false if killed indirectly', t => {
 		const cp = childProcess.exec('forever', error => {
--- a/test.js
+++ b/test.js
@@ -141,7 +141,9 @@ test('stripFinalNewline in sync mode', t => {
 });
 
 test('stripFinalNewline in sync mode on failure', t => {
-	const {stderr} = t.throws(() => execa.sync('noop-throw', ['foo'], {stripFinalNewline: true}));
+	const {stderr} = t.throws(() => {
+		execa.sync('noop-throw', ['foo'], {stripFinalNewline: true});
+	});
 	t.is(stderr, 'foo');
 });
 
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -318,6 +318,16 @@ test('result.killed is false if not killed, in sync mode', t => {
 	t.false(result.killed);
 });
 
+test('result.killed is false on process error', async t => {
+	const {killed} = await t.throwsAsync(execa('wrong command'));
+	t.false(killed);
+});
+
+test('result.killed is false on process error, in sync mode', t => {
+	const {killed} = t.throws(() => execa.sync('wrong command'));
```
sindresorhus: Don't use inline arrow function here.

sindresorhus:
Unrelated to this PR, but what do you think about renaming `.killed` to `.isKilled` for naming consistency?

sindresorhus:
> spawnSync() does not return this, but we can emulate most of the cases by checking the signal property.

It's weird that the synchronous version has this inconsistency. Should we open a Node.js issue about it?

test.js
ehmicky: Should I use an inline non-arrow function instead? Should I use a `try`/`catch` block?

sindresorhus: No, it should be like:

```js
const {killed} = t.throws(() => {
	execa.sync('wrong command');
});
```

ehmicky:
> It's weird that the synchronous version has this inconsistency. Should we open a Node.js issue about it?

That's because `spawn()` return a `ChildProcess` instance, while `spawnSync()` returns a plain object with fewer/different properties. It's not only `killed`. I think Node.js would say it's a design decision.

> Unrelated to this PR, but what do you think about renaming .killed to .isKilled for naming consistency?

`isKilled` is more consistent with `isCanceled`. However `spawned.killed` already exists in core Node.js, so I would think using property names users are already expecting or familiar with is more important.

test.js
ehmicky: Fixed. I also fixed it in two other places where this was done in the tests.

sindresorhus:
> isKilled is more consistent with isCanceled. However spawned.killed already exists in core Node.js, so I would think using property names users are already expecting or familiar with is more important.

You could argue that about `.code` vs `exitCode` though.

ehmicky:
Yes I thought of that as well. The thing is `.code` is actually different from `exitCode` because they don't have the same type. Whereas `.killed` is almost identical both in type and behavior.

sindresorhus:
Good point. Nah, we can keep it as-is.

PR #249, opened 2019-05-14
Fix `error.timedOut` not working with `execa.sync()`

ehmicky:
When `execa.sync()` times out because of the `timeout` option, `error.timedOut` is still `false`.

This PR fixes this.

### DIFF
--- a/test.js
+++ b/test.js
@@ -394,7 +394,9 @@ test('timeout kills the process if it times out', async t => {
 });
 
 test('timeout kills the process if it times out, in sync mode', async t => {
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -393,6 +393,12 @@ test('timeout kills the process if it times out', async t => {
 	t.true(error.timedOut);
 });
 
+test('timeout kills the process if it times out, in sync mode', async t => {
+	const {killed, timedOut} = await t.throws(() => execa.sync('forever', {timeout: 1, message: TIMEOUT_REGEXP}));
```
```suggestion
	const {killed, timedOut} = await t.throws(() => {
		execa.sync('forever', {timeout: 1, message: TIMEOUT_REGEXP});
	});
```

ehmicky:
Fixed.

PR #250, opened 2019-05-14
Remove `error.code`

ehmicky:
`error.code` is deprecated in favor of `error.exitCode` and `error.exitCodeName`. This PR removes it from the TypeScript definition.

I am not sure whether unknown properties are an issue with TypeScript. If they are, feel free to close this PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -242,10 +242,6 @@ declare namespace execa {
 
 	interface ExecaSyncReturnValue<StdoutErrorType = string>
 		extends ExecaReturnBase<StdoutErrorType> {
-		/**
-		The exit code of the process that was run.
-		*/
-		code: number;
 	}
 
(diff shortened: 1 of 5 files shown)


sindresorhus:
Let's just remove it completely. You were right. It's confusing and weird to have it now that we have the other properties.

sindresorhus:
https://github.com/sindresorhus/execa/blob/1f427fd71a515377a19640a3fc8a7e7a9f9b746d/index.js#L425 ?

ehmicky:
That's because for some reason `spawnSync()` returns the exit code with a `signal` property name, while `spawn()` use a `code` property name. We:
  - normalize both to `result.code`
  - use `result.code` to set `result.exitCode` and `result.exitCodeName`
  - delete `result.code`

I agree it looks bad though, I'm going to refactor it.

PR #256, opened 2019-05-18
Separate core options from execa-specific options

### LINKED ISSUE #252: Separate `child_process` options from `execa` options in the `README`
I think we should divide the `Options` section in the `README` in two parts:
  - first the options that are specific to `execa`.
  - then the options that we simply forward to `child_process.spawn()`. We can also add a sentence saying those are the same options as `child_process.spawn()`.

This would help users that are already familiar with `child_process.spawn()` and only want to know which extra options `execa` brings.

We could also fix `index.d.ts` to follow the same division.

If you think that's a good idea, I can submit a PR.

ehmicky:
Fixes #252 

I've also re-ordered the options in `index.test-d.ts` and `index.d.ts` accordingly.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -311,7 +311,7 @@ Set to `false` if you don't want to extend the environment variables when provid
 
 ---
 
-Execa also accepts the below options which are the same options for [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
+Execa also accepts the below options which are the same as the options for [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
 
 #### cwd
 
--- a/readme.md
+++ b/readme.md
@@ -309,9 +309,9 @@ Default: `true`
 
 Set to `false` if you don't want to extend the environment variables when providing the `env` property.
 
-### options from child_process#spawn and child_process#exec
+---
 
-The same options as both [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) and [`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) can also be used.
+Execa also accepts the below options which are passed directly to [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
 
 #### cwd
 
--- a/readme.md
+++ b/readme.md
@@ -311,7 +311,7 @@ Set to `false` if you don't want to extend the environment variables when provid
 
 ---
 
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -250,6 +309,24 @@ Default: `true`
 
 Set to `false` if you don't want to extend the environment variables when providing the `env` property.
 
+### options from child_process#spawn and child_process#exec
```
sindresorhus: Feels weird to have a heading here. It kinda makes it unclear how to use these options. I would rather have it like this:

```
---

Execa also accepts the below options which are passed directly to `child_process#spawn()`/`child_process#exec()`
---
```

sindresorhus: > `child_process#exec()`

We don't actually use `exec()` internally, so not sure why it's mentioned here.

ehmicky: We don't use it, but we re-implement its options (`timeout`, `maxBuffer`, `killSignal`).

sindresorhus: Ok. I still think we should make that clearer, with something like a proposed above.

ehmicky: Fixed in https://github.com/sindresorhus/execa/pull/256/commits/38c57c6f0a2718efd875d04cb04d25abf9353694

ehmicky:
Fixed. :)

readme.md
```diff
@@ -309,9 +309,9 @@ Default: `true`
 
 Set to `false` if you don't want to extend the environment variables when providing the `env` property.
 
-### options from child_process#spawn and child_process#exec
+---
 
-The same options as both [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) and [`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) can also be used.
+Execa also accepts the below options which are passed directly to [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
```
sindresorhus: Hmm. I still don't like the wording.

> Execa also accepts the below options which are the same as the options for [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)

?

ehmicky: Fixed in https://github.com/sindresorhus/execa/pull/256/commits/cc96aa23a03ba938dee98ea5e800b86df1828b27

readme.md
```diff
@@ -311,7 +311,7 @@ Set to `false` if you don't want to extend the environment variables when provid
 
 ---
 
-Execa also accepts the below options which are passed directly to [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
+Execa also accepts the below options which are the same options for [`child_process#spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)/[`child_process#exec()`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
```
sindresorhus: This is not exactly what I wrote.

ehmicky: Fixed in https://github.com/sindresorhus/execa/pull/256/commits/b18609495c02d580fc59871f86d6eb71f105b02d

PR #261, opened 2019-05-21
Add `execa.command()`

ehmicky:
This PR is a follow-up on #182 which allowed specifying the command and its arguments as a single string. 

It improves the common case where the program contains a space in its path. For example on Windows `C:\Program Files\node\node.exe`. 

With `execa(command, [array], [options])`, when:
  1) `command` contains a space
  2) `array` is defined
  3) `options.shell` is not `true`

Then:
  - former behavior: an exception is thrown.
  - new behavior: `command` is handled as a command with a space in its path.
 
```js
// Success
execa('echo hello')
// Fail
execa('echo hello', ['world'])
// Success
execa('echo hello', {shell: true})

// Fail
execa('C:\\Program Files\\node\\node.exe')
// Success
execa('C:\\Program Files\\node\\node.exe', [])
// Success
execa('C:\\Program\\ Files\\node\\node.exe')
```

I've fixed the tests. No changes need to be done to the `README` since it already implies that no spaces escaping is needed for programs containing a space in their path.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -368,6 +368,37 @@ declare const execa: {
 		file: string,
 		options?: execa.SyncOptions<null>
 	): execa.ExecaSyncReturnValue<Buffer>;
+
+	/**
+	Same as `execa()` except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as `execa.command('echo unicorns')`.
+
+	If the file or an argument contains spaces, they must be escaped with backslashes. Otherwise no escaping/quoting is needed.
+
+	@param command - The program/script to execute and its arguments.
+	@returns A [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess), which is enhanced to also be a `Promise` for a result `Object` with `stdout` and `stderr` properties.
+
+	@example
+	```
+	import execa from 'execa';
+
+	(async () => {
+		const {stdout} = await execa.command('echo unicorns');
+		console.log(stdout);
+		//=> 'unicorns'
+	})();
+	```
+	*/
+	command(command: string, options?: execa.Options): execa.ExecaChildProcess;
+	command(command: string, options?: execa.Options<null>): execa.ExecaChildProcess<Buffer>;
+
+	/**
+	Same as `execa.command()` but synchronously.
+
+	@param command - The program/script to execute and its arguments.
+	@returns A result `Object` with `stdout` and `stderr` properties.
+	*/
+	commandSync(command: string, options?: execa.SyncOptions): execa.ExecaSyncReturnValue;
+	commandSync(command: string, options?: execa.SyncOptions<null>): execa.ExecaSyncReturnValue<Buffer>;
 };
 
 export = execa;
--- a/index.d.ts
+++ b/index.d.ts
@@ -392,7 +392,7 @@ declare const execa: {
 	command(command: string, options?: execa.Options<null>): execa.ExecaChildProcess<Buffer>;
 
 	/**
-	Same as `execa.command()` but synchronously.
+	Same as `execa.command()` but synchronous.
 
 	@param command - The program/script to execute and its arguments.
 	@returns A result `Object` with `stdout` and `stderr` properties.
--- a/index.d.ts
+++ b/index.d.ts
@@ -372,7 +372,7 @@ declare const execa: {
 	/**
 	Same as `execa()` except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as `execa.command('echo unicorns')`.
 
-	If the file or an argument contains spaces, they must be escaped with backslashes. Otherwise no escaping/quoting is needed.
+	If the file or an argument contains spaces, they must be escaped with backslashes. This matters especially if `command` is not a constant but a variable, for example with `__dirname` or `process.cwd()`. Except for spaces, no escaping/quoting is needed.
 
 	@param command - The program/script to execute and its arguments.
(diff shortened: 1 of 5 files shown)


GMartigny:
I think this should be documented somewhere that `[]` is required as second argument when your command has spaces in it.

Weird behavior I see while testing:
If I want to escape myself the command `execa("fixtures/command\\ with\\ space")` it fail anyway. Maybe it's not an expected behavior, just saying.

ehmicky:
Good point! Fixed.

GMartigny:
I confirm that there's a bug with path with more than one spaces:

```js
execa("path/with/one\\ space"); // Success
execa("path/with/more \\than\\ one\\ space"); // Fail
```

This is due to [`previousToken` not being accurate](https://github.com/sindresorhus/execa/pull/261/files#diff-168726dbe96b3ce427e7fedce31bb0bcR26) when ignoring some item of the array.

ehmicky:
This is a different bug. I've opened a PR at #262 to fix it.

test.js
```diff
@@ -107,10 +107,13 @@ test('allow string arguments in synchronous mode', t => {
...
-	t.throws(() => {
-		execa('node fixtures/echo foo bar', ['foo', 'bar']);
-	}, /Arguments cannot be inside/);
+test('allow commands with spaces and array arguments', async t => {
+	const {stdout} = await execa('./fixtures/command with space', ['foo', 'bar']);
+	t.is(stdout, 'foo\nbar');
+});
+
+test('forbid commands with spaces and no array arguments', async t => {
+	await t.throwsAsync(execa('./fixtures/command with space'));
```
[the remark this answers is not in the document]
ehmicky: Yes I thought this might be problematic as well as soon as @GMartigny mentioned `C:\Program File\node\node.exe`...

If that's ok with you, I will update this PR with:
  - `execa(file, [args], [options)` (previous behavior, no `command` logic)
  - `execa.command(command, [options])`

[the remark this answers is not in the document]
sindresorhus: Sounds good 👍

ehmicky:
Done. I've separated the `command` feature into two new methods `execa.command()` and `execa.commandSync()`.

cliffordfajardo:
@ehmicky - would this new `command` method allow users to use pipe symbol inside their command like so: `execa.command('ls | head -n 5')`?

Side question:
If not, how would one go about doing something similar to that in the current API? 

Thanks!

ehmicky:
@cliffordfajardo this is not the point of this PR. `execa.command()` will just be a convenience method over `execa()`. 

In you case, you need to use the [`shell` option](https://github.com/sindresorhus/execa#shell).

Or even better, you don't use a shell at all. For example, you could use [`readdirp`](https://github.com/paulmillr/readdirp) and pipe the result in Node.js. 

If you still have questions, please open an issue though, as this is not related to this PR.

readme.md
```diff
@@ -147,12 +142,23 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
...
 
+### execa.command(command, [options])
+
+Same as [`execa()`](#execafile-arguments-options) except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as `execa.command('echo unicorns')`.
+
+If the file or an argument contains spaces, they must be escaped with backslashes. Otherwise no escaping/quoting is needed.
+
+### execa.commandSync(command, [options])
+
+Same as [`execa.command()`](#execacommand-command-options) but synchronously.
```
```suggestion
Same as [`execa.command()`](#execacommand-command-options) but synchronous.
```

readme.md
```diff
@@ -147,12 +142,23 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
...
 
 Execute a file synchronously.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### execa.command(command, [options])
+
+Same as [`execa()`](#execafile-arguments-options) except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as `execa.command('echo unicorns')`.
+
+If the file or an argument contains spaces, they must be escaped with backslashes. Otherwise no escaping/quoting is needed.
```
sindresorhus: Maybe add a note that this is especially important with template literal interpolation when passing in a file path from a variable, for example, `__dirname` or `process.cwd()` where the file path could contain spaces.

index.d.ts
```diff
@@ -368,6 +368,37 @@ declare const execa: {
...
+		console.log(stdout);
+		//=> 'unicorns'
+	})();
+	```
+	*/
+	command(command: string, options?: execa.Options): execa.ExecaChildProcess;
+	command(command: string, options?: execa.Options<null>): execa.ExecaChildProcess<Buffer>;
+
+	/**
+	Same as `execa.command()` but synchronously.
```
```suggestion
	Same as `execa.command()` but synchronous.
```

ehmicky:
Fixed!

readme.md
```diff
@@ -147,12 +142,23 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
...
 
 Execute a file synchronously.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### execa.command(command, [options])
+
+Same as [`execa()`](#execafile-arguments-options) except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as `execa.command('echo unicorns')`.
+
+If the file or an argument contains spaces, they must be escaped with backslashes. This matters especially if `command` is not a constant but a variable, for example with `__dirname` or `process.cwd()`. Except for spaces, no escaping/quoting is needed.
```
sindresorhus: Don't forget to add this to the TS definition too.

ehmicky: Fixed in 5d2bc3b224755001e076004c97e12c63301656ea

ehmicky:
Fixed.

PR #262, opened 2019-05-21
Fix not being able to escape more than one space in commands

ehmicky:
`execa('path\\ space')` works, but not `execa('path\\ with\\ multiple\\ spaces')`. This PR fixes this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,11 +25,11 @@ function handleEscaping(tokens, token, index) {
 
 	const previousToken = tokens[tokens.length - 1];
 
-	if (!previousToken.endsWith('\\')) {
-		return [...tokens, token];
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -23,13 +23,13 @@ function handleEscaping(tokens, token, index) {
 		return [token];
 	}
 
-	const previousToken = tokens[index - 1];
+	const previousToken = tokens[tokens.length - 1];
 
 	if (!previousToken.endsWith('\\')) {
```
GMartigny: Trivial: Removing the negation `!` would make this condition and its effect clearer.

GMartigny:
Looks good

PR #267, opened 2019-05-30
Add support for `finally()` in Node <10

ehmicky:
This adds support for the `finally()` method in Node <10. Previously this was only available in Node >=10.


sindresorhus:
I don't really see the point of this. Users wouldn't expect `.finally()` to work when it doesn't work elsewhere.

ehmicky:
Good point. I'm closing this

PR #270, opened 2019-05-30
Child process errors should reject the promise right away

ehmicky:
With #157 and #199 we had the following issue: the top-level promise would not be rejected right away when timing out. It would instead wait for stdout/stderr streams to close first.

We fixed this, but the same problem is still happening with child process errors. If a child process receives an `error` event, we should reject the promise right away, and not wait for stdout/stderr streams to close first (which might never happen when using a long-running command).

This PR fixes this. As an added bonus, it improves the deferred pattern introduced by #199 that @sindresorhus was [warning against](https://github.com/sindresorhus/execa/pull/199/files#r273337376).

### DIFF
--- a/index.js
+++ b/index.js
@@ -259,7 +259,7 @@ const execa = (file, args, options) => {
 	}
 
 	// TODO: Use native "finally" syntax when targeting Node.js 10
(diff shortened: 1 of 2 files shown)


sindresorhus:
👌

PR #280, opened 2019-06-10
Refactor `kill()` again

ehmicky:
This is a follow-up to #272 adding more refactoring to `kill()`.

@zokker13

### DIFF
--- a/index.js
+++ b/index.js
@@ -226,7 +226,9 @@ function setKillTimeout(kill, signal, options, killResult) {
 	}
 
 	const timeout = getForceKillAfterTimeout(options);
-	setTimeout(() => kill('SIGKILL'), timeout).unref();
+	setTimeout(() => {
+		kill('SIGKILL');
+	}, timeout).unref();
 }
 
 function shouldForceKill(signal, {forceKill}, killResult) {
--- a/index.js
+++ b/index.js
@@ -242,7 +242,7 @@ function isSigterm(signal) {
 
 function getForceKillAfterTimeout({forceKillAfter = DEFAULT_FORCE_KILL_TIMEOUT}) {
 	if (!Number.isInteger(forceKillAfter) || forceKillAfter < 0) {
-		throw new TypeError(`Option 'forceKillAfter' ${forceKillAfter} should be a positive integer`);
+		throw new TypeError(`Expected the \`forceKillAfter\` option to be a non-negative integer, got \`${forceKillAfter}\` (${typeof forceKillAfter})`);
 	}
 
 	return forceKillAfter;
--- a/index.js
+++ b/index.js
@@ -224,19 +224,28 @@ function setKillTimeout(kill, signal, options, killResult) {
 		return;
 	}
 
-	const forceKillAfter = Number.isInteger(options.forceKillAfter) ?
-		options.forceKillAfter :
-		5000;
-	setTimeout(() => kill('SIGKILL'), forceKillAfter).unref();
+	const timeout = getForceKillAfterTimeout(options);
+	setTimeout(() => kill('SIGKILL'), timeout).unref();
 }
 
-function shouldForceKill(signal, options, killResult) {
-	return ((typeof signal === 'string' &&
-		signal.toUpperCase() === 'SIGTERM') ||
-		signal === os.constants.signals.SIGTERM) &&
-		options.forceKill !== false &&
-		killResult;
+function shouldForceKill(signal, {forceKill}, killResult) {
+	return isSigterm(signal) && forceKill !== false && killResult;
 }
+
+function isSigterm(signal) {
+	return signal === os.constants.signals.SIGTERM ||
+		(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM');
+}
+
+function getForceKillAfterTimeout({forceKillAfter}) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -224,20 +224,29 @@ function setKillTimeout(kill, signal, options, killResult) {
...
-		signal === os.constants.signals.SIGTERM) &&
-		options.forceKill !== false &&
-		killResult;
+function shouldForceKill(signal, {forceKill}, killResult) {
+	return isSigterm(signal) && forceKill !== false && killResult;
 }
 
+function isSigterm(signal) {
+	return signal === os.constants.signals.SIGTERM ||
+		(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM');
```
sindresorhus: Does Node.js support lowercase for signals? If not, I think we should rather throw than just silently allow it.

ehmicky: Yes it does: https://github.com/nodejs/node/blob/04633eeeb93c5a86a6635940eb4f34a7839ada11/lib/internal/util.js#L225

ehmicky:
Ready for another round of code review.

index.js
```diff
@@ -224,20 +224,29 @@ function setKillTimeout(kill, signal, options, killResult) {
...
+
+function getForceKillAfterTimeout({forceKillAfter}) {
+	if (Number.isInteger(forceKillAfter)) {
+		return forceKillAfter;
+	}
+
+	return DEFAULT_FORCE_KILL_TIMEOUT;
+}
+
+const DEFAULT_FORCE_KILL_TIMEOUT = 5000;
```
ptusch: Shouldnt this be placed somewhere at the top of the file? Not sure about styling (obviously is valid) but seeing random constants in the sourcecode seems rather untidy to me.

zokker13: Shouldnt this be placed somewhere at the top of the file? Not sure about styling (obviously is valid) but seeing random constants in the sourcecode seems rather untidy to me.

ehmicky: Fixed: 88d280c6e7c861e5c718ed037c9636ab8695ff08

index.js
```diff
@@ -224,18 +225,25 @@ function setKillTimeout(kill, signal, options, killResult) {
 		return;
 	}
 
-	const forceKillAfter = Number.isInteger(options.forceKillAfter) ?
-		options.forceKillAfter :
-		5000;
-	setTimeout(() => kill('SIGKILL'), forceKillAfter).unref();
+	const timeout = getForceKillAfterTimeout(options);
+	setTimeout(() => kill('SIGKILL'), timeout).unref();
```
sindresorhus: I only use inline arrow functions if the return value is used, otherwise it makes the code harder to read.

```suggestion
	setTimeout(() => {
		kill('SIGKILL');
	}, timeout).unref();
```

index.js
```diff
@@ -224,18 +225,25 @@ function setKillTimeout(kill, signal, options, killResult) {
...
+	return isSigterm(signal) && forceKill !== false && killResult;
+}
+
+function isSigterm(signal) {
+	return signal === os.constants.signals.SIGTERM ||
+		(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM');
+}
+
+function getForceKillAfterTimeout({forceKillAfter}) {
+	if (Number.isInteger(forceKillAfter)) {
```
sindresorhus: Instead of silently falling back to the default if the `forceKillAfter` value is not an integer, I would rather throw an error. Invalid input should not be silently ignored.

ehmicky:
Fixed.

index.js
```diff
@@ -224,18 +225,27 @@ function setKillTimeout(kill, signal, options, killResult) {
...
+}
+
+function isSigterm(signal) {
+	return signal === os.constants.signals.SIGTERM ||
+		(typeof signal === 'string' && signal.toUpperCase() === 'SIGTERM');
+}
+
+function getForceKillAfterTimeout({forceKillAfter = DEFAULT_FORCE_KILL_TIMEOUT}) {
+	if (!Number.isInteger(forceKillAfter) || forceKillAfter < 0) {
+		throw new TypeError(`Option 'forceKillAfter' ${forceKillAfter} should be a positive integer`);
```
sindresorhus: 0 is not positive, so we cannot use that wording.

```suggestion
		throw new TypeError(`Expected the \`forceKillAfter\` option to be a non-negative integer, got \`${forceKillAfter}\` (${typeof forceKillAfter})`);
```

test.js
```diff
@@ -175,6 +175,14 @@ if (process.platform !== 'win32') {
 		const {signal} = await t.throwsAsync(subprocess);
 		t.is(signal, 'SIGKILL');
 	});
+
+	test('.kill() `forceKillAfter` should not be a float', t => {
+		t.throws(() => execa('noop').kill('SIGTERM', {forceKillAfter: 0.5}), TypeError);
```
```suggestion
		t.throws(() => {
			execa('noop').kill('SIGTERM', {forceKillAfter: 0.5});
		}, TypeError);
```

sindresorhus: You should also check that it's the correct error `/non-negative integer/`

sindresorhus:
> You should also check that it's the correct error /non-negative integer/

ehmicky:
Fixed.

PR #282, opened 2019-06-11
Fix variable name bug (`master` failing)

ehmicky:
#278 renamed `joinedCommand` to `command`

But #277 that just got merged added a new `joinedCommand` variable. I forgot to rebase the PR after #278  was merged. This leads to CI failing.

This PR fixes that.


sindresorhus:
Feel free to just commit tiny commits like this directly to master.

PR #284, opened 2019-06-12
Refactor tests of `kill()`

ehmicky:
This is a follow-up to #272 and #280 adding more refactoring to `kill()`:
  - remove some newlines in tests
  - improve test titles
  - add one test (when `forceKill` is `true`)
  - fix description in typings not matching the description in `readme`

@zokker13

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -288,7 +288,7 @@ declare namespace execa {
 		forceKill?: boolean;
 
 		/**
-		Milliseconds to wait for the child process to terminate before sending a `SIGKILL` signal.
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -288,7 +288,7 @@ declare namespace execa {
 		forceKill?: boolean;
 
 		/**
-		Milliseconds to wait for the child process to terminate before sending a `SIGKILL` signal.
+		Milliseconds to wait for the child process to terminate before sending `SIGKILL`.
```
ehmicky: Make it match the description in `README`

PR #286, opened 2019-06-14
Change `maxBuffer` default value from `10 MB` to `100 MB`

ehmicky:
This increases the `maxBuffer` default value.

Some users encountered issues with processing images. The limit is really only meant to protect against abuse, but `100 MB` is not abuse.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -159,7 +159,7 @@ declare namespace execa {
 		readonly timeout?: number;
 
 		/**
-		Largest amount of data in bytes allowed on `stdout` or `stderr`. Default: 10MB.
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -159,7 +159,7 @@ declare namespace execa {
 		readonly timeout?: number;
 
 		/**
-		Largest amount of data in bytes allowed on `stdout` or `stderr`. Default: 10MB.
+		Largest amount of data in bytes allowed on `stdout` or `stderr`. Default: 100 MB.
 
 		@default 10000000
```
sindresorhus: This must be updated too. You can use underscores.
```suggestion
		@default 100_000_000
```

ehmicky:
Fixed.

PR #293, opened 2019-06-17
Add a sentence about job search

ehmicky:
I am looking for a new position. This adds a promotional sentence at the beginning of the `readme` about it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -8,7 +8,7 @@
 
 ---
 
-One of the maintainers [@ehmicky](https://github.com/ehmicky) is looking for a remote full-time position as a Node.js back-end lead, feel free to [contact him](https://www.mickael-hebert.com)!
+One of the maintainers [@ehmicky](https://github.com/ehmicky) is looking for a remote full-time position as a Node.js back-end lead. Feel free to [contact him](https://www.mickael-hebert.com)!
 
 ---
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
You're mentioning back-end, but you're only looking for a Node.js position, right? If so, I would be explicit about that.

ehmicky:
Good point, fixed. I reworded to `Node.js back-end`, because most of the job ads with the word `Node.js developer` are actually full-stack jobs (where Node.js is just used for the build tools).

readme.md
```diff
@@ -6,6 +6,13 @@
 > Process execution for humans
 
 
+---
+
+One of the maintainers [@ehmicky](https://github.com/ehmicky) is looking for a remote full-time position as a Node.js back-end lead. Feel free to [contact him](https://www.mickael-hebert.com)!
```
sindresorhus: I would link directly to the contact form:

```suggestion
One of the maintainers [@ehmicky](https://github.com/ehmicky) is looking for a remote full-time position as a Node.js back-end lead. Feel free to [contact him](https://www.mickael-hebert.com#contact)!
```

ehmicky: I actually would like them to check out my website before contacting me. That's because I get many requests for roles that do not match (not remote, not Node.js/JavaScript or not back-end).

sindresorhus: Alright

PR #294, opened 2019-06-17
Refactor `timeout` option

ehmicky:
This refactors the `timeout` option handling. This does not change behavior.

With this PR and the upcoming ones, I am trying to split up the main `execa()` function into smaller functions. The main function partially relies on closure state, so I had to introduce a `context` object passed around instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -325,6 +325,15 @@ function handleSpawned(spawned, context) {
 	});
 }
 
+const setupTimeout = function (spawned, {timeout, killSignal}, context) {
+	if (timeout > 0) {
+		context.timeoutId = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -325,6 +325,15 @@ function handleSpawned(spawned, context) {
 	});
 }
 
+const setupTimeout = function (spawned, {timeout, killSignal}, context) {
```
```suggestion
const setupTimeout = (spawned, {timeout, killSignal}, context) => {
```

sindresorhus: I prefer using `const` and arrow functions for everything.

ehmicky: Alright I will also send a follow-up PR to do that on all functions in the repository.

ehmicky:
Fixed.

PR #296, opened 2019-06-18
Fix functions coding style

ehmicky:
This makes all functions use the `const funcName = (...) => {...}` coding style.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ const DEFAULT_FORCE_KILL_TIMEOUT = 1000 * 5;
 
 const SPACES_REGEXP = / +/g;
 
-function handleArgs(file, args, options = {}) {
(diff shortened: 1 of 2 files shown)


sindresorhus:

ehmicky:
Fixed.

kirillgroshkov:
Any reference to why this style is preferred? (just curious for myself)

ehmicky:
I don't think it matters too much, what matters for me most is consistency. Actually this is @sindresorhus preferred style, I am using a [different style](https://github.com/ehmicky/log-process-errors/blob/master/src/level.js#L7) for my other projects.

Also I don't like to rely on hoisting.

sindresorhus:
Mostly for consistency, more sane hoisting rules, and ability to have inline arrow functions statements.

PR #298, opened 2019-06-19
Allow the `shell` option with `execa.node()`

ehmicky:
As discussed before, users should avoid the `shell` option whenever possible.

However I don't think the `shell` option is worse with `execa.node()` than with `execa()`. For example, some users might want to do things like:

```js
execa.node('file.js', ['$RANDOM'], {shell: true})
```

This PR removes the fact that `shell` is always `false` with `execa.node()`. This makes:
  - `execa.node()` more orthogonal with the rest of the features
  - `execa.node()` documentation a little simpler

Let me know what you think of this.

@GMartigny

### DIFF
--- a/index.js
+++ b/index.js
@@ -544,8 +544,7 @@ module.exports.node = (scriptPath, args, options) => {
 			stdin: undefined,
 			stdout: undefined,
 			stderr: undefined,
-			stdio: stdioOption,
-			shell: false
+			stdio: stdioOption
 		}
 	);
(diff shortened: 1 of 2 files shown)


sindresorhus:
I strongly disagree with this. There's no good reason to run `node` as a shell, and even if users insist, they can just use `execa()` with the `shell` option.

ehmicky:
Alright, I'm closing this.

PR #299, opened 2019-06-19
Rename `nodeArguments` to `nodeOptions`

ehmicky:
I think the name `nodeArguments` makes it unclear it's about Node CLI flags, as opposed to the arguments passed to `node scriptPath.js ...`. 

Also this makes it more consistent with the way Node.js calls it [in the documentation](https://nodejs.org/api/cli.html).

@GMartigny


sindresorhus:
Yes, agreed 👍

PR #304, opened 2019-06-19
Document that `execa.node()` `stdio` default to `pipe` instead of `inherit`

ehmicky:
This was missing from the `readme`.


sindresorhus:
Huh?

> 'pipe' - equivalent to ['pipe', 'pipe', 'pipe'] **(the default)** - https://nodejs.org/api/child_process.html#child_process_options_stdio

ehmicky:
Sorry you're right, I got mixed up. `execa.fork()` defaults to `inherit`, but we decided not to keep that behavior. However the current `execa.node()` code still uses that, so I will submit another PR to remove it.

PR #309, opened 2019-06-20
Improve/refactor `.cancel()`

ehmicky:
This separates `cancel()` into its own function.

Also this makes `cancel()` return a `boolean` like `kill()` does. The fact that `kill()` returns a boolean is [not documented](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) and is not part of [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/child_process.d.ts#L21) so I think we should do the same, and not document it nor type it. However it would not hurt to still forward `kill()` return value.

### DIFF
--- a/index.js
+++ b/index.js
@@ -308,8 +308,6 @@ const spawnedCancel = (spawned, context) => {
 	if (killResult) {
 		context.isCanceled = true;
 	}
-
-	return killResult;
 };
 
 const handleSpawned = (spawned, context) => {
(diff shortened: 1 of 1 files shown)


sindresorhus:
> The fact that kill() returns a boolean is not documented

You should open a Node.js issue about either documenting it or removing that return value.

> However it would not hurt to still forward kill() return value.

I disagree. It breaks the least amount of surprise rule. We should not do undocumented stuff. I think we should not return a boolean. If anyone needs it, they will request it and we can consider adding it officially.

---

So I don't think cancel should return anything either, at least for now.

ehmicky:
Fixed.

PR #313, opened 2019-06-23
Make `@types/node` a production `dependency` instead of a `devDependency`

ehmicky:
I think TypeScript types dependencies should be production dependencies not development dependencies. 
Please correct me if I'm wrong. Otherwise TypeScript users would need to manually install `@types/node` themselves to properly use `execa` types.

This also seems to be what the [official documentation says](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies) and what one of the TypeScript guys is [saying here](https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338).

@BendingBender and @sindresorhus what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
 		"local"
 	],
 	"dependencies": {
+		"@types/node": "^12.0.7",
 		"cross-spawn": "^6.0.5",
 		"get-stream": "^5.0.0",
 		"is-stream": "^2.0.0",
@@ -48,7 +49,6 @@
 		"strip-final-newline": "^2.0.0"
 	},
 	"devDependencies": {
-		"@types/node": "^12.0.7",
 		"ava": "^2.1.0",
 		"coveralls": "^3.0.4",
 		"is-running": "^2.1.0",


sindresorhus:
I don't have time to read the whole thread in https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338, but that's definitely correct or wanted. What if different packages depend on different major versions of `@types/node`? It would cause a mess. The user is expected to have `@types/node` themselves anyway if they use TS. And if they forgot, they get a nice warning because of https://github.com/sindresorhus/execa/blob/4dd258d062722e69317707da5c1b73ffa506d42a/index.d.ts#L1

ehmicky:
Yes you're right.

It seems to me actually `@types/node` is a peer dependency, as we rely on the consumer to install this. However it's an optional one, since consumer might not use TypeScript. However optional peer dependencies are not supported by npm at the moment.

ehmicky:
Just by curiosity, is there any difference between using `import 'node';` and `/// <reference types="node"/>` here?

`@types/node` uses ambient modules, which can also be loaded with ES `import`.

I checked it and it does work. When doing `--traceResolution` I can see that `import` is first checking if `node_modules/node.{ts,tsx,d.ts}`, which can't happen, so it does make few extra I/O calls there. On the other hand, it's more consistent with ES modules.

Issue #196, opened 2019-03-28
preferLocal prevents local from working


arcanis:
https://github.com/sindresorhus/execa/issues/191#issuecomment-477650346

> Yarn already sets the PATH when running scripts so that the right Yarn binary (and all other binaries that your application should have access to) is always in the PATH. But because `execa` modifies it as well (and it does it without actually knowing which paths it should use, it only makes assumptions), it ends up overriding the Yarn-generated folder. In the end, the wrong files are called.
> 
> I would strongly suggest defaulting it to `false`, as it's very intrusive and can lead to pretty hard to debug issues (I myself spent an indecent amount of time trying to figure out what happened before I started to consider that maybe execa was doing something strange).

## Example:

**package.json**

```json
{
  "dependencies": {
    "execa": "^1.0.0"
  },
  "scripts": {
    "foo": "yarn --version",
    "bar": "node ./example.js"
  }
}
```

**example.js**

```js
const execa = require('execa');

execa('yarn', ['--version']).then(v => {
  console.log(v.stdout);
});
```

**command line**

```
# npm installs the bins in the same folder as Node, so `node` and `yarn` are siblings
$> npm install -g yarn@1.15.1

# This command enforces the Yarn version inside this one project
$> yarn policies set-version 1.15.2

$> yarn foo
1.15.2

$> yarn bar
1.15.1
```

In this case, you see that Yarn rightfully enforced the use of the 1.15.2 release when calling `foo`. Unfortunately, since execa overrides the PATH with `dirname(node)`, it also overrode in the process the Yarn binary that we set in the path, which causes the wrong one to be used.

It's even more of a problem with [Plug'n'Play](https://yarnpkg.com/lang/en/docs/pnp), since we inject the path to the binaries within the PATH (rather than always injecting the same `node_modules/.bin` folder). A similar problem then occurs where if you previously ran `npm install -g babel-cli`, then your execa scripts will use it instead of the local one.

ehmicky:
Thanks for reporting this.

Could you please describe in few words what `yarn policies set-version` is doing under the hood? For example how it manipulates `$PATH`, or anything else.

arcanis:
[...]

Issue #153, opened 2018-11-02
execa does not respect $PATH variable


lucasbraun:
I have two installations of git and `execa` seems to pick the wrong one, not respecting `$PATH` correctly, while `child-process` does it correctly (even if I explicitly remove `/usr/bin` from `$PTAH` and add my desired path twice, once in the beginning and once in the end):

```
$ /usr/bin/git --version 
git version 1.8.3.1
$ /opt/rh/rh-git29/root/usr/bin/git --version
git version 2.9.3
$ echo $PATH
/opt/rh/rh-git29/root/usr/bin:/bin:/usr/local/bin:/opt/rh/rh-git29/root/usr/bin
$ cat ./index.js
const execa = require('execa');
(async () => {
    const {stdout} = await execa('git', ['--version']);
    console.log(stdout);
    //=> 'unicorns'
})();
const execFile = require('child_process').execFile;
const child = execFile('git', ['--version'], (error, stdout, stderr) => {
    if (error) {
        console.error('stderr', stderr);
        throw error;
    }
    console.log('stdout', stdout);

});
$ node ./index.js
git version 1.8.3.1
stdout git version 2.9.3
```

I am missing something here?

ammarbinfaisal:
@lucasbraun This doesn't appear to be problem with execa. execa is an alternative to childProcess.spawn(actually better than using it) and you shouldn't compare execFile's output with execa. Try using childProcess.spawn instead of execa and it would show the same result. Also, have a look at this nodejs/node#12986

ehmicky:
@ammarbinfaisal `childProcess.spawn` is the base for all other `childProcess` methods variants including `execFile()`. See the [Node.js source](https://github.com/nodejs/node/blob/master/lib/child_process.js#L176).

ehmicky:
@lucasbraun There has been few changes to the code handling `env` lately, can you use the latest commit (not latest release) of execa and confirm this is still a bug when ?

Also could you try the following code instead?

```js
const execa = require('execa');
const childProcess = require('child_process');

console.log(execa.sync('echo $PATH', { shell: true }).stdout)
console.log(childProcess.spawnSync('echo $PATH', { shell: true, encoding: 'utf8' }).stdout)
```

lucasbraun:
[...]

PR #314, opened 2019-06-24
Change default value of the `preferLocal` option to `false`

ehmicky:
This changes the default value of the `preferLocal` option to `false`.

This option causes several issues for some users: #196, #153.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,7 +27,7 @@ declare namespace execa {
 
 		If you `$ npm install foo`, you can then `execa('foo')`.
 
-		@default true
+		@default false
 		*/
 		readonly preferLocal?: boolean;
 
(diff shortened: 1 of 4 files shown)


ehmicky:
CI tests are working now.

sindresorhus:
For posterity, here's my reasoning for doing this change:
- Most cases don't actually spawn local binaries from `node_modules`.
    I realized that most of the cases are global binaries and local ones with `{cwd: __dirname}`.
- The few times you actually need it, it might be more readable to make it explicit.
- It causes surprising behavior then because:
    1. Yarn mocks with the PATH too.
    2. Some Node.js installation has a `node_modules` directory or `npm` directory that will be used instead of what the user actually wanted. This can happen silently and cause confusion.

So basically, I think we should default to the least surprising behavior.

PR #328, opened 2019-06-26
Do not mutate process promise

ehmicky:
When a child process ends, we return [a promise](https://github.com/sindresorhus/execa/blob/master/index.js#L106). If that promise is awaited several times, each `await` will receive the same resolved|rejected object (which is normal). 

However that object is being [directly modified](https://github.com/sindresorhus/execa/blob/master/index.js#L166) (the `stdout`, `stderr` and `all` properties are added), which means the next `await` will re-use the modifications of the previous `await`. 

I could not find a precise bug this is creating, but this definitely might introduce problems. It also makes debugging a little trickier (which is how I encounter this issue first). I fixed it by avoiding those mutating assignments.

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,14 +162,19 @@ const execa = (file, args, options) => {
 	});
 
 	const handlePromise = async () => {
-		const [result, stdout, stderr, all] = await getSpawnedResult(spawned, parsed.options, processDone);
-		result.stdout = handleOutput(parsed.options, stdout);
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you add a regression test?

ehmicky:
I don't think it's possible because the mutated promise (the `promiseDone` variable) cannot be accessed from outside `execa()`. Also it does not create any bug yet. But it might in the future.

sindresorhus:
👍

PR #332, opened 2019-06-27
Improve tests for the `timeout` option

ehmicky:
One of the tests for the `timeout` option was wrong: the `message` validation was passed as argument to `execa` instead of `t.throwsAsync()`.

This PR also adds a test checking `timeout` when it's set to `0`.


sindresorhus:
Can you fix the conflict?

ehmicky:
Fixed.

PR #333, opened 2019-06-27
Refactor the `timeout` option

ehmicky:
At the moment the `timeout` option works like this:
  - we share and mutate a `context.timedOut` variable around to several functions.
  - on timeout, we set `context.timedOut` to `true` and kill the child process.
  - we expect the child process to emit an `exit` event right after being killed. Inside the event handler, if `context.timedOut` is `true`, we reject the `spawnedPromise` instead of resolving it.

This PR takes a different approach:
  - on timeout, we reject a `timeoutPromise` and kill the child process. 
  - we await `Promise.race([timeoutPromise, spawnedPromise])` instead of `spawnedPromise` (providing a timeout is used)

Which means we are using promise races instead of passing information around with a mutable variable. Although it adds few extra lines of code, I think this is cleaner. It decouples the "childProcess events -> promise" logic and the timeout logic. By removing half of the `context` variable, it also allows further refactoring down the road. Let me know what you think about this.

This PR also validates that the `timeout` option is a non-negative integer.

### DIFF
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -1,6 +1,7 @@
 'use strict';
 const os = require('os');
 const onExit = require('signal-exit');
+const pFinally = require('p-finally');
 
 const DEFAULT_FORCE_KILL_TIMEOUT = 1000 * 5;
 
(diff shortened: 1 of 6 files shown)


lib/kill.js
```diff
@@ -52,14 +53,33 @@ const spawnedCancel = (spawned, context) => {
...
+	if (timeout === 0 || timeout === undefined) {
+		return spawnedPromise;
+	}
+
+	if (!Number.isInteger(timeout) || timeout < 0) {
+		throw new TypeError(`Expected the \`timeout\` option to be a non-negative integer, got \`${timeout}\` (${typeof timeout})`);
 	}
+
+	let timeoutId;
+	const timeoutPromise = new Promise((resolve, reject) => {
```
ehmicky: We unfortunately have to use `new Promise` because `promisify(setTimeout)` does not return the `timeoutId`. However we need that `timeoutId` to clean up the timeout when the child process exits normally.

lib/kill.js
```diff
@@ -52,14 +53,33 @@ const spawnedCancel = (spawned, context) => {
...
+		timeoutId = setTimeout(() => {
+			timeoutKill(spawned, killSignal, reject);
+		}, timeout);
+	});
+
+	const safeSpawnedPromise = pFinally(spawnedPromise, () => {
+		clearTimeout(timeoutId);
+	});
+
+	return Promise.race([timeoutPromise, safeSpawnedPromise]);
```
ehmicky: That timeout pattern (timeout promise + `Promise.race` + timeout cleanup) might be already available, e.g. [`await-timeout`](https://github.com/vitalets/await-timeout). Maybe we should use this instead.

sindresorhus:
I definitely prefer the immutability approach of passing it around rather tha mutating.

ehmicky:
Me too, I'm a big functional JavaScript fan, but I also know not everyone is.

PR #345, opened 2019-07-02
Add `result.all` TypeScript definition

### LINKED ISSUE #342: `all` is missing on `ExecaChildProcess`
The `all` combined stream is missing on https://github.com/sindresorhus/execa/blob/c9b4d09cc70f80ed8423bd81fbd2c5849512c3e0/index.d.ts#L310

ehmicky:
Fixes #342

While the `all` property is defined on the resolved value or rejected error (where it's a string or buffer), we forgot it on the return value (where it's a stream).

@tomsotte

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -168,6 +168,8 @@ Type: `ReadableStream | undefined`
 
 Stream combining/interleaving [`stdout`](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout) and [`stderr`](https://nodejs.org/api/child_process.html#child_process_subprocess_stderr).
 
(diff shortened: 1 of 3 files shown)


tomsotte:
As per [`merge-stream`](https://github.com/grncdr/merge-stream/blob/a217797ff4a52581423ca1fc6eef409c4c992447/index.js#L7) current implementation the mixed `all` stream that we create it's a [PassThrough](https://nodejs.org/api/stream.html#stream_class_stream_passthrough).
For `execa` purpose I believe it's correct to limit its type to a Readable stream.

ehmicky:
Yes I think that it being a `PassThrough` is an implementation detail. From the user's perspective, it should be manipulated like `stdout` and `stderr` which are typed as `Readable` streams.

readme.md
```diff
@@ -164,6 +164,8 @@ Similar to [`childProcess.kill()`](https://nodejs.org/api/child_process.html#chi
 
 #### all
 
+Type: `ReadableStream | undefined`
```
sindresorhus: I think we should make it clear in what situations it could be `undefined`.

ehmicky:
Fixed.

PR #360, opened 2019-08-09
Fix errors being thrown when `detached: true` or `cleanup: false` is used

ehmicky:
Using either the `detached: true` and `cleanup: false` always throws an error:

```js
> await execa('echo',{cleanup: false})
Thrown:
TypeError: Cannot destructure property `error` of 'undefined' or 'null'.
    at handlePromise (/home/ether/code/execa/index.js:107:10)
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
> await execa('echo',{detached: true})
Thrown:
TypeError: Cannot destructure property `error` of 'undefined' or 'null'.
    at handlePromise (/home/ether/code/execa/index.js:107:10)
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
```

This was introduced by one my PRs [here](https://github.com/sindresorhus/execa/pull/337/files#diff-afe5430c779eb982786eccb39864cea0R86).

This is a quite critical bug, we should release this. However the current code contains breaking changes: users of the `all` property must now use the `all: true` option. So this would be a major release.

Additionally some tests that check `t.throwsAsync()` were incorrect but previously successful due to that bug being present (since it made `execa` fail but for the wrong reason). I have fixed those tests (mostly Windows issues).

### DIFF
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -85,7 +85,7 @@ const setupTimeout = (spawned, {timeout, killSignal = 'SIGTERM'}, spawnedPromise
 // `cleanup` option handling
 const setExitHandler = (spawned, {cleanup, detached}, timedPromise) => {
 	if (!cleanup || detached) {
-		return;
+		return timedPromise;
 	}
 
 	const removeExitHandler = onExit(() => {
--- a/test/fixtures/sub-process-exit
+++ b/test/fixtures/sub-process-exit
@@ -4,4 +4,14 @@ const execa = require('../..');
 
 const cleanup = process.argv[2] === 'true';
 const detached = process.argv[3] === 'true';
-execa('node', ['./test/fixtures/noop'], {cleanup, detached});
+
+const runChild = async () => {
+	try {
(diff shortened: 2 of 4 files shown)


sindresorhus:
> This is a quite critical bug, we should release this. However the current code contains breaking changes: users of the all property must now use the all: true option. So this would be a major release.

I don't think we should release the `all` change yet. I would prefer to resolve https://github.com/sindresorhus/execa/pull/351 first. What we could do is to create a branch based on v2.0.3, apply the one-line fix, and do a v2.0.4 release to an alternative dist-tag. ?

ehmicky:
Would the following be simpler?
  - revert #353 (`all` option breaking change)
  - release `2.0.4`
  - add #353 back

sindresorhus:
I prefer doing it from a branch. That way the main commit history doesn't get the revert noise.

Done: https://github.com/sindresorhus/execa/releases/tag/v2.0.4

---

> and do a v2.0.4 release to an alternative dist-tag. ?

Scratch the alternative dist-tag part. That was wrong.

PR #373, opened 2019-10-06
Add `error.originalMessage` property

ehmicky:
When an `error` event is thrown on the child process or when a timeout happens, we re-use the original `Error` instance instead of creating a new one. However we override the `error.message`.

This PR keep the original `error.message` as an additional property `error.originalMessage`.

I need this for one my projects using Execa. That project should proxy a user's command in such a way that users should not be aware of that proxy. In particular, the error messages produced by Execa should be not printed to the console. However the error messages coming from `error` events on the child process (such as `command not found`) should be printed. This PR allows for this to happen.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -262,6 +262,11 @@ declare namespace execa {
 		The signal that was used to terminate the process.
 		*/
 		signal?: string;
+
+		/**
+		Original error message when the child process exits due to an error or a timeout.
+		*/
+		originalMessage?: string;
 	}
 
 	interface ExecaSyncReturnValue<StdoutErrorType = string>
--- a/index.d.ts
+++ b/index.d.ts
@@ -262,11 +262,6 @@ declare namespace execa {
 		The signal that was used to terminate the process.
 		*/
 		signal?: string;
-
-		/**
-		Original error message when the child process exits due to an error or a timeout.
-		*/
-		originalMessage?: string;
 	}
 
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -262,6 +262,11 @@ declare namespace execa {
 		The signal that was used to terminate the process.
 		*/
 		signal?: string;
+
+		/**
+		Original error message when the child process exits due to an error or a timeout.
+		*/
+		originalMessage?: string;
```
sindresorhus: I don't think it should be placed here as it's only available on errors, not also the normal return value.

index.d.ts
```diff
@@ -262,6 +262,11 @@ declare namespace execa {
 		The signal that was used to terminate the process.
 		*/
 		signal?: string;
+
+		/**
+		Original error message when the child process exits due to an error or a timeout.
```
sindresorhus: Can you describe here how it's different from `.message`? (and in the readme)

Also in what scenarios it's `undefined`.

index.d.ts
ehmicky: Fixed in b0ba95aa885ce7531a5b21c09a2c47686dec98b0

index.d.ts
ehmicky: Done in 7290b6bdccd68f7327fa74d29a222fe60aecc91a

ehmicky:
Fixed :)

PR #377, opened 2019-10-14
Add `execPath` option

### LINKED ISSUE #153: execa does not respect $PATH variable

### LINKED ISSUE #196: preferLocal prevents local from working

ehmicky:
Fixes #153 
Fixes #196

This adds an `execPath` option to modify which Node.js executable to use in child processes.

See [here](https://github.com/sindresorhus/npm-run-path/pull/9#issuecomment-541437833) for some use cases.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,6 +41,8 @@ declare namespace execa {
 		/**
 		Path to the Node.js executable to use in child processes.
 
+		For example this can be used together with `get-node` to run a specific Node.js version in a child process.
+
 		This can be either an absolute path or a path relative to the `cwd` option.
 
 		Requires `preferLocal` to be `true`.
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,6 +39,17 @@ declare namespace execa {
 		readonly localDir?: string;
 
 		/**
+		Path to the Node.js executable to use in child processes.
+
+		This can be either an absolute path or a path relative to the `cwd` option.
+
+		Requires `preferLocal` to be `true`.
+
+		@default process.execPath
+		*/
+		readonly execPath?: string;
+
+		/**
 		Buffer the output from the spawned process. When set to `false`, you must read the output of `stdout` and `stderr` (or `all` if the `all` option is `true`). Otherwise the returned promise will not be resolved/rejected.
 
 		If the spawned process fails, `error.stdout`, `error.stderr`, and `error.all` will contain the buffered data.
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,11 +41,11 @@ declare namespace execa {
 		/**
 		Path to the Node.js executable to use in child processes.
 
-		For example this can be used together with `get-node` to run a specific Node.js version in a child process.
-
 		This can be either an absolute path or a path relative to the `cwd` option.
 
 		Requires `preferLocal` to be `true`.
+		
+		For example, this can be used together with [`get-node`](https://github.com/ehmicky/get-node) to run a specific Node.js version in a child process.
 
 		@default process.execPath
 		*/
--- a/index.js
+++ b/index.js
@@ -14,11 +14,11 @@ const {joinCommand, parseCommand} = require('./lib/command.js');
 
 const DEFAULT_MAX_BUFFER = 1000 * 1000 * 100;
 
-const getEnv = ({env: envOption, extendEnv, preferLocal, localDir}) => {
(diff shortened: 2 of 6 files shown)


index.d.ts
```diff
@@ -38,6 +38,17 @@ declare namespace execa {
 		*/
 		readonly localDir?: string;
 
+		/**
+		Path to the Node.js executable to use in child processes.
+
+		This can be either an absolute path or a path relative to the `cwd` option.
+
+		Requires `preferLocal` to be `true`.
```
sindresorhus: Why does it require `preferLocal` to be `true`? Maybe include a short reasoning here.

index.d.ts
```diff
@@ -38,6 +38,17 @@ declare namespace execa {
 		*/
 		readonly localDir?: string;
 
+		/**
+		Path to the Node.js executable to use in child processes.
+
```
sindresorhus: Would be great to include the use-cases for this option.

index.d.ts
ehmicky: It's because the feature is handled by `npm-run-path` which is run only when the `preferLocal` option is `true`. Since this is an internal reason (the way our dependencies are structured), we should we still document it?

index.d.ts
ehmicky: Done in 734450e6922e917ef5835a63b8edbaf47e43ce6c

ehmicky: Done in 11029d47b183dbb625fe1af5a8d7037218d296b1

index.d.ts
sindresorhus: Nah, it's fine.

PR #378, opened 2019-10-14
Add `error.signalDescription`

ehmicky:
At the moment we set `error.signal` to either `undefined` or the signal that terminated a process such as `"SIGFPE"`. This PR adds `error.signalDescription` which is a human-friendly description of that signal such as `"Floating point arithmetic error"`.

This also enhances error messages accordingly such as:

```
Command was killed with SIGFPE (Floating point arithmetic error)
```

instead of:

```
Command was killed with SIGFPE
```

This uses [`human-signals`](https://github.com/ehmicky/human-signals), a library I wrote which exports a simple plain object where the key is the signal name and the value the signal description. This works cross-platform.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -267,12 +267,16 @@ declare namespace execa {
 		killed: boolean;
 
 		/**
-		The name of the signal that was used to terminate the process.
+		The name of the signal that was used to terminate the process. For example `"SIGFPE"`.
+
+		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 		*/
 		signal?: string;
 
 		/**
-		A human-friendly description of the signal that was used to terminate the process.
+		A human-friendly description of the signal that was used to terminate the process. For example `"Floating point arithmetic error"`.
+
+		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 		*/
 		signalDescription?: string;
 	}
--- a/index.d.ts
+++ b/index.d.ts
@@ -276,7 +276,7 @@ declare namespace execa {
 		/**
 		A human-friendly description of the signal that was used to terminate the process. For example, `Floating point arithmetic error`.
 
-		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
+		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`. It is also `undefined` when the signal is very uncommon which should seldomly happen.
 		*/
 		signalDescription?: string;
 	}
--- a/index.d.ts
+++ b/index.d.ts
@@ -267,14 +267,14 @@ declare namespace execa {
 		killed: boolean;
 
 		/**
-		The name of the signal that was used to terminate the process. For example `"SIGFPE"`.
+		The name of the signal that was used to terminate the process. For example, `SIGFPE`.
 
 		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 		*/
 		signal?: string;
 
 		/**
-		A human-friendly description of the signal that was used to terminate the process. For example `"Floating point arithmetic error"`.
+		A human-friendly description of the signal that was used to terminate the process. For example, `Floating point arithmetic error`.
 
 		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 		*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -267,9 +267,14 @@ declare namespace execa {
 		killed: boolean;
 
 		/**
-		The signal that was used to terminate the process.
+		The name of the signal that was used to terminate the process.
 		*/
 		signal?: string;
+
+		/**
+		A human-friendly description of the signal that was used to terminate the process.
+		*/
+		signalDescription?: string;
 	}
 
 	interface ExecaSyncReturnValue<StdoutErrorType = string>
--- a/readme.md
+++ b/readme.md
@@ -270,13 +270,17 @@ Whether the process was killed.
 
 Type: `string | undefined`
 
-The name of the signal that was used to terminate the process.
+The name of the signal that was used to terminate the process. For example `"SIGFPE"`.
+
+If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 
 #### signalDescription
 
 Type: `string | undefined`
 
-A human-friendly description of the signal that was used to terminate the process.
+A human-friendly description of the signal that was used to terminate the process. For example `"Floating point arithmetic error"`.
+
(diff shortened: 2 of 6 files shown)


index.d.ts
```diff
@@ -254,9 +254,14 @@ declare namespace execa {
 		killed: boolean;
 
 		/**
-		The signal that was used to terminate the process.
+		The name of the signal that was used to terminate the process.
 		*/
 		signal?: string;
+
+		/**
+		A human-friendly description of the signal that was used to terminate the process.
```
sindresorhus: Can you mention that it's already included in the default Execa error message? So users don't try to be smart and include it themselves and getting it double.

Can you also include a short example of what the description looks like for a signal? (readme too)

test/error.js
```diff
@@ -133,6 +133,15 @@ if (process.platform !== 'win32') {
 		t.is(signal, 'SIGINT');
 	});
 
+	test('error.signalDescription is defined', async t => {
+		const cp = execa('noop');
```
sindresorhus: Let's try to use `subprocess` instead of `cp` everywhere. Maybe we should add it to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prevent-abbreviations.md

index.d.ts
```diff
@@ -254,9 +254,14 @@ declare namespace execa {
 		killed: boolean;
 
 		/**
-		The signal that was used to terminate the process.
+		The name of the signal that was used to terminate the process.
```
sindresorhus: Unrelated to this PR, but I just noticed, when something can be `undefined`, we should try to describe in what scenarios it's defined vs undefined. This can be a big help for the user in knowing when it will be used.

index.d.ts
sindresorhus: We should also be explicit about that there might be signals with a `signalDescription`. So users should assume that `signalDescription` will be defined if `signal` is.

readme.md
```diff
@@ -270,7 +270,13 @@ Whether the process was killed.
 
 Type: `string | undefined`
 
-The signal that was used to terminate the process.
+The name of the signal that was used to terminate the process.
+
+#### signalDescription
+
+Type: `string | undefined`
```
sindresorhus: Should be start using
```suggestion
Type: `string?`
```

?

sindresorhus:
`human-signals` is great, but it kinda feels like something that should be built-in Node.js. Worth opening an issue on the Node.js issue tracker about that?

sindresorhus:
`human-signals` brings in https://github.com/zloirock/core-js as a dependency, which I don't think is intentional. If it is, that's not great as it's huge.

ehmicky:
> human-signals brings in zloirock/core-js as a dependency, which I don't think is intentional. If it is, that's not great as it's huge.

That's a tough one as I'm using Babel to transpile the library. I've tried to find an option or plugin to inline Babel helpers instead of having Babel inject `require('core-js/...')` statements bu I could not find any. Do you know if that's possible?

sindresorhus:
Why are you even transpiling? Seems like you have a lot of tooling and output bloat/complexity just to be able to use import/export (as the target Node.js version supports most other things). All you really need is https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

ehmicky:
Yes it's true the only thing I am using is ES modules there. The thing is I am using this setup with all my projects (which are of different sizes), so I can use other things like `Array.flatMap()`, `Promise.finally()`, `Promise.allSettled()`, async iterators, etc.

However on a small library like this, this does look rather silly. 

Actually `core-js` is only injected by Babel in this project because I use array spread operators. That's supported by Node 8, but unfortunately `Array.values()` is not, and `core-js` [bundles the two features together at the moment](https://github.com/zloirock/core-js/issues/670).

I am looking into the best way to remove the `core-js` dependency in `human-signals` so that library can be used by Execa. Will come back soon on this PR once I'm done with it.

test/error.js
ehmicky: Done in 1a5d1cb421129a60b2dd61d809704bc906fcac1d

I also opened a PR #379 to fix it in all other tests.

readme.md
ehmicky: We have other properties in the `README` that are optional but have more one than type:

```md
#### all

Type: `string | Buffer | undefined`
```

```md
#### stdin (same for stdout, stderr)

Type: `string | number | Stream | undefined`
```

Should those stay the same or be `string? | Buffer?`?

index.d.ts
ehmicky: Fixed in a92abc0c1ed0e32868dcc5ed8b9106116001e838

ehmicky:
Fixed.

I removed the `core-js` dependency and resolved the points above (with one discussion pending: https://github.com/sindresorhus/execa/pull/378#discussion_r335405193).

index.d.ts
```diff
@@ -254,9 +254,18 @@ declare namespace execa {
...
+		The name of the signal that was used to terminate the process. For example `"SIGFPE"`.
+
+		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
 		*/
 		signal?: string;
+
+		/**
+		A human-friendly description of the signal that was used to terminate the process. For example `"Floating point arithmetic error"`.
+
+		If a signal terminated the process, this property is defined and included in the error message. Otherwise it is `undefined`.
```
sindresorhus: This is not entirely correct as there might be a signal that we don't have a description for.

readme.md
sindresorhus: Good point. Thinking more about this, it might be too convoluted and TypeScript doesn't even support `?` for return types. Lets forget it for now.

index.d.ts
ehmicky: Fixed in cf3161a09dcff8a38fb9f31681c3af077090988a

ehmicky:
> Seeing as you don't include the actual source in the package, the source map files are useless

The sources are included in the source map files in the `sourcesContent` property: https://unpkg.com/human-signals@1.1.1/build/src/main.js.map

> They are also not very valuable as Node.js doesn't use source maps.

By default they are not used for error stack traces. But this can be fixed with either:
  - something like [`node-source-map-support`](https://github.com/evanw/node-source-map-support) which monkey patches `error.stack`
  - the experimental [`--enable-source-maps`](https://nodejs.org/api/cli.html#cli_enable_source_maps) Node CLI flag

It's also used by debuggers. For example if users are debugging some code using [`--inspect`](https://nodejs.org/api/cli.html#cli_inspect_host_port) and Chrome DevTools then stepping into a minified dependency, that dependency will show the source code if source maps are defined.

sindresorhus:
> The sources are included in the source map files in the sourcesContent property: unpkg.com/human-signals@1.1.1/build/src/main.js.map

Oh, nice. I didn't know you could embed the source.

PR #388, opened 2019-11-09
Allow setting the `windowsHide` option

### LINKED ISSUE #371: Consider allowing windowsHide option
See nodejs/node#29837 for why this option can make a practical difference. I saw the previous discussion on #242 that was closed because there wasn't a known reason to allow it.

ehmicky:
Fixes #371. 

At the moment, we force the `windowsHide` option to be `true`. However this prevents `CTRL-C` from working on Windows. 

This PR allows `windowsHide` to be set to `false`. It keeps the default value as `true`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -198,6 +198,13 @@ declare namespace execa {
 		@default false
 		*/
 		readonly windowsVerbatimArguments?: boolean;
+
+		/**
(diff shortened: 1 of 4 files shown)


papb:
Ahhh!! So there is a reason Ctrl+C always seemed buggy for me :man_facepalming:

devmondo:
don't know if I am doing something wrong, but still `ctrl-c` does not work for me even when setting the option to false!

running latest `execa` & windows 10 & node 12

thanks in advanced.

PR #397, opened 2019-12-07
Add stdout/stderr to error messages

### LINKED ISSUE #395: Include stdout/stderr in `error.message`
We removed it in `execa` v2 for some reason, but it's very useful.

The lack of it leads to a lot of annoyance and makes it harder for users to see what actually went wrong...

It also makes it harder to do test assertions like this:

```js
test('missing file operands', async t => {
	await t.throwsAsync(execa('./cli.js'), /`source` and `destination` required/);
});
```

Open questions:
- Should we include `stdout` or only `stderr`?
- Should we truncate it at some length?

ehmicky:
Fixes #395.

This adds stderr and stdout (in that order) to error messages. 

stdout and stderr are not added when they are not available, for example when they are streamed (`stdio: 'inherit'`).

A new property `error.shortMessage` is added to retrieve error messages without stdout/stderr in case users need shorter error messages.

This PR includes tests, documentation, TypeScript types.

Examples for `error.errno` were also adjusted to match Node 13 output.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,7 @@ declare namespace execa {
 		This can be either an absolute path or a path relative to the `cwd` option.
 
 		Requires `preferLocal` to be `true`.
-		
+
 		For example, this can be used together with [`get-node`](https://github.com/ehmicky/get-node) to run a specific Node.js version in a child process.
 
 		@default process.execPath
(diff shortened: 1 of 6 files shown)


sindresorhus:
Should we maybe also update the `.message` docs to explicitly say it includes stdout/stderr? We could also mention which order they're added, that they're separated by a newline and that they're not interleaved.

ehmicky:
Sounds good. Fixed in 29a4737.

Issue #399, opened 2019-12-18
TypeError: setTimeout(...).unref is not a function


mifi:
I use this module all the time including in Electron. After upgrading to the newest version i get this error when trying to call childprocess.kill() from an Electron application:

```
TypeError: setTimeout(...).unref is not a function
setKillTimeout
.../node_modules/execa/lib/kill.js:23:14
```

Note: with execa@1 it works. But 2 or 3 fails with this error. Maybe we could add a check for whether unref exists before calling it, or is Electron not supported anymore?

mifi:
I created a PR to fix

ehmicky:
Thanks a lot @mifi!

sindresorhus:
Fixed by #400

ehmicky:
Published in [`v4.0.0`](https://github.com/sindresorhus/execa/releases/tag/v4.0.0).

PR #400, opened 2019-12-18
Fix unref error

mifi:
See https://github.com/sindresorhus/execa/issues/399

### DIFF
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -22,6 +22,7 @@ const setKillTimeout = (kill, signal, options, killResult) => {
 		kill('SIGKILL');
 	}, timeout);
 
+	// istanbul ignore else
 	if (t.unref) {
 		t.unref();
 	}
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -21,7 +21,10 @@ const setKillTimeout = (kill, signal, options, killResult) => {
 	const t = setTimeout(() => {
 		kill('SIGKILL');
 	}, timeout);
-	if (t.unref) t.unref();
+
+	if (t.unref) {
+		t.unref();
+	}
 };
 
 const shouldForceKill = (signal, {forceKillAfterTimeout}, killResult) => {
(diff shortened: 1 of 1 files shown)


lib/kill.js
```diff
@@ -18,9 +18,13 @@ const setKillTimeout = (kill, signal, options, killResult) => {
 	}
 
 	const timeout = getForceKillAfterTimeout(options);
-	setTimeout(() => {
+	const t = setTimeout(() => {
 		kill('SIGKILL');
-	}, timeout).unref();
+	}, timeout);
+
+	if (t.unref) {
```
ehmicky:
```suggestion
// istanbul ignore else
if (t.unref) {
```

We don't run tests on Electron, so this branch cannot be covered by Ava/Istanbul.

ehmicky:
Waiting for @sindresorhus to review it as well.

sindresorhus:
Would be good to open an issue on Electron about this and add a code comment that links to that issue.

ehmicky:
@sindresorhus Would it be ready to merge otherwise?

sindresorhus:
Yes

mifi:
I found these:
- https://github.com/electron/electron/issues/21162
- https://github.com/electron/electron/issues/12462

lib/kill.js
```diff
@@ -18,9 +18,14 @@ const setKillTimeout = (kill, signal, options, killResult) => {
...
 
 	const timeout = getForceKillAfterTimeout(options);
-	setTimeout(() => {
+	const t = setTimeout(() => {
 		kill('SIGKILL');
-	}, timeout).unref();
+	}, timeout);
+
+	// istanbul ignore else
+	if (t.unref) {
```
sindresorhus: @ehmicky This still needs a code comment about why it's guarded. For example:

> Guarded because there's no `.unref` when `execa` is used in the renderer process in Electron.

ehmicky: Oops, I thought this was ok to merge, sorry about that.
I've added the comment in #401.

thislooksfun:
Is it possible to back port this fix to `3.x.x`? (ref: https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/647#issuecomment-583176847)

PR #405, opened 2019-12-23
Clean up outdated code in tests

fisker:


ehmicky:
Hi @fisker, thanks for cleaning this up.

This looks good to me.
@sindresorhus what do you think?

PR #419, opened 2020-04-12
Is it not possible to kill shelled process?

gaggle:
It seems to me that a process started with `shell:true` don't always get killed properly.

Here I've added tests that uses Linux's sleep, and killing the non-shell version works fine but the shelled version hangs the test-suite for the 200 seconds. The sleep sits in the process list, unkilled.

Is this expected behaviour? To my understanding a shelled process should receive signals just as well as any other process, but happy to learn otherwise.

### DIFF
--- a/test/kill.js
+++ b/test/kill.js
@@ -19,6 +19,27 @@ test('kill("SIGKILL") should terminate cleanly', async t => {
 	t.false(isRunning(subprocess.pid));
 });
 
+// `sleep` is a Linux command so these tests do not make sense on Windows.
+if (process.platform !== 'win32') {
+	test('`kill` should kill a process cleanly', async t => {
+		const subprocess = execa('sleep', ['200'], { stdio: ['ipc'] });
+
+		subprocess.kill('SIGTERM', { forceKillAfterTimeout: 2000 });
+
+		await t.throwsAsync(subprocess);
+		t.false(isRunning(subprocess.pid));
+	});
+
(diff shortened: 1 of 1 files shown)


test/kill.js
```diff
@@ -16,8 +16,30 @@ test('kill("SIGKILL") should terminate cleanly', async t => {
...
+	});
+
+	test('`kill` should kill a shell process cleanly', async t => {
+		const subprocess = execa('sleep', ['200'], { shell: true, stdio: ['ipc'] });
+
+		subprocess.kill('SIGTERM', { forceKillAfterTimeout: 2000 });
+
+		await t.throwsAsync(subprocess);
+		t.false(isRunning(subprocess.pid));
+	})
```
gaggle, line 40: This test hangs the test-suite for the duration of the sleep, waiting for the process to stop

ehmicky:
This is due to #96. See my [new comment](https://github.com/sindresorhus/execa/issues/96#issuecomment-612658580) on that issue.

Since this is a well-known issue, I suggest this PR to be closed if this is ok with you @gaggle?

gaggle:
> This is due to #96. See my [new comment](https://github.com/sindresorhus/execa/issues/96#issuecomment-612658580) on that issue.
> 
> Since this is a well-known issue, I suggest this PR to be closed if this is ok with you @gaggle?

Oh I didn't see that issue, thanks. Yes of course, this PR was just to illustrate my question.

PR #424, opened 2020-05-15
Duplex stream

### LINKED ISSUE #143: Consider implementing a duplex mode
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/execa/issues/143)

Basically, a way to make a predictable duplex stream from `stdin` and `stdout` (or `stderr`), so you can use a child process as a transform stream.  It'd emit an error if the process errors, and it'd kill the process if the stream is destroyed.

```js

myStream.pipe(execa.duplex('myFile', ['my', 'arguments'])).pipe(myOtherStream);
```

It's not terribly difficult to do something like this as `execa` is now

```js

execa.stdout('myFile', ['my', 'arguments'], {input: myStream}).pipe(myOtherStream);
```

but this can behave in unexpected ways.

```js
// ENOENT gets swallowed, stream completes without error.
execa.stdout('myFiel', ['my', 'arguments'], {input: myStream});

const processStream = execa.stdout('myFile', ['my', 'arguments'], {input: myStream});
// `stdout` is destroyed, but the process doesn't die.
processStream.destroy();
```

> There is a $70.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/execa/issues/143).

sloonz:
First shot at an implementation for fixes #143

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 const path = require('path');
 const childProcess = require('child_process');
 const crossSpawn = require('cross-spawn');
+const stream = require('stream');
 const stripFinalNewline = require('strip-final-newline');
 const npmRunPath = require('npm-run-path');
 const onetime = require('onetime');
@@ -255,4 +256,136 @@ module.exports.node = (scriptPath, args, options = {}) => {
 			shell: false
 		}
 	);
+};
+
+module.exports.duplexStream = (file, args, options) => {
+	const parsed = handleArguments(file, args, options);
+	const command = joinCommand(file, args);
+
+	let spawned;
+	try {
+		spawned = childProcess.spawn(parsed.file, parsed.args, parsed.options);
+	} catch (error) {
+		const fullError = makeError({
+			error,
+			stdout: undefined,
+			stderr: undefined,
+			all: undefined,
+			command,
+			parsed,
+			timedOut: false,
+			isCanceled: false,
+			killed: false
+		});
+
+		// Dummy duplex that immediately returns an error
+		return new stream.Duplex({
+			read() {
+				this.destroy(fullError);
+			},
+			write(chunk, encoding, callback) {
+				callback(fullError);
+			}
+		});
+	}
+
+	const spawnedPromise = getSpawnedPromise(spawned);
+	const timedPromise = setupTimeout(spawned, parsed.options, spawnedPromise);
+	const processDone = setExitHandler(spawned, parsed.options, timedPromise);
+
+	const context = {isCanceled: false};
+
+	spawned.kill = spawnedKill.bind(null, spawned.kill.bind(spawned));
+	spawned.cancel = spawnedCancel.bind(null, spawned, context);
+
+	const all = makeAllStream(spawned, parsed.options);
+	const {stdin} = spawned;
+	const output = all ? all : spawned.stdout;
+
+	const duplex = new stream.Duplex({
+		autoDestroy: false,
+		write(chunk, encoding, callback) {
+			stdin.write(chunk, encoding, callback);
+		},
+		final(callback) {
+			stdin.end(async error => {
+				// Wait for the reader to be closed before closing the finalizing the write end
+				//
+				// The order usually doesn't matter, except when only the writer end is given to a pipeline
+				// (for example pipeline(getArchive(), duplexStream("tar", ["-x]))) : then the stream
+				// may end up closed before the process finishes (because getArchive() writer end is done and
+				// duplexStream() reader end is done, the pipeline is done even though tar may still be
+				// writing stuff to the disk) whereas the caller expects the process to be one by the time
+				// the pipeline is done
+				await readerClosed;
+				callback(error);
+			});
+		},
+		read() {
+			output.resume();
+		},
+		destroy(error, callback) {
+			if (spawned.exitCode === null && spawned.signalCode === null) {
+				spawned.once('exit', () => {
+					callback(error);
+				});
+
+				spawned.once('error', () => {
+					callback(error);
+				});
+
+				spawned.cancel();
+			} else {
+				callback(error);
+			}
+		}
+	});
+
+	if (output !== null) {
+		output.on('data', chunk => {
+			if (!duplex.push(chunk)) {
+				output.pause();
+			}
+		});
+	}
+
+	const stdinClosePromise = new Promise(resolve => {
+		if (stdin === null) {
+			resolve();
+		} else {
+			stdin.once('close', resolve);
+		}
+	});
+
+	const outputClosePromise = new Promise(resolve => {
+		if (output === null) {
+			resolve();
+		} else {
+			output.once('close', resolve);
+		}
+	});
+
+	const readerClosed = (async () => {
+		const [{error, exitCode, signal, timedOut}] = await getSpawnedResult({}, {}, processDone);
+		await Promise.all([stdinClosePromise, outputClosePromise]);
+		if (error || exitCode !== 0 || signal !== null) {
+			duplex.destroy(makeError({
+				error,
+				exitCode,
+				signal,
+				stdout: undefined,
+				stderr: undefined,
+				all: undefined,
+				command,
+				parsed,
+				timedOut,
+				isCanceled: context.isCanceled,
+				killed: spawned.killed
+			}));
+		} else {
+			duplex.push(null);
+		}
+	})();
+
+	return duplex;
 };
--- a/test/duplex.js
+++ b/test/duplex.js
@@ -1,0 +1,92 @@
+import path from 'path';
+import test from 'ava';
+import execa from '..';
+import stream from 'stream';
+import {promisify} from 'util';
+
+const pipeline = promisify(stream.pipeline);
+
+process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
+
+function makeCollector() {
+	const chunks = [];
+	const writableStream = new stream.Writable({
+		write(chunk, encoding, callback) {
+			chunks.push(chunk);
+			callback();
+		},
+		final(callback) {
+			callback();
+		}
+	});
+	writableStream.collect = () => Buffer.concat(chunks).toString('utf-8');
+	return writableStream;
+}
+
+function makeEmptyReadableStream() {
+	return new stream.Readable({
+		read() {}
+	});
+}
+
+function createReadableStream(data) {
+	return new stream.Readable({
+		read() {
+			this.push(data);
+			this.push(null);
+		}
+	});
+}
+
+test('simple pipeline', async t => {
+	const duplex = execa.duplexStream('stdin');
+	const collector = makeCollector();
+	t.is(await pipeline(createReadableStream('hello, world'), duplex, collector), undefined);
+	t.is(collector.collect(), 'hello, world');
+});
+
+test('command failure should result in pipeline failure', async t => {
+	const duplex = execa.duplexStream('fail');
+	const error = await t.throwsAsync(pipeline(makeEmptyReadableStream(), duplex, makeCollector()));
+	t.is(error.exitCode, 2);
+});
+
+test('pipeline failure should kill the process', async t => {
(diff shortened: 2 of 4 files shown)


test/duplex.js
```diff
@@ -0,0 +1,144 @@
...
+	t.is(err.exitCode, 2);
+});
+
+test('pipeline failure should kill the process', async t => {
+	const duplex = execa.duplexStream('forever');
+	const failingStream = makeEmptyReadableStream();
+	failingStream.destroy(new Error('oops'));
+	const {message} = await t.throwsAsync(pipeline(failingStream, duplex, makeCollector()));
+	t.is(message, 'oops');
+	const err = await t.throwsAsync(duplex);
```
sindresorhus, line 53: `err` => `error`

sindresorhus:
Thanks for working on this 🙌

Make sure you add a *lot* of tests.

sloonz:
No need to implement `writev`, it's optional method that is here for optimization purposes if you can handle multiple chunks at once, which we can't since we just pass them to process stdin.

Do you have an idea on what other tests I could make ?

Unless someone can lend me a hand, it will take some times for me to track the test failures on windows ; I don't have access to a windows machine right now.

sloonz:
Okay, found why some tests where failing on Windows and not Linux, but that raises more questions @sindresorhus 

It’s on unit tests testing this code path : https://github.com/sloonz/execa/blob/28ac048159f6ea1fa2c27499451c014a74e80265/index.js#L268-L287

I tried to trigger an exception on line 266 by passing `null` to file. This works as intended on Linux ; however, on Windows, the exception gets triggered earlier (at line 261, `handleArgs`), and the exception directly unwind to the caller instead of being caught and handled.

Two questions : 

* How can I consistently trigger an exception on `childProcess.spawn` ?
* I have not tested, but my problem most likely is also here for the main `execa()` function, where `execa(null)` returns an error (which doubles as a rejected promise) on Linux whereas it just throws on Windows. Is this the intended behavoir ? Should I fill a bug for this ?

sloonz:
I think the feature is in a good spot for review/integration in the current iteration ; should I squash the commits ?

index.js
```diff
@@ -254,3 +255,135 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			isCanceled: false,
+			killed: false
+		};
+	});
+
+	Promise.all([
+		result.then(() => {}).catch(error => error),
+		new Promise((resolve, _) => stdin.on('close', resolve)),
+		new Promise((resolve, _) => output.on('close', resolve))
+	]).then(([error]) => {
```
sindresorhus, line 400: Use `await`.

index.js
```diff
@@ -254,3 +255,152 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			all: undefined,
+			failed: false,
+			timedOut: false,
+			isCanceled: false,
+			killed: false
+		};
+	});
+
+	Promise.all([
+		result.then(() => {}).catch(error => error),
```
sindresorhus, line 393: Do you really need `.then` here?

index.js
```diff
@@ -254,3 +255,152 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			duplex.push(null);
+		}
+	});
+	output.on('data', chunk => {
+		if (!duplex.push(chunk)) {
+			output.pause();
+		}
+	});
+
+	const result = processDone.then(({error, exitCode, signal, timedOut}) => {
```
sindresorhus, line 360: Use `await`.

test/duplex.js
```diff
@@ -0,0 +1,157 @@
...
+import crypto from 'crypto';
+import {promisify} from 'util';
+
+const pipeline = promisify(stream.pipeline);
+
+process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
+
+function makeCollector() {
+	const chunks = [];
+	const w = new stream.Writable({
```
sindresorhus, line 14: Use descriptive variable names.

test/duplex.js
```diff
@@ -0,0 +1,157 @@
...
+		write(chunk, encoding, cb) {
```
sindresorhus, line 15:
```suggestion
		write(chunk, encoding, callback) {
```

test/duplex.js
```diff
@@ -0,0 +1,157 @@
...
+	const duplex = execa.duplexStream('fail');
+	const error = await t.throwsAsync(pipeline(makeEmptyReadableStream(), duplex, makeCollector()));
+	t.is(error.exitCode, 2);
+});
+
+test('pipeline failure should kill the process', async t => {
+	const duplex = execa.duplexStream('forever');
+	const failingStream = makeEmptyReadableStream();
+	failingStream.destroy(new Error('oops'));
+	const {message} = await t.throwsAsync(pipeline(failingStream, duplex, makeCollector()));
```
sindresorhus, line 52:
```suggestion
	await t.throwsAsync(pipeline(failingStream, duplex, makeCollector()), {message: 'oops'});
```

sindresorhus:
> should I squash the commits ?

No need. We'll squas when merging.

sindresorhus:
I would like to see even more tests for this. It's a complicated feature (streams have a *lot* of edge-cases).

index.js
sloonz: Yes, otherwise non-error result will be resolved by the complete promise. A more explicit intent would be

`result.then(() => { return null; }).catch(error => error)`

But anyway this is changed by the use of await instead of then.

sloonz:
Sorry for the hiatus.

New iteration of the branch is there. 

No new test because my dumb brain can't generate any idea of what to test that isn't currently covered. Do you have any suggestion ?

sloonz:
Small bump ?

index.js
```diff
@@ -254,3 +255,166 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+		const promisesResult = (await Promise.all([
+			(async () => {
+				try {
+					await result;
+					return null;
+				} catch (error) {
+					return error;
+				}
+			})(),
+			new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
```
sindresorhus, line 407: Don't disable the rule. Just adhere to it.

index.js
```diff
@@ -254,3 +255,166 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			all: undefined,
+			failed: false,
+			timedOut: false,
+			isCanceled: false,
+			killed: false
+		};
+	})();
+
+	(async () => {
+		const promisesResult = (await Promise.all([
```
sindresorhus, line 398: Use destructuring.

index.js
```diff
@@ -254,3 +255,166 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			(async () => {
+				try {
+					await result;
+					return null;
```
sindresorhus, line 402:
```suggestion
					return;
```

use `undefined`

index.js
```diff
@@ -254,3 +255,166 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+				}
+			})(),
+			new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
+				stdin.on('close', resolve);
+			}),
+			new Promise((resolve, reject) => { // eslint-disable-line no-unused-vars
+				output.on('close', resolve);
+			})
+		]));
+		if (promisesResult[0] !== null) {
```
sindresorhus, line 414:
```suggestion

		if (promisesResult[0] !== null) {
```

index.js
```diff
@@ -254,3 +255,166 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+	const parsed = handleArgs(file, args, options);
+	const command = joinCommand(file, args);
+
+	let spawned;
+	try {
+		spawned = childProcess.spawn(parsed.file, parsed.args, parsed.options);
+	} catch (error) {
+		const errorPromise = Promise.reject(makeError({
+			error,
+			stdout: undefined,
```
sindresorhus, line 269: This should not be `undefined`.

test/duplex.js
```diff
@@ -0,0 +1,157 @@
...
+	const writableStream = new stream.Writable({
+		write(chunk, encoding, callback) {
+			chunks.push(chunk);
+			callback();
+		},
+		final(callback) {
+			callback();
+		}
+	});
+	writableStream.collect = () => Buffer.concat(chunks).toString('utf-8');
```
sindresorhus, line 23:
```suggestion
	writableStream.collect = () => Buffer.concat(chunks).toString('utf8');
```

test/duplex.js
```diff
@@ -0,0 +1,157 @@
...
+
+test('pipeline failure should kill the process', async t => {
+	const duplex = execa.duplexStream('forever');
+	const failingStream = makeEmptyReadableStream();
+	failingStream.destroy(new Error('oops'));
+	const {message} = await t.throwsAsync(pipeline(failingStream, duplex, makeCollector()));
+	t.is(message, 'oops');
+	const error = await t.throwsAsync(duplex);
+	t.is(error.isCanceled, true);
+	t.is(error.killed, true);
```
sindresorhus, line 56:
```suggestion
	t.true(error.killed);
```

sindresorhus:
Would be nice to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up some code between the duplex method and the others.

sindresorhus:
In general for PRs:
- Look over your own diff.
- Try to improve the implementation.
- Try to refactor and simplify.
- Improve docs.

sloonz:
I would like your opinion on something.

I planned to refactor in two parts, but now I wonder if the second part is needed.

In the previous implementation the result of `duplexStream` was both a duplex stream AND and `execa` object (spawned process  object + promise). Now I wonder if this is a good idea, or if the result of `duplexStream` should not be just a stream and don't return the spawned process or a promise at all

Pros: way simpler code and interface
Cons: can't wait for the process to be killed after a failure in the pipeline (a kill signal is sent if the pipeline fails, but with just the duplex stream you have no way of waiting for the process to actually have ended) ; can't get the exit code (or really any information) of the process after a failure in the pipeline. I think the biggest con is that I see no way of unit testing "pipeline failure should kill the process".

I tend to prefer the simpler solution, but at the end it's your call :)

sloonz:
Attempt n˚ 2

I simplified both the API (discarded the idea that the returned value is both a stream and a promise ; it’s a pure duplex now) and the implementation.

As a side-node, I’m running it in production for my backup system for a few months now (fortunately for testing, unfortunately for me including two backup recoveries).

ehmicky:
It appears that this PR might be stalled.

Also, it seems like using a `TransformStream` might now be more proper than a `Duplex`?

In terms of implementation, I am wondering whether instead of duplicating the code from the main `execa()` function, a smaller function could first call `execa()` then create a `TransformStream` wrapping the return value's `stdin`, `stdout` (or `all` if the `all` option is `true`). This would be easier to maintain. This is the approach that both `execa.node()` and `execa.$` are using.

Based on this, I am wondering whether this PR should be closed?
cc @sindresorhus

sindresorhus:
> Also, it seems like using a TransformStream might now be more proper than a Duplex?

How so?

ehmicky:
For similar reasons as `Buffer` vs `Uint8Array`: web streams can be used in both browsers and Node.js.
However, `Duplex` is fine too, considering it is easy to convert using `Stream.Duplex.toWeb(duplex)`.

Regardless, this PR is quite behind the current codebase. Also, it relies on re-implementing the main logic of `execa()` instead of wrapping it like `execa.node` and `execa.$`, which might come with some maintenance cost. What do you think?

sindresorhus:
> However, Duplex is fine too, considering it is easy to convert using Stream.Duplex.toWeb(duplex).

I do I agree it would be nicer to use modern APIs whenever we add new features. So yes, `TransformStream` (web streams) would be best.

> instead of wrapping it like execa.node and execa.$, which might come with some maintenance cost. What do you think?

Agreed. It would be better to wrap it like you said.

PR #427, opened 2020-05-22
Fix problems with third-party Promise

### LINKED ISSUE #426: 4.0.1 causes problems with Bluebird
I've gotten an error from Bluebird when using execa at version `4.0.1`:
```
TypeError: this._then is not a function
    at Promise.then (~/.config/yarn/global/node_modules/bluebird/js/release/promise.js:154:17)
    at ChildProcess.then (~/.config/yarn/global/node_modules/execa/lib/promise.js:13:25)
```

This works fine if I downgrade execa to `4.0.0`.

fisker:
Fixes #426

### DIFF
--- a/lib/promise.js
+++ b/lib/promise.js
@@ -1,8 +1,9 @@
 'use strict';
 
+const nativePromisePrototype = (async () => {})().constructor.prototype;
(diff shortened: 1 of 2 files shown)


ehmicky:
Looks good, thanks!
Waiting for review from @sindresorhus too.

PR #430, opened 2020-06-09
Fully automate dev setup with Gitpod

skyperx:
This commit implements a fully-automated development setup using Gitpod.io, an
online IDE for GitHub and GitLab that enables Dev-Environments-As-Code.
This makes it easy for anyone to get a ready-to-code workspace for any branch,
issue or pull request almost instantly with a single click.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,7 @@
+FROM gitpod/workspace-full
+
+# Install custom tools, runtimes, etc.
+# For example "bastet", a command-line tetris clone:
+# RUN brew install bastet
+#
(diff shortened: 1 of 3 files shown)


ehmicky:
IDE-specific tooling should not live in source code.

This also could be seen as a PR promoting a specific product rather than trying to improve this project.

@sindresorhus Should this be closed?

sindresorhus:
Thanks for the suggestion, but I'm not interested in adding support for random services.

All npm packages work the same regarding `npm install`. So instead of you going around adding this to every single npm package, it would be better for Gitpod to just support npm packages natively without any changes to the npm packages.

PR #431, opened 2020-06-09
Fix use of floating number for the `timeout` and `forceKillAfterTimeout` options

Kikobeats:
When the `timeout` or `forceKillAfterTimeout` a float number is causes an error that can be avoided

### DIFF
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -71,7 +71,7 @@ const setupTimeout = (spawned, {timeout, killSignal = 'SIGTERM'}, spawnedPromise
 		return spawnedPromise;
 	}
 
-	if (Number.isFinite(number) || timeout < 0) {
+	if (!Number.isFinite(number) || timeout < 0) {
 		throw new TypeError(`Expected the \`timeout\` option to be a non-negative integer, got \`${timeout}\` (${typeof timeout})`);
 	}
 
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -44,7 +44,7 @@ const getForceKillAfterTimeout = ({forceKillAfterTimeout = true}) => {
 		return DEFAULT_FORCE_KILL_TIMEOUT;
 	}
 
-	if (typeof forceKillAfterTimeout !== 'number' || forceKillAfterTimeout < 0) {
+	if (Number.isFinite(forceKillAfterTimeout) || forceKillAfterTimeout < 0) {
 		throw new TypeError(`Expected the \`forceKillAfterTimeout\` option to be a non-negative integer, got \`${forceKillAfterTimeout}\` (${typeof forceKillAfterTimeout})`);
 	}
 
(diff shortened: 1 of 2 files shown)


Kikobeats:
Ok I can see it's an intentional behavior

```
   64:   test('`forceKillAfterTimeout` should not be a float', t => {    
```

not sure why 🤔

sindresorhus:
I don't remember. Maybe try Git blame.

Kikobeats:
the code was refactored several times but `isInteger` was introduced here:
https://github.com/sindresorhus/execa/commit/0bd5596bfab81f67de44545adeb3d6c912a0db85#diff-168726dbe96b3ce427e7fedce31bb0bcR246

I couldn't determinate if number over float is necessary for a thing, maybe @ehmicky can know 🙂

ehmicky:
Yes, I think we should allow floats.
However we should still disalllow `NaN` and `Infinity`, so `Number.isFinite()` might be better than `typeof value === 'number'` I believe?

sindresorhus:
Yes, let’s do that.

lib/kill.js
```diff
@@ -44,7 +44,7 @@ const getForceKillAfterTimeout = ({forceKillAfterTimeout = true}) => {
 		return DEFAULT_FORCE_KILL_TIMEOUT;
 	}
 
-	if (!Number.isInteger(forceKillAfterTimeout) || forceKillAfterTimeout < 0) {
+	if (Number.isFinite(forceKillAfterTimeout) || forceKillAfterTimeout < 0) {
```
sindresorhus, line 47:
```suggestion
	if (!Number.isFinite(forceKillAfterTimeout) || forceKillAfterTimeout < 0) {
```

lib/kill.js
```diff
@@ -71,7 +71,7 @@ const setupTimeout = (spawned, {timeout, killSignal = 'SIGTERM'}, spawnedPromise
 		return spawnedPromise;
 	}
 
-	if (!Number.isInteger(timeout) || timeout < 0) {
+	if (Number.isFinite(number) || timeout < 0) {
```
ehmicky, line 74:
```suggestion
	if (!Number.isFinite(number) || timeout < 0) {
```

lib/kill.js
```diff
@@ -71,7 +71,7 @@ const setupTimeout = (spawned, {timeout, killSignal = 'SIGTERM'}, spawnedPromise
 		return spawnedPromise;
 	}
 
-	if (!Number.isInteger(timeout) || timeout < 0) {
+	if (!Number.isFinite(number) || timeout < 0) {
```
ehmicky, line 74:
```suggestion
	if (!Number.isFinite(timeout) || timeout < 0) {
```

ehmicky:
The CI test are failing due to the following test: 

https://github.com/sindresorhus/execa/blob/a827d82203a1440e585276bef5d399a5953801f1/test/kill.js#L64

This probably should be changed to check for `NaN` instead?

Kikobeats:
@ehmicky test updated; still, some tests are failing  but for a different reason no related with this PR

ehmicky:
Thanks @Kikobeats,

It looks like you accidentally also removed the test that checks that the argument is not negative. Could you please add it back? Thanks :)

Kikobeats:
@ehmicky ops, done!

PR #435, opened 2020-09-14
Remove `--inspect` & `--inspect-brk` from `execArgv`

wardpeet:
When using `execa.node` and the parent process has --inspect or --inspect-brk defined, it will run the child process with those options as well. It will throw an error that the inspector already is running on port 9229.

I've added a filter for those options on the default execArgv value. I tried to test it but I'm not sure this is the correct way to do so.

### DIFF
--- a/index.js
+++ b/index.js
@@ -234,20 +234,7 @@ module.exports.node = (scriptPath, args, options = {}) => {
 	}
 
 	const stdio = normalizeStdio.node(options);
-
-	// Filter out `--inspect` & `--inspect-brk` from default `execArgv`.
-	const defaultExecArgv = process.execArgv.filter(arg => {
-		if (
-			arg === '--inspect' ||
-			arg.startsWith('--inspect=') ||
-			arg === '--inspect-brk' ||
-			arg.startsWith('--inspect-brk=')
-		) {
-			return false;
-		}
-
-		return true;
-	});
+	const defaultExecArgv = process.execArgv.filter(arg => !arg.startsWith('--inspect'));
 
 	const {
 		nodePath = process.execPath,
--- a/index.js
+++ b/index.js
@@ -234,8 +234,24 @@ module.exports.node = (scriptPath, args, options = {}) => {
 	}
 
 	const stdio = normalizeStdio.node(options);
+	// Filter out --inspect & --inspect-brk from default execArgv
+	const defaultExecArgv = process.execArgv.filter(arg => {
+		if (
+			arg === '--inspect' ||
+			arg.startsWith('--inspect=') ||
+			arg === '--inspect-brk' ||
+			arg.startsWith('--inspect-brk=')
+		) {
+			return false;
+		}
 
-	const {nodePath = process.execPath, nodeOptions = process.execArgv} = options;
+		return true;
+	});
+
+	const {
+		nodePath = process.execPath,
+		nodeOptions = defaultExecArgv
+	} = options;
 
 	return execa(
 		nodePath,
--- a/index.js
+++ b/index.js
@@ -234,7 +234,8 @@ module.exports.node = (scriptPath, args, options = {}) => {
 	}
 
 	const stdio = normalizeStdio.node(options);
-	// Filter out --inspect & --inspect-brk from default execArgv
+
+	// Filter out `--inspect` & `--inspect-brk` from default `execArgv`.
 	const defaultExecArgv = process.execArgv.filter(arg => {
 		if (
 			arg === '--inspect' ||
--- a/test/node.js
+++ b/test/node.js
@@ -13,7 +13,7 @@ async function inspectMacro(t, input) {
 			reject: false
 		});
 
-		const {stdout, stderr} = await subprocess
+		const {stdout, stderr} = await subprocess;
 
 		t.is(stdout, 'foo');
 		t.is(stderr, '');
@@ -57,7 +57,7 @@ test('node pass on nodeOptions', async t => {
 test.serial(
 	'node removes --inspect from nodeOptions when defined by parent process',
 	inspectMacro,
-  `--inspect`
+	'--inspect'
 );
 
 test.serial(
--- a/test/node.js
+++ b/test/node.js
@@ -3,9 +3,15 @@ import test from 'ava';
 import pEvent from 'p-event';
 import execa from '..';
 
-process.env.PATH =
-	path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
+process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
 
+test.beforeEach(t => {
+	t.context.originalArgv = process.execArgv;
+});
+
+test.afterEach(t => {
+	process.execArgv = t.context.originalArgv;
+});
 test('node()', async t => {
 	const {exitCode} = await execa.node('test/fixtures/noop');
 	t.is(exitCode, 0);
@@ -20,14 +26,11 @@ test('node pipe stdout', async t => {
 });
 
 test('node correctly use nodePath', async t => {
-	const {stdout} = await execa.node(
-		process.platform === 'win32' ? 'hello.cmd' : 'hello.sh',
-		{
-			stdout: 'pipe',
-			nodePath: process.platform === 'win32' ? 'cmd.exe' : 'bash',
-			nodeOptions: process.platform === 'win32' ? ['/c'] : []
-		}
-	);
+	const {stdout} = await execa.node(process.platform === 'win32' ? 'hello.cmd' : 'hello.sh', {
+		stdout: 'pipe',
+		nodePath: process.platform === 'win32' ? 'cmd.exe' : 'bash',
+		nodeOptions: process.platform === 'win32' ? ['/c'] : []
+	});
 
 	t.is(stdout, 'Hello World');
 });
@@ -44,12 +47,10 @@ test('node pass on nodeOptions', async t => {
 test.serial(
 	'node removes --inspect from nodeOptions when defined by parent process',
 	async t => {
-		const originalArgv = process.execArgv;
 		process.execArgv = ['--inspect', '-e'];
 		const {stdout, stderr} = await execa.node('console.log("foo")', {
-			stdout: 'pipe'
+			reject: false
 		});
-		process.execArgv = originalArgv;
 
 		t.is(stdout, 'foo');
 		t.is(stderr, '');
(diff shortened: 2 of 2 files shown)


test/node.js
```diff
@@ -3,7 +3,8 @@ import test from 'ava';
 import pEvent from 'p-event';
 import execa from '..';
 
-process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
+process.env.PATH =
+	path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
```
sindresorhus, line 7: Don't do unrelated changes.

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
...
+	async t => {
+		const {stdout, stderr} = await execa.node('console.log("foo")', {
+			stdout: 'pipe',
+			nodeOptions: ['--inspect', '-e']
+		});
+
+		t.is(stdout, 'foo');
+		t.regex(
+			stderr,
+			/^Debugger listening on ws:\/\/127.0.0.1:9229\/.*[\r\n]+For help, see: https:\/\/nodejs.org\/en\/docs\/inspector$/
```
sindresorhus, line 127:
```suggestion
			/Debugger listening/
```

I don't want to have to change it if it slightly changes on the Node.js side.

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
...
+		t.is(stderr, '');
+	}
+);
+
+test.serial(
+	'node removes --inspect-brk=9222 from nodeOptions when defined by parent process',
+	async t => {
+		const originalArgv = process.execArgv;
+		process.execArgv = ['--inspect-brk=9222', '-e'];
+		const childProc = execa.node('console.log("foo")', {
```
sindresorhus, line 100:
```suggestion
		const subprocess = execa.node('console.log("foo")', {
```

index.js
```diff
@@ -235,7 +235,24 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+			return false;
+		}
+
+		return true;
+	});
+
+	const {
+		nodePath = process.execPath,
+		nodeOptions = defaultExecArgv
+	} = options;
```
sindresorhus, line 255: Any thoughts on doing it no matter what? I don't really see any good reason users would want to override it.

sindresorhus (changes requested):

index.js
```diff
@@ -235,7 +235,24 @@ module.exports.node = (scriptPath, args, options = {}) => {
 
 	const stdio = normalizeStdio.node(options);
 
-	const {nodePath = process.execPath, nodeOptions = process.execArgv} = options;
+	// Filter out `--inspect` & `--inspect-brk` from default `execArgv`.
+	const defaultExecArgv = process.execArgv.filter(arg => {
```
ehmicky, line 239: The following might be simpler:

```js
const defaultExecArgv = process.execArgv.filter(arg => arg !== '--inspect' && !arg.startsWith('--inspect=') && arg !== '--inspect-brk' && !arg.startsWith('--inspect-brk='))
```

index.js
```diff
@@ -235,7 +235,24 @@ module.exports.node = (scriptPath, args, options = {}) => {
...
+		if (
+			arg === '--inspect' ||
```
ehmicky, line 241: I think using only `arg.startsWith('--inspect')` would work here since no `process.execArgv` only targets Node CLI flags, and only `--inspect` and `--inspect-brk` start with `--inspect`. If new `--inspect-*` are added in the future, we most likely will want to exclude them as well.

index.js
```diff
@@ -235,7 +235,24 @@ module.exports.node = (scriptPath, args, options = {}) => {
 
 	const stdio = normalizeStdio.node(options);
 
-	const {nodePath = process.execPath, nodeOptions = process.execArgv} = options;
+	// Filter out `--inspect` & `--inspect-brk` from default `execArgv`.
```
ehmicky, line 238: I think this comment might not be necessary since the code is quite clear regardless.

test/node.js
```diff
@@ -19,11 +20,14 @@ test('node pipe stdout', async t => {
...
-		nodeOptions: process.platform === 'win32' ? ['/c'] : []
-	});
+	const {stdout} = await execa.node(
+		process.platform === 'win32' ? 'hello.cmd' : 'hello.sh',
+		{
+			stdout: 'pipe',
+			nodePath: process.platform === 'win32' ? 'cmd.exe' : 'bash',
+			nodeOptions: process.platform === 'win32' ? ['/c'] : []
+		}
+	);
```
ehmicky, line 30: Same comment as https://github.com/sindresorhus/execa/pull/435/files#r496222326

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
 	t.is(stdout, 'foo');
 });
 
+test.serial(
```
ehmicky, line 44: Those tests do not need to be `test.serial()`, unless I am mistaken?

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
...
+	'node removes --inspect from nodeOptions when defined by parent process',
+	async t => {
+		const originalArgv = process.execArgv;
+		process.execArgv = ['--inspect', '-e'];
+		const {stdout, stderr} = await execa.node('console.log("foo")', {
+			stdout: 'pipe'
+		});
+		process.execArgv = originalArgv;
```
ehmicky, line 52: This statement should be inside `try`/`finally` in case `execa.node()` throws (due to a bug for example) so it does not impact the other tests.

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
 	t.is(stdout, 'foo');
 });
 
+test.serial(
+	'node removes --inspect from nodeOptions when defined by parent process',
+	async t => {
+		const originalArgv = process.execArgv;
+		process.execArgv = ['--inspect', '-e'];
+		const {stdout, stderr} = await execa.node('console.log("foo")', {
+			stdout: 'pipe'
```
ehmicky, line 50: `stdout` defaults to `pipe`, so does not need to be defined here (and in the following tests).

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
...
+
+test.serial(
+	'node removes --inspect-brk from nodeOptions when defined by parent process',
+	async t => {
+		const originalArgv = process.execArgv;
+		process.execArgv = ['--inspect-brk', '-e'];
+		const childProc = execa.node('console.log("foo")', {
+			stdout: 'pipe'
+		});
+		process.execArgv = originalArgv;
```
ehmicky, line 82: The few lines above are repeated across tests: would it be possible to create an helper function instead?

test/node.js
```diff
@@ -37,6 +41,94 @@ test('node pass on nodeOptions', async t => {
...
+
+		setTimeout(() => {
```
ehmicky, line 84: `async`/`await` could be used here:

```js
const { setTimeout } = require('util');
const pSetTimeout = promisify(setTimeout);
```

Then:

```js
const subprocess = execa.node(..., { reject: false });
```

Then:

```js
await pSetTimeout(1000);
subprocess.kill();
const {stdout, stderr} = await subprocess
```

test/node.js
ehmicky: This would also allow using `string.includes()` directly, which might be simpler:

```js
t.true(stderr.includes('Debugger listening'));
```

test/node.js
wardpeet: it's my vscode :'( sorry i'll revert

index.js
wardpeet: I find this harder to read but happy to update it.

index.js
wardpeet: Sometimes you want to set inspect on a subprocess when your main process is just a small wrapper. We could also tell people to move towards https://nodejs.org/api/inspector.html instead for this functionality.

I'm okay either way, just remember that changing this would be a breaking change.

ehmicky: > Sometimes you want to set inspect on a subprocess when your main process is just a small wrapper. 

Yes, it looks to me that could be the case, and this change would then be a breaking change if any users is doing this. Based on this, I personally think only doing this as a default value (what the PR currently does) might be better?

test/node.js
```diff
@@ -37,6 +44,75 @@ test('node pass on nodeOptions', async t => {
...
+
+test.serial(
+	'node removes --inspect-brk=9222 from nodeOptions when defined by parent process',
+	async t => {
+		process.execArgv = ['--inspect-brk=9222', '-e'];
+		const subprocess = execa.node('console.log("foo")', {
+			reject: false
+		});
+
+		const {stdout, stderr} = await subprocess.catch(error => error);
```
ehmicky, line 96: Is `.catch()` needed considering the `reject: false` option is used?

test/node.js
```diff
@@ -37,6 +44,75 @@ test('node pass on nodeOptions', async t => {
...
+	async t => {
+		process.execArgv = ['--inspect', '-e'];
+		const {stdout, stderr} = await execa.node('console.log("foo")', {
+			reject: false
+		});
+
+		t.is(stdout, 'foo');
+		t.is(stderr, '');
+	}
+);
```
ehmicky, line 58: The tests (except the last one) are almost identical except for the value of `process.execArgv`.
Using [test macros](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#reusing-test-logic-through-macros) might reduce code size here.

test/node.js
```diff
@@ -5,6 +5,13 @@ import execa from '..';
 
 process.env.PATH = path.join(__dirname, 'fixtures') + path.delimiter + process.env.PATH;
 
+test.beforeEach(t => {
+	t.context.originalArgv = process.execArgv;
+});
+
+test.afterEach(t => {
+	process.execArgv = t.context.originalArgv;
+});
```
ehmicky, line 14: If using test macros (see comment below), a `try`/`finally` block inside the tests could be used instead of `beforeEach`/`afterEach` hooks.

ehmicky (approved):
This looks good to me :+1: 
What do you think @sindresorhus ?

wardpeet:
Thanks! I wasn't familiar with ava so sorry for taking so long 👍

sindresorhus:
Nice work, @wardpeet

PR #447, opened 2020-11-25
Remove faulty emulated ENOENT error on Windows

### LINKED ISSUE #446: Incorrect ENOENT error on Windows with shell=true when child exits with code 1
## Summary

On Windows, when the execa child process exits with exit code 1 and the `shell` option is set to `true`, the error thrown is always `ENOENT` even though the child process was found and executed correctly (albeit with a non-zero exit code).

Using this code as an example:

```js
execa('node -e "process.exit(1)"', { shell: true })
    .catch(e => console.log(`Exit 1: ${e}\n`));
```

## Expected behavior

The output should be:
```
Exit 1: Error: Command failed with exit code 1: node -e "process.exit(1)"
```

## Actual behavior

The current output is:
```
Exit 1: Error: Command failed with ENOENT: node -e "process.exit(1)"
spawn node -e "process.exit(1)" ENOENT
```

## Additional information

The issue appears to be related to `verifyENOENT` from the `cross-spawn` library and its incorrect detection of whether a valid file was found to execute.

The following code shows that the issue only occurs when the child's exit code is 1, but not for exit code 2:

```js

execa('node -e "process.exit(2)"', { shell: true })
    .catch(e => console.log(`Exit 2: ${e}\n`));

execa('node -e "process.exit(1)"', { shell: true })
    .catch(e => console.log(`Exit 1: ${e}\n`));
```

As an additional test case, I'd recommend the following code because `exit` is a shell command, not an executable in the path. This code should **not** throw an `ENOENT` error.

```js

execa('exit 1', { shell: true })
    .catch(e => console.log(`Exit 1: ${e}\n`));
```

mterrel:
## Old behavior

Prior to this PR, `execa` on Windows would emulate behavior seen on POSIX platforms where attempting to run a non-existent command would throw an error that had `code` set to `ENOENT`. However, this emulation worked incorrectly for many common cases (fixes #446 and moxystudio/node-cross-spawn#104), especially when the `shell` option is set to a truthy value.

## New behavior

There is no obvious way to fix the emulation of `ENOENT` on Windows in a reasonable way, so this PR removes it. That means that `execa` will no longer throw `ENOENT` on Windows, but still will on POSIX platforms.
After this PR is merged, when `execa` is given a non-existent command to run on Windows, `execa` will have the default Node.js `spawn` behavior of throwing an error that has `exitCode` set to `1` and any error message given by Windows will appear in the `stderr` output/stream.

## Detecting "command not found" on Windows

On English versions of Windows, the error message given by the OS looks similar to this:

```
'somebadcommand' is not recognized as an internal or external command,
operable program or batch file.
```

However, this message will likely be translated on non-English versions of Windows.

## Additional info

Prior to this PR, there were two failing test cases on Windows. Removing the faulty `ENOENT` emulation actually fixed those two test cases, but caused two new ones to break. I fixed both by inducing a different error.

### DIFF
--- a/test/error.js
+++ b/test/error.js
@@ -73,8 +73,8 @@ test('error.shortMessage does not contain stdout/stderr', async t => {
 });
 
 test('Original error.message is kept', async t => {
-	const {originalMessage} = await t.throwsAsync(execa('wrong command'));
-	t.is(originalMessage, 'spawn wrong command ENOENT');
+	const {originalMessage} = await t.throwsAsync(execa('noop', {cwd: 1}));
+	t.true(originalMessage.startsWith('The "options.cwd" property must be of type string. Received type number'));
 });
 
 test('failed is false on success', async t => {
@@ -205,6 +205,6 @@ test('error.code is undefined on success', async t => {
 });
 
 test('error.code is defined on failure if applicable', async t => {
-	const {code} = await t.throwsAsync(execa('invalid'));
-	t.is(code, 'ENOENT');
+	const {code} = await t.throwsAsync(execa('noop', {cwd: 1}));
+	t.is(code, 'ERR_INVALID_ARG_TYPE');
 });
--- a/index.js
+++ b/index.js
@@ -147,8 +147,6 @@ const execa = (file, args, options) => {
 
 	const handlePromiseOnce = onetime(handlePromise);
 
-	crossSpawn._enoent.hookChildProcess(spawned, parsed.parsed);
-
 	handleInput(spawned, parsed.options.input);
 
 	spawned.all = makeAllStream(spawned, parsed.options);


test/error.js
```diff
@@ -73,8 +73,8 @@ test('error.shortMessage does not contain stdout/stderr', async t => {
 });
 
 test('Original error.message is kept', async t => {
-	const {originalMessage} = await t.throwsAsync(execa('wrong command'));
-	t.is(originalMessage, 'spawn wrong command ENOENT');
+	const {originalMessage} = await t.throwsAsync(execa('noop', {cwd: 1}));
+	t.regex(originalMessage, /The "options.cwd" property must be of type string. Received type number/);
```
ehmicky, line 77: This assertion might not need a regular expression.

```suggestion
	t.true(originalMessage.includes('The "options.cwd" property must be of type string');
```

test/error.js
```diff
@@ -204,7 +204,10 @@ test('error.code is undefined on success', async t => {
...
 
-test('error.code is defined on failure if applicable', async t => {
-	const {code} = await t.throwsAsync(execa('invalid'));
-	t.is(code, 'ENOENT');
-});
+// Windows doesn't throw ENOENT on invalid command
+if (process.platform !== 'win32') {
+	test('error.code is defined on failure if applicable', async t => {
+		const {code} = await t.throwsAsync(execa('invalid'));
+		t.is(code, 'ENOENT');
```
ehmicky, line 211: There might be ways to fix this test in a way that works on Windows as well. 
We need to find an `error.code` that is cross-platform.

The following might work:

```js
test('error.code is defined on failure if applicable', async t => {
	const {code} = await t.throwsAsync(execa(''));
	t.is(code, 'ERR_INVALID_ARG_VALUE');
});
```

ehmicky (changes requested):
Thanks @mterrel!

When it comes to detecting `ENOENT` by looking at the error message, I do agree that it might not be right to make that logic work only for users with English locales. Not only wouldn't it be fair to non-English speakers, but it might confuse users would get very different behaviors between machines and might not realize this is due to the locale.

test/error.js
mterrel: Changed as suggested

test/error.js
mterrel: Unfortunately, using an empty command on Windows doesn't generate `ERR_INVALID_ARG_VALUE`. There aren't many errors thrown consistently across both POSIX & Windows and also across multiple Node versions. The one I found that seems to work is giving an incorrect type for `cwd`. So I used that again here in the updated version.

ehmicky: That works :+1:

ehmicky (approved):
This is looking good to me. Thanks @mterrel!

PR #463, opened 2021-05-30
Fix `timeout` option validation

ehmicky:
Before Node `15.13.0` ([link to PR](https://github.com/nodejs/node/pull/37256)), the `timeout` was only available in `execFile()` and `exec()` but not in `spawn()`, which is used by Execa.

Execa implemented its own `timeout` option to fix this. That option probably won't be needed after support for Node `15.13.0` is dropped.

In the meantime, the `timeout` option added to `child_process.spawn()` introduces the following inconsistent behavior in Execa depending on the Node.js version:
  - In Node `<15.13.0`, using an invalid `timeout` (such as a negative integer) will throw an exception, using Execa's validation logic
  - In Node `>=15.13.0`, it will return a rejected promise instead, using `child_process.spawn()` validation logic

This PR fixes this inconsistency by moving Execa's validation logic earlier, before `child_process.spawn()` is called.

CI tests are failing due to #462.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ const npmRunPath = require('npm-run-path');
 const onetime = require('onetime');
 const makeError = require('./lib/error');
 const normalizeStdio = require('./lib/stdio');
-const {spawnedKill, spawnedCancel, setupTimeout, setExitHandler} = require('./lib/kill');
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you fix the conflict?

ehmicky:
Fixed :+1: 

Some CI tests are failing due to #464 (unrelated to this PR).

PR #466, opened 2021-06-01
Add `.escapedCommand` property

### LINKED ISSUE #455: Print the executed command?
I can pass command to execa via `execa.command('echo unicorns')` but it would also be great if execa could **give** me a string version of a command when I use the main syntax. Like this (pseudocode):

```js

(async () => {
  const { command } = await execa('echo', ['unicorns']);
  console.log(command);
  //=> 'echo unicorns'
})();
```

More realistic example where I'd use this:

```js

(async () => {
  const { command } = await execa("git", [
    "log",
    "-n",
    "1",
    "--pretty=format:%h",
    "--",
    ...paths,
  ]);
  console.log(command);
  //=> 'git log -n 1 --pretty=format:%h -- a.txt b.txt c.txt'
})();
```

The problem is probably shell-specific escaping, and the Git example above should be more safely something like

```
//=> "git" "log" "-n" "1" "--pretty=format:%h" "--" "a.txt" "b.txt" "c.txt"
```

I realize that there are many different shells, from Bash to Zsh to PowerShell to cmd.exe, and that it might be difficult to create a universal command. But it's a feature I was thinking about several times already so at least I wanted to write it up to see if it's a total nonsense or not.

ehmicky:
Closes #455.

This adds a `debugString` property to the return value and error instances, as described in the above issue.

As I implemented this, I realized that this property is very similar to the existing `command` property. What are your thoughts on either of the following options?
  1. Rename `debugString` to `escapedCommand` or `quotedCommand` to reflect this.
  2. Add escaping to the `command` property instead. This would be a breaking change.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -259,7 +259,7 @@ declare namespace execa {
 		/**
 		Same as `command` but quoted so it can be run in most shells.
 		*/
-		debugString: string;
+		escapedCommand: string;
 
 		/**
 		The numeric exit code of the process that was run.
--- a/index.d.ts
+++ b/index.d.ts
@@ -261,7 +261,7 @@ declare namespace execa {
 		/**
 		Same as `command` but escaped.
 
-		This is meant to be copy/pasted in a shell, for debugging purpose.
+		This is meant to be copy and pasted into a shell, for debugging purposes.
 		Since the escaping is fairly basic, this should be passed to neither `execa()` nor `execa.command()`
 		*/
 		escapedCommand: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -254,7 +254,7 @@ declare namespace execa {
 		/**
 		The file and arguments that were run, for logging purposes.
 
-		This is not escaped and should be passed to neither `execa()` nor `execa.command()`
+		This is not escaped and should not be executed directly as a process, including using `execa()` or `execa.command()`.
 		*/
 		command: string;
 
@@ -262,7 +262,7 @@ declare namespace execa {
 		Same as `command` but escaped.
 
 		This is meant to be copy and pasted into a shell, for debugging purposes.
-		Since the escaping is fairly basic, this should be passed to neither `execa()` nor `execa.command()`
+		Since the escaping is fairly basic, this should not be executed directly as a process, including using `execa()` or `execa.command()`.
 		*/
 		escapedCommand: string;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -252,7 +252,7 @@ declare namespace execa {
 
 	interface ExecaReturnBase<StdoutStderrType> {
 		/**
-		The file and arguments that were run, for logging purpose.
+		The file and arguments that were run, for logging purposes.
 
 		This is not escaped and should be passed to neither `execa()` nor `execa.command()`
 		*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -257,6 +257,11 @@ declare namespace execa {
 		command: string;
 
 		/**
+		Same as `command` but quoted so it can be run in most shells.
+		*/
+		debugString: string;
+
+		/**
 		The numeric exit code of the process that was run.
 		*/
 		exitCode: number;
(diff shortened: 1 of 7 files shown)


borekb:
Great to see this happening! I'm a fan of being able to pass the `debugString` (possibly under a different name) to `execa.command` and see it work. This was the example in https://github.com/sindresorhus/execa/issues/455#issuecomment-819642380:

```js

(async () => {
  const { command } = await execa('echo', ['unicorns']);
  await execa.command(command); // works
})();
```

`debugString` is a good name but I like `command` as well, and it would be an intuitive counterpart to `execa.command`.

ehmicky:
> Great to see this happening! I'm a fan of being able to pass the debugString (possibly under a different name) to execa.command and see it work. This was the example in #455 (comment):

The discussion in #455 resolved towards a different purpose though: returning a string for debugging purpose (per @sindresorhus comment [here](https://github.com/sindresorhus/execa/issues/455#issuecomment-821805376)), not to re-use programmatically. Re-using the same command + arguments programmatically can already be achieved by using variables to store them, or by refactoring similar code into single functions ([see my previous comment](https://github.com/sindresorhus/execa/issues/455#issuecomment-819658678)).

borekb:
My comment was specifically about your no. 2 suggestion above, i.e., changing `execa.command` so that it accepts the "debug string". If that was the case, I think that the "debug string" could be called "command".

But I understand it's slightly controversial, and all your suggestions like `debugString`, `quotedCommand` or `escapedCommand` sound good to me.

ehmicky:
> My comment was specifically about your no. 2 suggestion above, i.e., changing execa.command so that it accepts the "debug string". If that was the case, I think that the "debug string" could be called "command".

My bad, I am realizing my comment was not clear. By `command`, I meant the property that is returned by `execa()`, not the `execa.command()` method itself. So by `Add escaping to the command property instead.`, I meant merging the `debugString` and `command` properties into a single one named `command`. However, that idea might not be a good one since it might hint that the property can be passed to `execa.command()` although this is not the case.

borekb:
Got it, thanks for the explanation, it was at least partially a misunderstanding on my end 😄.

sindresorhus:
I would go with renaming it to `escapedCommand`. I would also update the docs for both `command` and `escapedCommand` to make it clear that it should not be passed to `execa.command`, and that `.command` is for logging and `.escapedCommand` is for debugging in an actual shell.

ehmicky:
Fixed :+1: 
Please let me know what you think.

index.d.ts
```diff
@@ -252,10 +252,20 @@ declare namespace execa {
...
+
+		This is not escaped and should be passed to neither `execa()` nor `execa.command()`
 		*/
 		command: string;
 
+		/**
+		Same as `command` but escaped.
+
+		This is meant to be copy/pasted in a shell, for debugging purpose.
+		Since the escaping is fairly basic, this should be passed to neither `execa()` nor `execa.command()`
```
sindresorhus, line 265: To prevent any vulnerability report coming from this, I think we should say that it should not be used with any child process type method. Same with `.command`.

index.d.ts
```diff
@@ -252,10 +252,20 @@ declare namespace execa {
 
 	interface ExecaReturnBase<StdoutStderrType> {
 		/**
-		The file and arguments that were run.
+		The file and arguments that were run, for logging purpose.
```
sindresorhus, line 255:
```suggestion
		The file and arguments that were run, for logging purposes.
```

index.d.ts
```diff
@@ -252,10 +252,20 @@ declare namespace execa {
...
+
+		This is not escaped and should be passed to neither `execa()` nor `execa.command()`
 		*/
 		command: string;
 
+		/**
+		Same as `command` but escaped.
+
+		This is meant to be copy/pasted in a shell, for debugging purpose.
```
sindresorhus, line 264:
```suggestion
		This is meant to be copy and pasted into a shell, for debugging purposes.
```

sindresorhus (changes requested):

index.d.ts
ehmicky: Fixed :+1:

Issue #469, opened 2021-06-10
Subprocess should emit error if executable path doesn't exist in Windows


lynxtaa:
Returned from `execa` subprocess should emit 'error' event if executable path doesn't exist in Windows. Just like `spawn` and `execFile` in `child_process`

Consider this code:

```js
const { spawn } = require('child_process')
const execa = require('execa')

const p1 = spawn('wrong-path')
p1.on('error', err => {
	console.log(err) // logs ENONENT error
})

const p2 = execa('wrong-path')
p2.on('error', err => {
	console.log(err) // never called in Windows, but called in Linux
})
```

ehmicky:
Hi @lynxtaa,

Thanks for reaching out.
Which version of `execa` are you using? I cannot reproduce this behavior.

This is an example with running your example on my machine on Windows. Note: I replaced the second `wrong-path` by `wrong-path-2` to differentiate them.

![windows](https://user-images.githubusercontent.com/8136211/121517844-18c79880-c9f0-11eb-951a-37f0637e38af.png)

A potential issue you might be experiencing is if you forgot to add a `.catch()` handler to the return value of `execa`. The return value is both  a child process and a promise. On errors, that promise is rejected. Unhandled promise rejections make the process exit by default on Node.js, which might prevent further code (like the `error` event handler) from running.

Aside from this, `execa` actually calls `spawn()` under the hood and just return the child process as is, so there should not be any difference of behavior.

lynxtaa:
@ehmicky I'm running Node@16.3.0 and execa@5.1.1.

> A potential issue you might be experiencing is if you forgot to add a `.catch()` handler to the return value of `execa`. The return value is both a child process and a promise. On errors, that promise is rejected. Unhandled promise rejections make the process exit by default on Node.js, which might prevent further code (like the `error` event handler) from running.

I've tried running `node  --unhandled-rejections=warn t.js` -- same issue

ehmicky:
Is `t.js` contained anymore more than the sample above?
Considering I get a different behavior when running it myself on Windows (in both `cmd.exe` and Bash), I am curious which differences of environment might explain that we get different results? Is there any additional reason you think might explain those differences?

lynxtaa:
> Does t.js contain any additional code beside the sample above?

No, that's all. I've reproduced this problem on three different Windows 10 machines running Node 16.3.0 and Node 14.17.0. In WSL in works okay. To be sure it has time to handle an error, I'm running this code:
```js
[...]

PR #470, opened 2021-06-10
Return error ENOENT when program not found on Windows

lynxtaa:
Fixes https://github.com/sindresorhus/execa/issues/469

See https://github.com/moxystudio/node-cross-spawn/issues/16

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,6 +81,9 @@ const execa = (file, args, options) => {
 	let spawned;
 	try {
 		spawned = childProcess.spawn(parsed.file, parsed.args, parsed.options);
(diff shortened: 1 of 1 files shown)


ehmicky (changes requested):
Thanks a lot for trying to fix this issue @lynxtaa!

However, this PR would revert https://github.com/sindresorhus/execa/pull/447, which was fixing the problem described in https://github.com/sindresorhus/execa/issues/446.

Several users (including me) have actually suggested removing the `ENOENT` logic from `node-cross-spawn` because it is unreliable in too many cases, so using it in `execa` is probably not a good idea unfortunately.

PR #478, opened 2021-11-15
Require Node.js 12.20 and move to ESM

NMinhNguyen:
Apologies for a huge PR but it's best to review ignoring whitespace: https://github.com/sindresorhus/execa/pull/478/files?w=1

I tried to follow the following resources:

- https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
- https://github.com/sindresorhus/typescript-definition-style-guide
- https://github.com/sindresorhus/is-stream/commit/5831295859122e5a246f31d6b92c8d5e8b551d5e?w=1.

Opening this as a draft PR since I'm not 100% sure of some of the changes:

- ~What should the named exports be for `command`, `commandSync` and `node`? Should they be `execaCommand`, `execaCommandSync` and `execaNode`?~ Prefixed the named exports with `execa`.
- Had to switch from `nyc` to `c8` to work around https://github.com/istanbuljs/nyc/issues/1287
- ~Removed Node.js 12 from the test matrix as I've changed `require` to `import()` + top-level await (TLA) in [`main...NMinhNguyen:esm?expand=1&w=1`#diff-8f4f0a53c9](https://github.com/sindresorhus/execa/compare/main...NMinhNguyen:esm?expand=1&w=1#diff-8f4f0a53c9879299286013b77c69f2d05ebe13ac2580acf641f774a4cf94937a), and TLA isn't supported in Node.js 12.~ Changed the test in [`aefa85c` (#478)](https://github.com/sindresorhus/execa/pull/478/commits/aefa85ce20840e6d6e26cd50f0700070dab22826) to ensure tests can still be run in Node.js 12.
- Moved the fixtures to ESM which required adding a `.js` extension. Alternatively, I could've added a `package.json` with `"type": "commonjs"`.

## TODOs

- [x] Ensure all fixtures are referenced using `.js` extension


sindresorhus:
> What should the named exports be for command, commandSync and node? Should they be execaCommand, execaCommandSync and execaNode?

Yes, that's what I would go with, but let's wait for @ehmicky's opinion.

test/node.js
```diff
@@ -57,42 +59,44 @@ test('node pass on nodeOptions', async t => {
...
 		t.is(stdout, 'foo');
 		t.true(stderr.includes('Debugger listening'));
-	}
+	},
 );
 
 test('node\'s forked script has a communication channel', async t => {
-	const subprocess = execa.node('test/fixtures/send');
+	const subprocess = execa.node('test/fixtures/send.js');
+	await pEvent(subprocess, 'message');
```
NMinhNguyen, line 98: Without waiting for a [message](https://github.com/sindresorhus/execa/pull/478/files#diff-26a0afea745973df4bc95c26bdead61741221bdd9952dbf06540d5f69178ceec) from `send.js`, it seems like `subprocess.send('ping')` would happen before it gets a chance to install a listener. Possibly because ES modules are async?

index.test-d.ts
```diff
@@ -1,13 +1,22 @@
+import {Buffer} from 'node:buffer';
+// For some reason a default import of `process` causes
+// `process.stdin`, `process.stderr`, and `process.stdout`
+// to get treated as `any` by `@typescript-eslint/no-unsafe-assignment`.
+import * as process from 'node:process';
```
NMinhNguyen, line 5: Not sure how to best debug this, but I couldn't use

```js
```

as documented at https://nodejs.org/api/process.html#process without running into TypeScript issues.

test/fixtures/detach.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+import process from 'node:process';
+import {execa} from '../../index.js';
+
+const subprocess = execa('node', ['./test/fixtures/forever.js'], {detached: true});
+console.log(subprocess.pid);
+process.exit(0);
```
NMinhNguyen, line 7: Not sure if it was worth adding a comment in the code but there might be a bug in Node.js (https://github.com/nodejs/node/issues/40808) preventing us from using `process.exit()` (without 0) in ESM.

index.d.ts
```diff
@@ -1,44 +1,43 @@
...
 		Kill the spawned process when the parent process exits unless either:
 			- the spawned process is [`detached`](https://nodejs.org/api/child_process.html#child_process_options_detached)
 			- the parent process is terminated abruptly, for example, with `SIGKILL` as opposed to `SIGTERM` or a normal exit
 
 		@default true
 		*/
-		readonly cleanup?: boolean;
+	readonly cleanup?: boolean;
 
-		/**
```
ehmicky, line 25: Would it be possible to separate all whitespace changes to a separate PR?
This will make reviewing this PR easier. Thanks!

index.test-d.ts
ehmicky: Thanks @NMinhNguyen. I think the code comment will do then :+1: 
Especially considering this is only a test file, and only for types.

package.json
```diff
@@ -41,26 +43,26 @@
...
 		"signal-exit": "^3.0.3",
-		"strip-final-newline": "^2.0.0"
+		"strip-final-newline": "^3.0.0"
 	},
 	"devDependencies": {
-		"@types/node": "^14.14.10",
-		"ava": "^2.4.0",
+		"@types/node": "^16.11.7",
+		"ava": "^3.15.0",
+		"c8": "^7.10.0",
```
ehmicky, line 57: Thanks for fixing `nyc` -> `c8` too!

package.json
```diff
@@ -41,26 +43,26 @@
...
+		"onetime": "^6.0.0",
 		"signal-exit": "^3.0.3",
-		"strip-final-newline": "^2.0.0"
+		"strip-final-newline": "^3.0.0"
 	},
 	"devDependencies": {
-		"@types/node": "^14.14.10",
-		"ava": "^2.4.0",
+		"@types/node": "^16.11.7",
+		"ava": "^3.15.0",
```
ehmicky, line 56: Thanks for fixing the Ava update :+1:

index.d.ts
NMinhNguyen: I think I had to update `xo` for ESM (see other PRs by Sindre where he updated his dependencies). Don't remember the exact reasons though. I'd recommend reviewing with `?w=1` in the URL, it helps a lot.

It's a bit tricky to not have indentation changes because I had to "unwrap" definitions and hoist them to the top level. Maybe I can just apply Prettier in a separate commit, and then my code changes on top?

test/command.js
```diff
@@ -1,15 +1,17 @@
-import path from 'path';
+import path from 'node:path';
+import process from 'node:process';
+import {fileURLToPath} from 'node:url';
 import test from 'ava';
-import execa from '..';
+import * as execa from '../index.js';
```
ehmicky, line 5: Could we use named imports instead?

test/error.js
```diff
@@ -124,26 +127,34 @@ test('result.killed is false on process error', async t => {
...
 
 if (process.platform === 'darwin') {
-	test.cb('sanity check: child_process.exec also has killed.false if killed indirectly', t => {
-		const {pid} = childProcess.exec('noop', error => {
-			t.truthy(error);
-			t.false(error.killed);
-			t.end();
+	test('sanity check: child_process.exec also has killed.false if killed indirectly', async t => {
+		const promise = new Promise((resolve, reject) => {
+			const {pid} = childProcess.exec('noop', error => {
```
ehmicky, line 138: If we use promises, we might want to use `const pExec = utils.promisify(childProcess.exec)` instead of `new Promise`.

test/command.js
NMinhNguyen: See PR description and Sindre's comment but the reason I did a namespace import here was to keep the diff smaller. Will probably have to change anyway.

test/fixtures/detach.js
ehmicky: I can reproduce the bug you are mentioning.
I don't think this deserves a code comment since using an explicit exit code instead of making it implicit seems a rather normal best practice.

test/command.js
ehmicky: Yes, I believe this would be better to use named imports there, also for consistency with the other test files, despite the fact that it will make the PR diff bigger.

index.d.ts
ehmicky: Yes, I understand this might be a little of a chicken-and-egg problem there.
If Sindre is doing it like this on other repositories, let's be consistent and update both at once :+1:

test/node.js
ehmicky: Interesting. Your explanation is a possible cause.
Regardless, the core of the test, i.e. testing the forked script's communication channel is kept, so this looks good to me. Thanks for fixing it :+1:

index.d.ts
```diff
@@ -486,23 +488,26 @@ declare const execa: {
...
+	options?: SyncOptions<null>
+): ExecaSyncReturnValue<Buffer>;
+export function execaSync(
+	file: string,
+	options?: SyncOptions
+): ExecaSyncReturnValue;
+export function execaSync(
+	file: string,
+	options?: SyncOptions<null>
+): ExecaSyncReturnValue<Buffer>;
```
ehmicky, line 508: This looks so much cleaner.

index.test-d.ts
```diff
@@ -139,55 +151,55 @@ execa('unicorns').kill('SIGKILL', {forceKillAfterTimeout: false});
...
-expectType<ExecaReturnValue<string>>(await execa.node('unicorns'));
-expectType<ExecaReturnValue<string>>(
-	await execa.node('unicorns', {encoding: 'utf8'})
+expectType<ExecaChildProcess>(node('unicorns'));
+expectType<ExecaReturnValue>(await node('unicorns'));
+expectType<ExecaReturnValue>(
+	await node('unicorns', {encoding: 'utf8'}),
 );
-expectType<ExecaReturnValue<Buffer>>(await execa.node('unicorns', {encoding: null}));
-expectType<ExecaReturnValue<string>>(
```
ehmicky, line 188: Nice catch with the unnecessary `<string>`. :+1:

NMinhNguyen: Haha wasn't me. It was all `xo --fix` :)

ehmicky (changes requested):
Thank you so much for working on this @NMinhNguyen, this is super awesome! That's a huge amount of work.

I reviewed it thoroughly and this is looking pretty close to perfect to me, except a few points:
  - Renaming the exported function names. The function names you've suggested sound good to me too. I am wondering if it might worth doing in a second/follow-up PR, to minimize the diff?
  - [Using named imports in all test files.](https://github.com/sindresorhus/execa/pull/478#discussion_r749638868)
  - Small thing about [a specific test file](https://github.com/sindresorhus/execa/pull/478#discussion_r749640903).

Thanks a lot for focusing this PR only on the conversion itself, without changing any logic, except for dependencies upgrade, that's very helpful. :heart:

NMinhNguyen:
Thanks for the thorough review @ehmicky! I'll see if I can split the changes into a more reviewable diff when I get home tonight.

ehmicky:
Note: I think only the exported function names would need to be split into a separate PR, and that renaming has not been done yet. In particular, we do not need to split a PR for the `xo` update, per our above discussion.

NMinhNguyen:
> Note: I think only the exported function names would need to be split into a separate PR, and that renaming has not been done yet. In particular, we do not need to split a PR for the `xo` update, per our above discussion.

Ah I meant I'd try to split this single commit into smaller ones for a better diff, unless you've already thoroughly reviewed all the changes? I find that `?w=1` really helps in these situations.

ehmicky:
Breaking it down to smaller commits might help @sindresorhus, since he might want to do a second round of review before we merge this PR. But as far as I am concerned, I've reviewed all of this PR, so I would not benefit from breaking down the commits.

test/error.js
NMinhNguyen: Thanks for pointing it out! Thanks to you, I was able to spot a bug - it's meant to say `'noop.js'`. I'll make sure to search for all references to the new `.js` fixtures and ensure they have an extension.

lib/stream.js
```diff
@@ -77,21 +76,13 @@ const getSpawnedResult = async ({stdout, stderr, all}, {encoding, buffer, maxBuf
...
 };
 
-const validateInputSync = ({input}) => {
+export const validateInputSync = ({input}) => {
 	if (isStream(input)) {
 		throw new TypeError('The `input` option cannot be a stream in sync mode');
 	}
 };
-
-module.exports = {
```
NMinhNguyen, line 91: Is there a preference for `export { foo, bar } `  vs individually exporting (`export const foo`)?

readme.md
```diff
@@ -108,7 +108,7 @@ const execa = require('execa');
 
 ```js
 try {
```
NMinhNguyen, line 110: Should I import `execaSync` here?

index.d.ts
NMinhNguyen: Actually this is really bugging me, so I'll take another stab at reducing the whitespace diff (similar to https://github.com/sindresorhus/is-stream/commit/5831295859122e5a246f31d6b92c8d5e8b551d5e?w=1)

NMinhNguyen:
@ehmicky I've pushed separate commits (so you can review them more easily) to

- restore the original formatting
- update the named exports

I believe that these changes aren't too bad to review now (remember to use `?w=1`) but if you have any objections, it is trivial for me to drop them.

test/error.js
```diff
@@ -124,26 +130,27 @@ test('result.killed is false on process error', async t => {
...
-	test.cb('sanity check: child_process.exec also has killed.false if killed indirectly', t => {
-		const {pid} = childProcess.exec('noop', error => {
-			t.truthy(error);
-			t.false(error.killed);
-			t.end();
-		});
-
-		process.kill(pid, 'SIGINT');
+	test('sanity check: child_process.exec also has killed.false if killed indirectly', async t => {
+		const pExec = promisify(childProcess.exec);
```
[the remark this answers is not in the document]
NMinhNguyen: Happy to make that change. My thinking was it was only used in this test so I colocated it with the test, but happy to move it. Won't be able to for a good few hours though, so feel free to push it yourself.

NMinhNguyen:
Pushed a change that restores running tests against Node.js 12. @ehmicky let me know if there's anything further you'd like me to do.

sindresorhus:
There are a couple of outdated references:

> test/stdio.js
51:forkMacro.title = macroTitle('execa.fork()');

> lib/command.js
26:// Handle `execa.command()`

sindresorhus:
It would also be great if you could remove all the `(async () => {` wrappers in the docs.

NMinhNguyen:
> There are a couple of outdated references:
> 
> > test/stdio.js
> > 51:forkMacro.title = macroTitle('execa.fork()');

What should this be updated to since there's no `execaFork`?

By the way, I found another way to rewrite the `override-promise.js` test such that it works in Node.js 12 but I'm not sure if it's preferable as it's a little bit hacky:

```js
test('should work with third-party Promise', async t => {
	// Introduce an artificial delay to avoid breaking `ava`.
	await Promise.resolve();

	// Can't use `test.before`, because `ava` needs `Promise`.
	const nativePromise = Promise;
	global.Promise = class BrokenPromise {
		then() {
			throw new Error('error');
		}
	};
	const {execa} = await import('../index.js');
	global.Promise = nativePromise;

	const {stdout} = await execa('noop.js', ['foo']);
	t.is(stdout, 'foo');
});
```

sindresorhus:
> What should this be updated to since there's no execaFork?

`execaNode`

sindresorhus:
> By the way, I found another way to rewrite the override-promise.js test such that it works in Node.js 12 but I'm not sure if it's preferable as it's a little bit hacky:

We can just keep the existing workaround until we can target Node.js 14.

sindresorhus:
Thanks for working on this, @NMinhNguyen 👍

PR #490, opened 2022-02-04
Support `AbortController`

### LINKED ISSUE #449: Support `AbortController`
Node.js is adopting it for everything, so I think we should support it too.

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

Node.js just added support for it in `.spawn()`: https://nodejs.org/en/blog/release/v15.5.0/

This would not replace the current `.cancel()` function, in the near-term at least.

jopemachine:
Fixes #449.

I'd appreciate it if you could let me know if I need to write more tests or documents.

Thanks for all your efforts for this awesome library :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -205,6 +205,24 @@ export interface CommonOptions<EncodingType> {
 	readonly killSignal?: string | number;
 
 	/**
+	You can abort the spawned process by the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+	It works on Node 16 and higher versions.
+
+	@example
+	```
+	import {execa} from 'execa';
+
+	const abortController = new AbortController();
+	const subprocess = execa('node', [], { abortSignal: abortController.signal });
+
+	abortController.abort();
+
+	console.log(subprocess.killed); // true
+	```
+	*/
+	readonly abortSignal?: AbortSignal;
+
+	/**
 	If `true`, no quoting or escaping of arguments is done on Windows. Ignored on other platforms. This is set to `true` automatically when the `shell` option is `true`.
 
 	@default false
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,6 +207,8 @@ export interface CommonOptions<EncodingType> {
 	/**
 	You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
 
+	When `AbortController.abort()` is called, [isCanceled](https://github.com/sindresorhus/execa#iscanceled) becomes `false`.
+
 	*Requires Node.js 16 or later.*
 
 	@example
@@ -363,6 +365,8 @@ export interface ExecaReturnValue<StdoutErrorType = string>
 
 	/**
 	Whether the process was canceled.
+
+	You can cancel the spawned process using the [`signal`](https://github.com/sindresorhus/execa#signal-1) option.
 	*/
 	isCanceled: boolean;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,7 +207,7 @@ export interface CommonOptions<EncodingType> {
 	/**
 	You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
 
-	When `AbortController.abort()` is called, [isCanceled](https://github.com/sindresorhus/execa#iscanceled) becomes `false`.
+	When `AbortController.abort()` is called, [`.isCanceled`](https://github.com/sindresorhus/execa#iscanceled) becomes `false`.
 
 	*Requires Node.js 16 or later.*
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -210,16 +210,22 @@ export interface CommonOptions<EncodingType> {
 	*Requires Node.js 16 or later.*
 
 	@example
-	```
+	```js
 	import {execa} from 'execa';
 
 	const abortController = new AbortController();
 	const subprocess = execa('node', [], {signal: abortController.signal});
 
-	abortController.abort();
+	setTimeout(() => {
+		abortController.abort();
+	}, 1000);
 
-	console.log(subprocess.killed);
-	//=> true
+	try {
+		await subprocess;
+	} catch (error) {
+		console.log(subprocess.killed); // true
+		console.log(error.isCanceled); // true
+	}
 	```
 	*/
 	readonly signal?: AbortSignal;
--- a/index.d.ts
+++ b/index.d.ts
@@ -205,19 +205,21 @@ export interface CommonOptions<EncodingType> {
 	readonly killSignal?: string | number;
 
 	/**
-	You can abort the spawned process by the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
-	It works on Node 16 and higher versions.
+	You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+
+	*Requires Node.js 16 or later.*
 
 	@example
 	```
 	import {execa} from 'execa';
 
 	const abortController = new AbortController();
-	const subprocess = execa('node', [], { signal: abortController.signal });
+	const subprocess = execa('node', [], {signal: abortController.signal});
 
 	abortController.abort();
 
-	console.log(subprocess.killed); // true
+	console.log(subprocess.killed);
+	//=> true
 	```
 	*/
 	readonly signal?: AbortSignal;
--- a/index.d.ts
+++ b/index.d.ts
@@ -213,14 +213,14 @@ export interface CommonOptions<EncodingType> {
 	import {execa} from 'execa';
 
 	const abortController = new AbortController();
-	const subprocess = execa('node', [], { abortSignal: abortController.signal });
+	const subprocess = execa('node', [], { signal: abortController.signal });
 
 	abortController.abort();
 
 	console.log(subprocess.killed); // true
 	```
 	*/
-	readonly abortSignal?: AbortSignal;
+	readonly signal?: AbortSignal;
 
 	/**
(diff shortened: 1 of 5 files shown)


ehmicky:
Thanks a lot for this contribution @jopemachine!

We might want to name this option `signal` instead of `abortSignal` to match the name currently used by `child_process.spawn()`.

More fundamentally, I am wondering whether the core `signal` option should just be used instead? I have written some potential ideas in [the issue](https://github.com/sindresorhus/execa/issues/449#issuecomment-1030065002), please let me know what you think.

jopemachine:
> We might want to name this option signal instead of abortSignal to match the name currently used by child_process.spawn().

Thanks for pointing it out, I just updated the option.

index.d.ts
```diff
@@ -204,6 +204,26 @@ export interface CommonOptions<EncodingType> {
...
+
+	const abortController = new AbortController();
+	const subprocess = execa('node', [], {signal: abortController.signal});
+
+	abortController.abort();
+
+	console.log(subprocess.killed);
+	//=> true
+	```
+	*/
```
sindresorhus, line 224: The same docs should be in the readme too.

sindresorhus (changes requested):

index.d.ts
jopemachine: Could we add an example [here](https://github.com/sindresorhus/execa/pull/490/commits/7ed6580e7ed657be1d7e16349ec85f17777ba2e0#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R567) link to avoid document duplication?

readme.md
```diff
@@ -97,6 +99,20 @@ try {
...
+import {execa} from 'execa';
+
+const abortController = new AbortController();
+const subprocess = execa('node', [], {signal: abortController.signal});
+
+abortController.abort();
+
+console.log(subprocess.killed);
+//=> true
+```
```
sindresorhus, line 114: Move this to the `signal` option.

jopemachine:
Should I delete [this document](https://github.com/sindresorhus/execa/blob/29bc38e6ed2ce8add74baa5a06555f97edd43a14/readme.md#cancel) too? or should I mark the cancel method as `deprecated`?

index.js
```diff
@@ -109,6 +109,12 @@ export function execa(file, args, options) {
 	spawned.kill = spawnedKill.bind(null, spawned.kill.bind(spawned));
 	spawned.cancel = spawnedCancel.bind(null, spawned, context);
 
+	if (parsed.options.signal) {
+		parsed.options.signal.addEventListener('abort', () => {
```
ehmicky, line 113: When the `signal` option is passed to `child_process.spawn()`, `spawned.kill()` is already automatically called by Node.js when the `abort` event is emitted.

`spawned.cancel()` also sets `error.isCanceled: true`. However, we could set it instead by replacing the following line:

https://github.com/sindresorhus/execa/blob/e4b929547ade335b0dd9d7afe1a35989e19fbc8a/index.js#L130

By:

```js
    isCanceled: parsed.options.signal && parsed.options.signal.aborted,
```

readme.md
```diff
@@ -542,6 +523,26 @@ Default: `SIGTERM`
 
 Signal value to be used when the spawned process will be killed.
 
+#### signal
+
+Type: `object (AbortSignal)`
+
+You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+
+```js
```
ehmicky, line 532: Could you please move this example to the `Cancelling a spawned process` section above? For consistency, since all examples are currently grouped in the "Usage" and "Tips" sections. Thanks!

readme.md
```diff
@@ -78,25 +78,6 @@ try {
...
-
-setTimeout(() => {
-	subprocess.cancel();
-}, 1000);
-
-try {
-	await subprocess;
-} catch (error) {
-	console.log(subprocess.killed); // true
-	console.log(error.isCanceled); // true
```
ehmicky, line 96: We might want to keep the following from this example:
  - Printing `isCanceled`
  - Showing that aborts make the process rejects its returned promise
  - Calling `abort()` inside `setTimeout()`, since this mimics how it would be used in real use cases (calling it right after the process has started is unlikely)

index.d.ts
```diff
@@ -204,6 +204,26 @@ export interface CommonOptions<EncodingType> {
 	*/
 	readonly killSignal?: string | number;
 
+	/**
```
ehmicky, line 207: Should we add some test for this new type in `index.test-d.ts`?

test/kill.js
```diff
@@ -262,3 +262,48 @@ test('calling cancel method on a process which has been killed does not make err
 	const {isCanceled} = await t.throwsAsync(subprocess);
 	t.false(isCanceled);
 });
+
+test('calling abort throws an error with message "Command was canceled"', async t => {
```
ehmicky, line 266: Nice tests, they look great!

readme.md
```diff
@@ -78,25 +78,6 @@ try {
 }
 ```
 
-### Cancelling a spawned process
-
```
ehmicky, line 82: We should also remove the documentation [of the `cancel()` method](https://github.com/sindresorhus/execa/blob/main/readme.md#cancel).

ehmicky (changes requested):
I believe we can directly delete that section of the documentation, good catch! :+1: 

Thanks again for this PR @jopemachine! There are a couple of things I have noted. The tests look great!

jopemachine:
> I believe we can directly delete that section of the documentation, good catch! 👍
> 
> Thanks again for this PR @jopemachine! There are a couple of things I have noted. The tests look great! (But the CI tests are currently failing)

I appreciate your detailed feedback :)
I tried to reflect on all the feedback as much as I can.
If I'm missing something or got something wrong, I appreciate your letting me know.

index.js
```diff
@@ -147,7 +149,7 @@ export function execa(file, args, options) {
 			all,
 			failed: false,
 			timedOut: false,
-			isCanceled: false,
```
ehmicky, line 150: `isCanceled` is always `false` when the process is successful.

readme.md
```diff
@@ -542,6 +541,16 @@ Default: `SIGTERM`
 
 Signal value to be used when the spawned process will be killed.
 
+#### signal
+
+Type: `object (AbortSignal)`
+
+You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+
+When the `abort` is called, [isCancel](#iscanceled) becomes false.
```
ehmicky, line 550:
```suggestion
When `AbortController.abort()` is called, [isCanceled](#iscanceled) becomes `false`.
```

index.js
```diff
@@ -109,6 +109,8 @@ export function execa(file, args, options) {
 	spawned.kill = spawnedKill.bind(null, spawned.kill.bind(spawned));
 	spawned.cancel = spawnedCancel.bind(null, spawned, context);
 
+	const isCanceled = () => context.isCanceled || (parsed.options.signal ? parsed.options.signal.aborted : false);
```
ehmicky, line 112:
```suggestion
	const isCanceled = () => context.isCanceled || (parsed.options.signal && parsed.options.signal.aborted);
```

test/kill.js
ehmicky, line 266: Those tests fail on older Node.js versions due to `AbortController` not being supported. They should be put inside a condition, maybe something like this?

```js
if (globalThis.AbortController !== undefined) {
```

ehmicky:
Thanks a lot for those edits, they look great! And the tests are now passing on Node 16 :tada: 
I have a few more comments, but we're getting there. Thanks for your help @jopemachine.

index.js
jopemachine: It seems that when `signal` option is not using and context.isCanceled is false, 
isCanceled is `undefined` (It should be false in this case) (`false || undefined === undefined`)
Would it be better to set `signal` option's default value to false to handle this?

ehmicky: Good point! :+1:

index.js
```diff
@@ -127,7 +129,7 @@ export function execa(file, args, options) {
 				escapedCommand,
 				parsed,
 				timedOut,
-				isCanceled: context.isCanceled,
```
ehmicky, line 130: Should `isCanceled()` be directly inlined there?

```suggestion
				isCanceled: context.isCanceled || (parsed.options.signal ? parsed.options.signal.aborted : false);
```

ehmicky (approved):
This looks good to me, thanks a lot @jopemachine!

Now waiting for another review from @sindresorhus.

readme.md
```diff
@@ -290,6 +287,8 @@ Type: `boolean`
 
 Whether the process was canceled.
 
+You can cancel the spawned process using the [signal](#signal-1) option.
```
sindresorhus, line 290: This needs to be added to the TS docs too.

sindresorhus, line 290:
```suggestion
You can cancel the spawned process using the [`signal`](#signal-1) option.
```

readme.md
```diff
@@ -542,6 +541,16 @@ Default: `SIGTERM`
 
 Signal value to be used when the spawned process will be killed.
 
+#### signal
+
+Type: `object (AbortSignal)`
```
sindresorhus, line 546:
```suggestion
Type: [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)
```

readme.md
```diff
@@ -542,6 +541,16 @@ Default: `SIGTERM`
...
+
+You can abort the spawned process using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+
+When `AbortController.abort()` is called, [isCanceled](#iscanceled) becomes `false`.
```
sindresorhus, line 550: The readme and TS docs should be in sync.

sindresorhus (changes requested):

PR #496, opened 2022-03-02
URL is not standard

Bohdan-Anderson-Rangle:
URL is not common everywhere - adding this resolves a bug in some setups
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/19799

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,7 @@
 import {Buffer} from 'node:buffer';
 import {ChildProcess} from 'node:child_process';
 import {Stream, Readable as ReadableStream} from 'node:stream';
-import { URL } from 'url';
+import {URL} from 'url';
 
 export type StdioOption =
 	| 'pipe'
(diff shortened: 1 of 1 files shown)


ehmicky:
Hi @Bohdan-Anderson-Rangle,

`URL` became [a global object with Node.js 10.0.0](https://nodejs.org/api/url.html#class-url). While it does look like `@types/dom` was previously required, the issue above is from 2017, and I am assuming the latest version of typescript might have fix this.

The following file is transpiled `tsc` without errors for me with Node.js `17.6.0`, TypeScript `4.6.2` and no `tsconfig.json`:

```ts
const url: URL = new URL('https://example.com')
```

Which version of `node` and `typescript` are you using? Also, what's your `tsconfig.json` and `tsc` CLI flags?

sindresorhus:
This was an issue before, but it should be fixed now. See: https://github.com/sindresorhus/globby/pull/230

Bohdan-Anderson-Rangle:
100% it was an issue with my tsconfig thank you for calling that out.

PR #497, opened 2022-05-09
Drop support for Node 12

ehmicky:
This drops support for Node 12 (breaking change) since it is not maintained anymore.
This also adds CI tests for Node 18, and upgrades some GitHub actions.

The also drops Node `14.13.1 - 14.17.0` and `16.0.0 - 16.3.0` in order to support unconditionally several features: `cwd: URL`, `stdio: 'overlapped'` and the `signal` option.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,21 +10,21 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 16
-          - 14
-          - 12
+          - 18.0.0
+          - 16.4.0
+          - 14.18.0
         os:
           - ubuntu-latest
           - macos-latest
           # - windows-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
       - run: npm test
       - uses: codecov/codecov-action@v2
-        if: matrix.os == 'ubuntu-latest' && matrix.node-version == 16
+        if: matrix.os == 'ubuntu-latest' && matrix.node-version == 18.0.0
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -10,21 +10,21 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 16
-          - 14
-          - 12
+          - 18.0.0
+          - 16.4.0
+          - 14.18.0
```
sindresorhus, line 15: I would prefer using non-specific minors so it upgrades when new minor versions are out.

package.json
```diff
@@ -13,7 +13,7 @@
 	"type": "module",
 	"exports": "./index.js",
 	"engines": {
-		"node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+		"node": "^14.18.0 || ^16.4.0 || >=18.0.0"
```
sindresorhus, line 16:
```suggestion
		"node": "^14.18.0 || ^16.14.0 || >=18.0.0"
```

.github/workflows/main.yml
ehmicky: Fixed!

BrycensRanch:
How did you not publish a breaking change for months? I'm still learning Semantic Versioning but if I was to merge a breaking change it'd be immediately added to releases

ehmicky:
Hi @BrycensRanch, [`7.0.0`](https://github.com/sindresorhus/execa/releases/tag/v7.0.0) was just released. :+1:

BrycensRanch:
oh

MaksimMedvedev:
Hi! I see you mentioned dropping support for node < 16.3, but on `7.1.1` the `package.json` indicates `16.14.0` as a minimal version. I could consider asking our DevOps to update node but is there a reason you narrow 16.x version of node this tight? Thanks!

PR #500, opened 2022-05-17
Re-enable Windows in CI

ehmicky:
This fixes the tests on Windows CI, and re-enables it. :tada:


test/command.js
```diff
@@ -1,10 +1,8 @@
-import path from 'node:path';
-import process from 'node:process';
-import {fileURLToPath} from 'node:url';
 import test from 'ava';
 import {execa, execaSync, execaCommand, execaCommandSync} from '../index.js';
+import {setFixtureDir} from './helpers/fixtures-dir.js';
 
-process.env.PATH = fileURLToPath(new URL('fixtures', import.meta.url)) + path.delimiter + process.env.PATH;
+setFixtureDir();
```
ehmicky, line 5: How the `PATH`/`Path` environment variable in the GitHub actions Windows environment seems to have changed, making the CI tests fail. In particular, its name is now `Path` instead of `PATH`.
Using `path-key` fixes this. I put this as a shared helper to avoid code duplication.

test/test.js
```diff
@@ -76,16 +77,23 @@ test('stripFinalNewline in sync mode on failure', t => {
...
+
+const BIN_DIR_REGEXP = /node_modules[\\/]\.bin/;
+
 test('preferLocal: true', async t => {
-	await t.notThrowsAsync(execa('ava', ['--version'], {preferLocal: true, env: {Path: '', PATH: ''}}));
+	await t.notThrowsAsync(execa('ava', ['--version'], {preferLocal: true, env: getPathWithoutLocalDir()}));
 });
 
 test('preferLocal: false', async t => {
-	await t.throwsAsync(execa('ava', ['--version'], {preferLocal: false, env: {Path: '', PATH: ''}}), {message: ENOENT_REGEXP});
```
ehmicky, line 84: I am not completely sure why, but this fixes those tests on Windows.
Filtering the `.bin` directories from the `PATH` variable seems more adequate than making it completely empty as well.

test/test.js
```diff
@@ -189,12 +197,12 @@ test('use environment variables by default', async t => {
...
 
 test('extend environment variables by default', async t => {
-	const {stdout} = await execa('environment.js', [], {env: {BAR: 'bar'}});
+	const {stdout} = await execa('environment.js', [], {env: {BAR: 'bar', [PATH_KEY]: process.env[PATH_KEY]}});
 	t.deepEqual(stdout.split('\n'), ['foo', 'bar']);
 });
 
 test('do not extend environment with `extendEnv: false`', async t => {
-	const {stdout} = await execa('environment.js', [], {env: {BAR: 'bar', PATH: process.env.PATH}, extendEnv: false});
+	const {stdout} = await execa('environment.js', [], {env: {BAR: 'bar', [PATH_KEY]: process.env[PATH_KEY]}, extendEnv: false});
```
ehmicky, line 205: This fixes the tests on Windows.

sindresorhus:
Good investigation 👍

PR #507, opened 2022-08-06
Fix MarkDown in TSDoc comments

tukusejssirs:
Changes:

- add blank line:
   - before lists;
   - between two lines of text which seem to be two separate paragraph (e.g. [`index.d.ts` (L290-291)](https://github.com/sindresorhus/execa/blob/main/index.d.ts#L290-L291));
- remove indentation before first-level in lists;
- specify syntax highlighting language as `typescript` for code blocks when missing or when `js` was used.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,7 +26,7 @@ export interface CommonOptions<EncodingType> {
 	/**
 	Prefer locally installed binaries when looking for a binary to execute.
 
-	If you `npm install foo`, you can then `execa('foo')`.
+	If you `$ npm install foo`, you can then `execa('foo')`.
 
 	@default false
 	*/
@@ -65,7 +65,7 @@ export interface CommonOptions<EncodingType> {
 
 	/**
 	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio).
-   *
+
 	@default 'pipe'
 	*/
 	readonly stdin?: StdioOption;
@@ -138,7 +138,7 @@ export interface CommonOptions<EncodingType> {
 
 	@default 'pipe'
 	*/
-	readonly stdio?: 'pipe' | 'ignore' | 'inherit' | readonly StdioOption[];
+	readonly stdio?: 'pipe' | 'overlapped' | 'ignore' | 'inherit' | readonly StdioOption[];
 
 	/**
 	Specify the kind of serialization used for sending messages between processes when using the `stdio: 'ipc'` option or `execaNode()`:
@@ -353,15 +353,15 @@ export interface ExecaSyncReturnValue<StdoutErrorType = string>
 
 /**
 Result of a child process execution. On success this is a plain object. On failure this is also an `Error` instance.
- *
+
 The child process fails when:
- *
+
 - its exit code is not `0`
 - it was killed with a signal
 - timing out
 - being canceled
 - there's not enough memory or there are already too many child processes
- */
+*/
 export interface ExecaReturnValue<StdoutErrorType = string>
 	extends ExecaSyncReturnValue<StdoutErrorType> {
 	/**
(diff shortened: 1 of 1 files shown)


tukusejssirs:
### Why this is needed?

VS Code outputs TSDoc comments correctly even without these changes, however Sublime Text (using LSP + TypeScript server, using [`sublime-markdown-popups`](https://github.com/facelessuser/sublime-markdown-popups)) does not output the HTML (converted from MarkDown) correctly.

It might be considered a bug in `sublime-markdown-popups`, however, I think that TSDoc comments _should_ have leading asterisk on each comment line.

Note: It might be good idea to add missing TSDocs for `ExecaSyncReturnValue` ([`index.d.ts` (L340)](https://github.com/sindresorhus/execa/blob/main/index.d.ts#L340)).

sindresorhus:
> It might be considered a bug in sublime-markdown-popups, however, I think that TSDoc comments should have leading asterisk on each comment line.

You should report that to that project. I prefer without leading asterisk and will not change that.

tukusejssirs:
Okay, thanks for your reply. "hat about the other changes I suggest

> Changes:
>
> - ~~fix TSDoc comments by adding leading asterisks;~~
> - add blank line:
>    - before lists;
>    - between two lines of text which seem to be two separate paragraph (e.g. [`index.d.ts` (L290-291)](https://github.com/sindresorhus/execa/blob/main/index.d.ts#L290-L291));
> - remove indentation before first-level in lists;
> - specify syntax highlighting language as `typescript` for code blocks when missing or when `js` was used.

tukusejssirs:
I have removed the leading asterisks from TSDoc comments. I’ll squash the commits later.

Are the current changes acceptable to you? :smiley:

index.d.ts
```diff
@@ -137,8 +142,11 @@ export interface CommonOptions<EncodingType> {
 
 	/**
 	Specify the kind of serialization used for sending messages between processes when using the `stdio: 'ipc'` option or `execaNode()`:
-		- `json`: Uses `JSON.stringify()` and `JSON.parse()`.
-		- `advanced`: Uses [`v8.serialize()`](https://nodejs.org/api/v8.html#v8_v8_serialize_value)
+
+	- `json`: Uses `JSON.stringify()` and `JSON.parse()`.
+	- `advanced`: Uses [`v8.serialize()`](https://nodejs.org/api/v8.html#v8_v8_serialize_value)
+
+	Requires Node.js `13.2.0` or later.
```
sindresorhus, line 149: This is moot. Execa requires Node.js 14.

index.d.ts
```diff
@@ -15,8 +15,9 @@ export type StdioOption =
 export interface CommonOptions<EncodingType> {
 	/**
 	Kill the spawned process when the parent process exits unless either:
-		- the spawned process is [`detached`](https://nodejs.org/api/child_process.html#child_process_options_detached)
-		- the parent process is terminated abruptly, for example, with `SIGKILL` as opposed to `SIGTERM` or a normal exit
+
```
sindresorhus, line 18: I don't really see the point in having an empty line before the list?

index.d.ts
```diff
@@ -460,7 +474,7 @@ Think of this as a mix of `child_process.execFile` and `child_process.spawn`.
 @returns A [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess), which is enhanced to also be a `Promise` for a result `Object` with `stdout` and `stderr` properties.
 
 @example
-```
+```typescript
```
sindresorhus, line 477: TypeScript is inferred. So you don't need any specifier.

index.d.ts
```diff
@@ -472,14 +486,14 @@ console.log(stdout);
 const subprocess = execa('node');
 
 setTimeout(() => {
-	subprocess.cancel()
+  subprocess.cancel()
```
sindresorhus, line 489: Don't make unrelated changes. It should use tab indentation.

sindresorhus (changes requested):

index.d.ts
tukusejssirs: Fix in latest commit.

tukusejssirs (commented):
I fixed most issues raised and for the rest I tried to provide a valid reason why I need it formatted like I propose, which while not required per se, it is optional in GitHub Flavoured Markdown as well as CommonMark.

I hope you’ll understand, @sindresorhus. :pray:

sindresorhus:
I have to be honest, I'm not really interested in working around `sublime-markdown-popups` limitations. CommonMark is what I'm used to and prefer.

sindresorhus:
> It is inferred in VS Code, but not in ST. When it is not specified, it’ll be output in single colour (no syntax highlighting), see below.

Seems like you should open an issue on ST or the plugin.

PR #510, opened 2022-11-20
Add `$` tagged templates API

### LINKED ISSUE #484: Steal some ideas/syntax from google/zx
I think zx has a very sexy syntax:
```
await $`echo ${things}`
```
However, lots of little things about it I find confusing. (quoting behavior, weird dollar signs added into final commands: `git add $'.*ignore'`, etc. I just want it to run the string that I give it, no modifications at all.)

I wonder if execa were to try and produce a similarly seemingly-simple "sexy" `$` function what would it be?

This is more of a discussion, not an issue per se.

aaronccasanova:
Fixes #484 

This PR introduces a `$` API inspired by [google/zx](https://www.npmjs.com/package/zx), which allows users to execute commands using familiar `execa` options with the elegant DX of [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates):
 
#### Basic
```js
import {$} from 'execa';

const {stdout} = await $`echo unicorns`;
// const {stdout} = await $`echo ${'unicorns'}`;
// const {stdout} = await $`echo ${['unicorns', 'rainbows']}`;

console.log(stdout);
//=> 'unicorns'
```

#### With options
```js
import {$} from 'execa';

await $({stdio: 'inherit'})`echo unicorns`;
//=> 'unicorns'
```

#### With pre-defined options
```js
import {$} from 'execa';

const $$ = $({stdio: 'inherit'});

await $$`echo unicorns`;
//=> 'unicorns'
await $$({shell: true})`echo unicorns && echo rainbows`;
//=> 'unicorns'
//=> 'rainbows'
```

#### Synchronous
```js
import {$} from 'execa';

const {stdout} = $.sync`echo unicorns`;
console.log(stdout);
//=> 'unicorns'

$({stdio: 'inherit'}).sync`echo rainbows`;
//=> 'rainbows'
```

Co-authored-by: ehmicky <ehmicky@users.noreply.github.com>


index.js
```diff
@@ -224,6 +224,43 @@ export function execaSync(file, args, options) {
...
+	let command = '';
+	for (const [i, template] of templates.entries()) {
+		command += `${template}${parseExpression(expressions[i])}`;
+	}
+
+	return command;
+}
+
+function parseExpression(expression) {
+	return Array.isArray(expression) ? expression.join(' ') : expression ?? '';
```
ehmicky, line 261: The current logic would remove whitespace escaping. For example:

```js
$`file ${'Hello world'}` -> execa('file', ['Hello', 'world'])
```

instead of:

```js
$`file ${'Hello world'}` -> execa('file', ['Hello world'])
```

Also `parseCommand()` allow backslashes as escape characters, even though they are not needed there, so this function might not fit there. Calling just `string.split()` should be enough actually.

We also need to take into account that `template` might be `undefined`, when the template string contains invalid backlash sequences like an isolated `\u`. 

One possible implementation could be:

```js
const $ = (templates, ...expressions) => {
	const {file, args} = parseTemplates(templates, expressions);
	return execa(file, args);
};

const parseTemplates = (templates, expressions) => {
	const [file, ...args] = templates.flatMap((template, index) => parseTemplate(template, index, templates, expressions));
	return {file, args};
};

const parseTemplate = (template, index, templates, expressions) => {
	const templateString = template ?? templates.raw[index];
	const templateTokens = templateString.split(SPACES_REGEXP).filter(Boolean);

	if (index === expressions.length) {
		return templateTokens;
	}

	const expression = expressions[index];
	return Array.isArray(expression) ? [...templateTokens, ...expression.map(String)] : [...templateTokens, String(expression)];
};

const SPACES_REGEXP = / +/g;
```

I left the sync logic and the options binding logic out for the sake of the example.

This produces the following result:

```js
$`file one two` -> execa(file, ['one', 'two'])
$`file  one  two` -> execa(file, ['one', 'two'])
$`file ${'one'} ${'two'}` -> execa(file, ['one', 'two'])
$`file ${0}` -> execa(file, ['0'])
$`file "one two"` -> execa(file, ['"one', 'two"'])

$`file ${'one two'}` -> execa(file, ['one two'])
$`file ${['one', 'two three']}` -> execa(file, ['one', 'two three'])
$`file \n` -> execa(file, ['\n'])
$`file ${'\n'}` -> execa(file, ['\n'])
$`file \u` -> execa(file, ['\\u'])
$`file ${undefined}` -> execa(file, ['undefined'])
$`file one\\ two` -> execa(file, ['one\\', 'two'])
$`file ${'"one two"'}` -> execa(file, ['"one two"'])
```

Compared to the current implementation:

```js
$`file one two` -> execa(file, ['one', 'two'])
$`file  one  two` -> execa(file, ['one', 'two'])
$`file ${'one'} ${'two'}` -> execa(file, ['one', 'two'])
$`file ${0}` -> execa(file, ['0'])
$`file "one two"` -> execa(file, ['"one', 'two"'])

$`file ${'one two'}` -> execa(file, ['one', 'two'])
$`file ${['one', 'two three']}` -> execa(file, ['one', 'two', 'three'])
$`file \n` -> execa(file, [])
$`file ${'\n'}` -> execa(file, [])
$`file \u` -> execa(undefined, [])
$`file ${undefined}` -> execa(file, [])
$`file one\\ two` -> execa(file, ['one two'])
$`file ${'"one two"'}` -> execa(file, ['"one', 'two"'])
```

What do you think?

ehmicky:
Thanks a lot for tackling this and for navigating the back-and-forth of the discussion about escaping! :heart: 

Before digging too deep into the review, I left a general comment about escaping. Please let me know what you think. Thanks!

index.js
aaronccasanova: Ah! Okay these examples are great and really drive home your point in our original discussion:

> Which means that quotes and backslashes do not need to be interpreted as escaping characters, since ${...} already fulfills escaping needs. They would be interpreted as regular characters instead.

I'm totally on board with this approach! Also, thanks for the sample implementation. It works great!
https://github.com/sindresorhus/execa/pull/510/commits/8aa4e0714cb98198641eed48b9aeb7266f6c4346

index.d.ts
```diff
@@ -546,6 +546,64 @@ console.log(stdout);
 export function execaCommand(command: string, options?: Options): ExecaChildProcess;
 export function execaCommand(command: string, options?: Options<null>): ExecaChildProcess<Buffer>;
 
+type Expression = string | number | Array<string | number>;
+
+/**
+Same as `execa()` except both file and arguments are specified in a single `command` string. For example, `execa('echo', ['unicorns'])` is the same as $\`echo unicorns\`.
+
+If the file or an argument contains spaces, they must be escaped with backslashes. This matters especially if `command` is not a constant but a variable, for example with `__dirname` or `process.cwd()`. Except for spaces, no escaping/quoting is needed.
```
aaronccasanova, line 554: Note: I'll update this JSDoc once we iron out details on escaping

lib/command.js
```diff
@@ -39,3 +39,22 @@ export const parseCommand = command => {
 
 	return tokens;
 };
+
+export const parseTemplate = (template, index, templates, expressions) => {
```
ehmicky, line 43:
```suggestion
const parseTemplate = (template, index, templates, expressions) => {
```

index.js
```diff
@@ -224,6 +224,30 @@ export function execaSync(file, args, options) {
...
+export function $(templatesOrOptions, ...expressions) {
+	if (Array.isArray(templatesOrOptions)) {
+		const [file, ...args] = parseTemplates(templatesOrOptions, expressions);
+		return execa(file, args);
+	}
+
+	return (templates, ...expressions) => {
+		const [file, ...args] = parseTemplates(templates, expressions);
+		return execa(file, args, templatesOrOptions);
+	};
```
ehmicky, line 236: Should we add a way to bind options multiple times? With each call being shallowly merged to the previous one. This could be useful when someone wants to set some global options once, but then add more options for specific calls. 

Also, when users might be binding options before knowing whether the call will be sync or not. Or they might bind options for multiple calls, sync and async.

For example:

```js
// Global options
$ = $({ env: { ... } })

// ...

// Additional options for a specific call
const result = await $({ shell: true })`ls *`
// And also a sync call
await $.sync({ cwd: '...' })`rm foo`
```

What are your thoughts on this?

aaronccasanova: While this behavior could be well documented, my initial thoughts are that this could also be confusing for users and difficult to provide accurate type definitions.

A similar experience, albeit less elegant, might be encouraging the following workflows:

1. Passing around a global `execaOptions` object
```js
// src/constants.js
export const execaOptions = {...}

// src/script.js
import {execaOptions} from './constants.js'

await $({...execaOptions, shell: true})`ls *`
```

2. Create a project specific wrapper for the `$` API
```js
// src/$.js
import {$ as base$} from 'execa';

const execaOptions = {...}

export const $ = (templatesOrOptions, ...expressions) => {
  if (Array.isArray(templatesOrOptions)) {
    return base$(execaOptions)(templateOrOptions, ...expressions);
  }

  return (templates, ...expressions) => {
    return base$({...execaOptions, ...templatesOrOptions})(templates, ...expressions);
  };
}

// src/script.js
import {$} from './$.js'

await $({shell: true})`ls *`
```

> Note: We could also expose the above recipe as a `create$(baseOptions)` utility

In both cases, I think the code might be easier for users to understand and navigate within their projects. Thoughts?

ehmicky: I think the second option might be too much work for users. However, the first option is a possibility.

Yet another possibility would be to use an imperative approach instead of functional, i.e. expose a global mutable `$.options` object that users can modify. I personally dislike that kind of approach because it unnecessarily introduces global state. However, it might be easy to understand and avoid passing option objects around.

I'd like to get @sindresorhus input on this before moving forward. @sindresorhus What do you think?

Summary: how to pass options to `$` tagged template?
  1. `$(options)` binds options. For global options, it can be called several times.
  2. `$(options)` binds options, but can be called only once. For global options, users are expected to pass options objects around.
  3. `$.options` global object.

ehmicky: @sindresorhus What do you think?

sindresorhus: It should definitely be immutable. Options should be passed in normally. Would also be nice to provide an easy API to create a new `$` with custom options (as shown with the "Above example with create$ API" suggestion above). So `2` with a convenience API to create new instances bound to some options.

ehmicky: This sounds good, and take the best of both approaches.
What about renaming the following?

```js
// Global options
$ = $({ env: { ... } })
```

To:

```js
// Global options
$ = $.create({ env: { ... } })
```

This removes the need for two different methods `$(options)` and `$.create(options)` which would be quite similar to each other. Like this, users only need to learn a single method `$.create(options)`.

sindresorhus: Agreed

ehmicky: @aaronccasanova What do you think?

aaronccasanova: I like it! I'll do a quick experiment with the `$.create` API and let ya'll know if I hit any blockers (already anticipating some TypeScript complexities).

Some initial thoughts, concerns, clarifications:

**concern:** The examples thus far have been re-assigning the `$` API (e.g. `$ = $.create(...)`). However, from my understanding, this won't really be possible without renaming the imported `$`:
<details>
  <summary>Can't re-assign constant variable <code>$</code></summary>

Screen Shot 2023-02-12 at 10 47 06 AM
</details>
<details>
  <summary>Can't import <code>$</code> variable with same name as local identifier</summary>

Screen Shot 2023-02-12 at 11 01 43 AM
</details>

Perhaps, this is not an issue, but a bit clunky IMO

Note: The `create$` API doesn't have this "issue", but removes the ability to recursively assign and merge options e.g. `$ = $.create(...) = $.create(...) = $.create(...)`

**clarification:** Just want to make sure my previous example is inline with the intention of the API (e.g. `$.create` can be recursively re-assigned and options are shallow merged). I assume so from @sindresorhus's comment: `So each call can extend the existing defaults. ``.create`` can make new defaults be merging with existing ones.`

ehmicky: Good catch, you're correct.
So users would need to use a different name and variable than `$` for the return value of `$.create()`. I am actually thinking this might be a good thing? Like this, the original `$` is immutable, and options-bound "copies" of `$` have their own distinct variables, which might create less confusion for users.

```js
import { $ } from 'execa'

const $$ = $.create({ env: { ... } })
await $$({ shell: true })`ls *`
await $$.async({ cwd: '...' })`rm foo`
```

Or:

```js
import { $ as $$ } from 'execa'

const $ = $$.create({ env: { ... } })
await $({ shell: true })`ls *`
await $.async({ cwd: '...' })`rm foo`
```

sindresorhus: > I am actually thinking this might be a good thing? Like this, the original $ is immutable, and options-bound "copies" of $ have their own distinct variables, which might create less confusion for users.

👍

aaronccasanova: Sounds good 👍 Okay, here's my initial pass at the `$.create` API: https://github.com/sindresorhus/execa/pull/510/commits/5fac1b7ecdafa4b871a81aaeb191cf537116c4ca

I'll pause here for feedback before moving on to types, tests, and docs!

aaronccasanova: While addressing the eslint errors, I noticed and included refactor that removed all the duplicate conditional logic: https://github.com/sindresorhus/execa/pull/510/commits/1bd04ec27d5a055671e7bd8e0983dcde6fd0939e

ehmicky: Thanks @aaronccasanova,

The current implementation allows the following two syntaxes:

```js
await $({ shell: true })`ls *`
await $.create({ shell: true })`ls *`
```

I suggested this above, but now I am wondering: those two are actually functionally equivalent, so would it make sense to merge `$(...)` and `$.create(...)` into a single method?

aaronccasanova: I considered this as well, but still liked the idea of a dedicated method for binding options. To play devils advocate, if we merged `$(...)` and `$.create(...)` are there any concerns around creating extraneous functions and memory? Here's a super contrived example:

```js
const $$ = $({ shell: true })

for (const command of commands) {
  await $$(command.options)(...command.args)
}
```

In this example, each `$$(...)` would create a brand new `$` and `$.sync` function. As opposed to calling the same reference returned by `$.create(...)`.

ehmicky: When it comes to memory consumption, I think the memory and CPU time wasted by creating anonymous function is quite negligible compared to spawning the child process. For example, CPU time is in the milliseconds-range for child process spawning, while creating anonymous function is in the nanoseconds-range. Memory consumption is also along those lines.

I think having a dedicated method for options binding is nice too, but I am wondering whether there is a technical difference between: 
  - Binding options early (e.g. global options), then spawn multiple processes with it
  - Spawning a single process with specific options

Conceptually, those two could be done with a single method like `$(options)`. Having a single method instead of two could simplify the API for users, and also its implementation. :thinking:

ehmicky: Great idea to move `$.sync(...)` to `$(...).sync`! I am wondering how @sindresorhus feels about this, but this look good to me!

aaronccasanova: Okay, here are the two APIs/commits to compare:
- https://github.com/sindresorhus/execa/pull/510/commits/1bd04ec27d5a055671e7bd8e0983dcde6fd0939e: Implements `$.create` for binding options to both the async `$` and `$.sync` APIs and allows one off override options per call e.g. `$(overrideOptions)` and `$.sync(overrideOptions)`
- https://github.com/sindresorhus/execa/pull/510/commits/50055a6a521b5edb8cd805b907e13851a97dbc77: Consolidates option binding and merging to a single method e.g. `$(options)` and `$(options).sync`

sindresorhus: > Having a single method instead of two could simplify the API for users, and also its documentation, implementation, typing, tests, etc. 

:+1:

> Great idea to move $.sync(...) to $(...).sync! I am wondering how @sindresorhus feels about this, but this look good to me!

:+1:

ehmicky: The approach in the current state of the PR (50055a6) looks good to me, thanks again for working on this! :+1:

test/command.js
```diff
@@ -85,3 +85,88 @@ test('execaCommandSync()', t => {
...
 	t.is(stdout, 'foo\nbar');
 });
+
+test('$', async t => {
+	const {stdout} = await $`node test/fixtures/echo.js foo bar`;
+	t.is(stdout, 'foo\nbar');
+});
+
+test('$ accepts options', async t => {
+	const {stdout} = await $({encoding: 'utf8'})`node test/fixtures/echo.js foo bar`;
```
ehmicky, line 95: We might want to use an example where the options changes `stdout`. The `encoding` does not change `stdout` here, so the test would pass even if the `encoding` option was ignored.

playground.js
```diff
@@ -0,0 +1,12 @@
...
+let $$ = $({stdio: 'inherit'});
+
+await $$({shell: true})`ls *.ts`;
+
+await $$`echo "$TEST"`;
+
+$$ = $$({shell: true, env: {TEST: 'test'}});
+
+// eslint-disable-next-line no-unused-expressions
+$$.sync`echo "$TEST"`;
```
ehmicky, line 12: I am wondering whether you intended to commit this file, or whether it was meant for debugging instead?

index.d.ts
```diff
@@ -546,6 +546,64 @@ console.log(stdout);
...
+//=> 'unicorns'
+
+const my$ = $({stdio: 'inherit'});
+await my$`echo unicorns`;
+//=> 'unicorns'
+await my$`echo rainbows`;
+//=> 'rainbows'
+```
+*/
+export const $: {
```
ehmicky, line 582: We should add some type tests as well.

index.d.ts
```diff
@@ -546,6 +546,64 @@ console.log(stdout);
 export function execaCommand(command: string, options?: Options): ExecaChildProcess;
 export function execaCommand(command: string, options?: Options<null>): ExecaChildProcess<Buffer>;
 
+type Expression = string | number | Array<string | number>;
```
ehmicky, line 549: Small nit-picking: what do you think about naming this type more specifically. `Expression` is rather generic, and I did not realize at first it was meant only for the template strings expressions under `${...}`. Maybe `TemplateExpression`, or anything else more specific?

playground.js
aaronccasanova: Thanks for flagging. Yeah, this file was temporarily committed for debugging and exploring the different options binding APIs we discussed above. Removed: https://github.com/sindresorhus/execa/pull/510/commits/3f6de3d730bbbe3ce54a426f6bef8213d74cfea3

index.d.ts
aaronccasanova: Good call! `TemplateExpression` is 👌 https://github.com/sindresorhus/execa/pull/510/commits/6ed5b89f7d37f15e707bed0239788ef2555e9f45

index.d.ts
```diff
@@ -546,6 +546,64 @@ console.log(stdout);
...
+		: (templates: TemplateStringsArray, ...expressions: Expression[]) => T extends Options<null>
+			? ExecaChildProcess<Buffer>
+			: ExecaChildProcess;
+
+	/**
+	Same as `$` but synchronous.
+
+	@returns A result `Object` with `stdout` and `stderr` properties.
+	*/
+	sync<T extends TemplateStringsArray | Options<string | null>>(
```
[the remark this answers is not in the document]
aaronccasanova: It's definitely worth exploring 👍 I just pushed up some [initial type updates](https://github.com/sindresorhus/execa/pull/510/commits/e1365af02748a49aa48e0d323e9c14fcc2c1fcfd), matching the new options binding API (using generics), so we at least have a working solution. However, I can circle back to see if function overloads is an option when I start fleshing out the `index.test-d.ts` file.

test/command.js
aaronccasanova: Nice catch! For now, I just updated the tests to [properly validate the APIs accept options](https://github.com/sindresorhus/execa/pull/510/commits/0cc0fc3aa6072763252ebd664f7abfefe5e6a023), but agree we should include tests to validate different `stdout`s 👍

package.json
```diff
@@ -61,6 +61,7 @@
 		"path-key": "^4.0.0",
 		"tempfile": "^4.0.0",
 		"tsd": "^0.19.1",
+		"type-fest": "^3.5.7",
```
ehmicky, line 64: I believe this should be a production dependency? Otherwise, TypeScript users that do `npm install execa` would receive a type error due to the missing dependency.

`type-fest` is rather big, so I would be wondering how @sindresorhus feels about introducing that dependency :thinking:

test/command.js
```diff
@@ -85,3 +85,88 @@ test('execaCommandSync()', t => {
...
 	t.is(stdout, 'foo\nbar');
 });
+
+test('$', async t => {
+	const {stdout} = await $`node test/fixtures/echo.js foo bar`;
+	t.is(stdout, 'foo\nbar');
+});
+
+test('$ accepts options', async t => {
+	const {stdout} = await $({shell: true, env: {TEST: 'test'}})`echo $TEST`;
```
ehmicky, line 95: I believe this might not work on `cmd.exe`? This is the shell used by `shell: true` on Windows, and environment variables are using [a different syntax `%TEST%`](https://github.com/ehmicky/cross-platform-node-guide/blob/main/docs/4_terminal/environment_variables.md#referencing).

What about using the `stripFinalNewline` option instead, and checking to see if the newline is stripped?

aaronccasanova: Another great catch! Updated the tests to use the `stripFinalNewline` option: https://github.com/sindresorhus/execa/pull/510/commits/bddf3179e26a50aff59c9eec72d703bf55d9eb3a

index.d.ts
aaronccasanova: Added! https://github.com/sindresorhus/execa/pull/510/commits/a287fc97c2c60ada452bc1b0b3755e596623bfac

Note: I just copied the `execaCommand` and `execaCommandSync` tests and swapped out implementations with the respective `$` APIs. Thoughts?

package.json
aaronccasanova: I have since removed `type-fest` with the refactor to use type overloads: https://github.com/sindresorhus/execa/pull/510/commits/6a283cb0dd156c4060b3935f253be758dfceba98

index.d.ts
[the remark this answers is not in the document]
aaronccasanova: Wow, okay so using overloads drastically simplified the types imo. Thoughs? https://github.com/sindresorhus/execa/pull/510/commits/6a283cb0dd156c4060b3935f253be758dfceba98

index.d.ts
ehmicky: This makes sense. Thanks!

index.d.ts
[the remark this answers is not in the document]
ehmicky: Yes, it did! Overloads can be quite nice in situations like this. :)

index.test-d.ts
```diff
@@ -219,3 +220,16 @@ expectType<ExecaChildProcess<Buffer>>(
...
+expectType<ExecaReturnValue>(await $({encoding: 'utf8'})`unicorns`);
+expectType<ExecaReturnValue<Buffer>>(await $({encoding: null})`unicorns`);
+expectType<ExecaReturnValue>(await $({encoding: 'utf8'})`unicorns foo`);
+expectType<ExecaReturnValue<Buffer>>(await $({encoding: null})`unicorns foo`);
+
+expectType<ExecaSyncReturnValue>($.sync`unicorns`);
+expectType<ExecaSyncReturnValue>($({encoding: 'utf8'}).sync`unicorns`);
+expectType<ExecaSyncReturnValue<Buffer>>($({encoding: null}).sync`unicorns`);
+expectType<ExecaSyncReturnValue>($({encoding: 'utf8'}).sync`unicorns foo`);
+expectType<ExecaSyncReturnValue<Buffer>>($({encoding: null}).sync`unicorns foo`);
```
ehmicky, line 235: We should add some tests with `${...}` to test the type of `TemplateExpression`. With `string`, `number`, `string[]` and `number[]`, and also with another type (like `boolean`), where it should then fail. What do you think?

index.d.ts
```diff
@@ -541,6 +541,27 @@ console.log(stdout);
 export function execaCommand(command: string, options?: Options): ExecaChildProcess;
 export function execaCommand(command: string, options?: Options<null>): ExecaChildProcess<Buffer>;
 
+type TemplateExpression = string | number | Array<string | number>;
+
+type Execa$<EncodingType = string> = {
+	(options: Options): Execa$;
```
ehmicky, line 547:
```suggestion
	(options: Options): Execa$<EncodingType>;
```

We should also add some type tests for it, like:

```ts
expectType<ExecaReturnValue<Buffer>>(await $({encoding: null})({})`unicorns foo`);
expectType<ExecaReturnValue>(await $({encoding: null})({encoding: 'utf8'})`unicorns foo`);
```

index.d.ts
```diff
@@ -541,6 +541,27 @@ console.log(stdout);
...
+	): EncodingType extends null
+		? ExecaChildProcess<Buffer>
+		: ExecaChildProcess;
+	sync(
+		templates: TemplateStringsArray,
+		...expressions: TemplateExpression[]
+	): EncodingType extends null
+		? ExecaSyncReturnValue<Buffer>
+		: ExecaSyncReturnValue;
+};
```
ehmicky, line 561: I believe the conditionals might be simplified by using directly:
 
```ts
type Execa$<StdoutErrorType = string> = {
	(options: Options): Execa$<StdoutErrorType>;
	(options: Options<null>): Execa$<Buffer>;
	(
		templates: TemplateStringsArray,
		...expressions: TemplateExpression[]
	): ExecaChildProcess<StdoutErrorType>;
	sync(
		templates: TemplateStringsArray,
		...expressions: TemplateExpression[]
	): ExecaSyncReturnValue<StdoutErrorType>;
};
```

What do you think?

aaronccasanova: It just keeps getting better with each iteration! I'll update to this and add the above mentioned test cases 👌

aaronccasanova: Actually, I'm seeing a mix of `StdoutStderrorType` and `StdoutErrorType` arguments thorought the file. I have a slight preference for the more explicit `StdoutStderrorType` variant. Mind if I standardize all references?

ehmicky: Yes, absolutely, good catch :+1:

aaronccasanova: I had to make one small adjustment, but your suggestions worked great! https://github.com/sindresorhus/execa/pull/510/commits/989f6bc1fa5f007106a9cc5a24dfd1e0e27ac033

> Yes, absolutely, good catch 👍

Done https://github.com/sindresorhus/execa/pull/510/commits/f40759b2cfeb6eeaf9fb5e7e302a655e3bb5c8fe

index.d.ts
aaronccasanova: Included the above suggestion here: https://github.com/sindresorhus/execa/pull/510#discussion_r1109290963

aaronccasanova: Included the above suggestion here: https://github.com/sindresorhus/execa/pull/510#discussion_r1109290963

index.test-d.ts
aaronccasanova: I'm not super familiar with `tsd`. How does this look? https://github.com/sindresorhus/execa/pull/510/commits/656a69b05c9bae4330503f295a8d6d8386ce3fa2

aaronccasanova:
Thanks for the incredibly thorough review and assistance with the implementation @ehmicky! I too am very happy with the API we've landed on and believe I have addressed all the feedback (with the exception of updating the `readme.md`).

Regarding the `readme.md`, what are we thinking here? Should we add a collection of [usage examples here](https://github.com/sindresorhus/execa#usage) or just add a `$` section to the [API docs](https://github.com/sindresorhus/execa#api)?

index.test-d.ts
ehmicky: That's exactly it. Thank!

ehmicky:
Personally, I would do both. This is quite a big feature, so it'd deserve some higher visibility by showing a high-level usage example. And then, the API section would get into more details. What do you think?

aaronccasanova:
Sounds good 👍 I kicked off the process by copy/pasting the examples in the PR description, adjusted them to match other examples, and fixed up a few typos in the JSDoc types: https://github.com/sindresorhus/execa/pull/510/commits/f8fdd69ef2883b55c946066a91bebaee09029344

From here, I may need some assistance moving things around and dialing in the right balance between usage examples and written documentation.

readme.md
```diff
@@ -191,6 +240,18 @@ Execute a file synchronously.
 
 Returns or throws a [`childProcessResult`](#childProcessResult).
 
+### \$(options?)\`command\`
+
+Same as `execa()` except both file and arguments are specified in a single tagged template string. For example, ``$`echo unicorns` ``is the same as `execa('echo', ['unicorns'])`.
```
ehmicky, line 245:
```suggestion
Same as [`execa()`](#execafile-arguments-options) except both file and arguments are specified in a single tagged template string. For example, ``$`echo unicorns` ``is the same as `execa('echo', ['unicorns'])`.
```

ehmicky, line 245: We should have some information that any string or array of strings (not sure whether number arguments need be documented?) in `${...}` is automatically escaped.

readme.md
```diff
@@ -191,6 +240,18 @@ Execute a file synchronously.
...
+
+The [`shell` option](#shell) must be used if the `command` uses shell-specific features (for example, `&&` or `||`), as opposed to being a simple `file` followed by its `arguments`.
```
ehmicky, line 247: We might want to add that quotes, backslashes and spaces are automatically escaped and have no special meaning unless the `shell` option is used. I can see this becoming a common mistake. About half of the issues in this repository are related to confusion about shell syntax and escaping.

readme.md
```diff
@@ -191,6 +240,18 @@ Execute a file synchronously.
...
+
+### $(options?).sync\`command\`
+
+Same as [$(options?)\`command\`](#optionscommand) but synchronous.
+
+Returns or throws a [`childProcessResult`](#childProcessResult).
```
ehmicky, line 253: The async method should include a similar line IMHO. This is implied by "Same as `execa()`". However, it might be good to be explicit about the return value. What do you think?

readme.md
```diff
@@ -191,6 +240,18 @@ Execute a file synchronously.
...
 
+### \$(options?)\`command\`
+
+Same as `execa()` except both file and arguments are specified in a single tagged template string. For example, ``$`echo unicorns` ``is the same as `execa('echo', ['unicorns'])`.
+
+The [`shell` option](#shell) must be used if the `command` uses shell-specific features (for example, `&&` or `||`), as opposed to being a simple `file` followed by its `arguments`.
+
+### $(options?).sync\`command\`
+
+Same as [$(options?)\`command\`](#optionscommand) but synchronous.
```
ehmicky, line 251:
```suggestion
Same as [$(options?)\`command\`](#optionscommand) but synchronous like [`execaSync()`](#execasyncfile-arguments-options).
```

readme.md
```diff
@@ -39,6 +39,55 @@ console.log(stdout);
...
+import {$} from 'execa';
+
+const {stdout} = $.sync`echo unicorns`;
+console.log(stdout);
+//=> 'unicorns'
+
+$({stdio: 'inherit'}).sync`echo rainbows`;
+//=> 'rainbows'
+```
+
```
ehmicky, line 90: Those examples are looking good to me, thanks! :tada:

index.d.ts
```diff
@@ -541,8 +541,80 @@ console.log(stdout);
...
+//=> 'unicorns'
+await $$`echo rainbows`;
+//=> 'rainbows'
+```
+*/
+export const $: Execa$;
+
 /**
-Same as `execaCommand()` but synchronous.
+Same as `$` but synchronous.
```
ehmicky, line 617: This documents `execaCommandSync()`, so comparing with `execaCommand()` was correct?

aaronccasanova: Whoops, that was unintentional: https://github.com/sindresorhus/execa/pull/510/commits/97bdf47bc93067322e51f01b2df7bd6758397b53

readme.md
```diff
@@ -191,6 +242,24 @@ Execute a file synchronously.
...
 
+### \$(options?)\`command\`
+
+Same as [`execa()`](#execafile-arguments-options) except both file and arguments are specified in a single tagged template string. For example, `` $`echo unicorns` `` is the same as `execa('echo', ['unicorns'])`.
+
+The [`shell` option](#shell) must be used if the `command` uses shell-specific features (for example, `&&` or `||`), as opposed to being a simple `file` followed by its `arguments`.
+
+The return value of this function depends on whether you are binding options or executing a command:
+- If you are binding options, for example with `const $$ = $(options)`, the return value is a new `$` instance with the options bound to both the asynchronous `$` and synchronous `$.sync` APIs. It's important to note that consecutive calls to this API will shallow merge the options.
+- If you are executing a command, the result is a `Promise` that resolves or rejects with a [`childProcessResult`](#childProcessResult).
```
ehmicky, line 253: I am wondering whether we should split this section into two?
  - $\`command\`, which executes a command
  - `$(options)`, which returns a new `$` with specific default `options`

This might also make it clearer that users do not need to call `$(options)` to execute a command.

aaronccasanova: Good call 👍 Here's my first pass at breaking the `$`, `$.sync`, and `$(options)` APIs into dedicated sections: https://github.com/sindresorhus/execa/pull/510/commits/33f4d065a0d02172ae23e658ce05126d7a679e89

readme.md
aaronccasanova: Addressed here: https://github.com/sindresorhus/execa/pull/510#discussion_r1111455947

index.d.ts
```diff
@@ -541,6 +541,78 @@ console.log(stdout);
 export function execaCommand(command: string, options?: Options): ExecaChildProcess;
 export function execaCommand(command: string, options?: Options<null>): ExecaChildProcess<Buffer>;
 
+type TemplateExpression = string | number | Array<string | number>;
+
+type Execa$<StdoutStderrType = string> = {
+	/**
+  Same as `execa()` except both file and arguments are specified in a single tagged template string. For example, `` $`echo unicorns` `` is the same as `execa('echo', ['unicorns'])`.
```
sindresorhus, line 548: Incorrect indentation.

index.d.ts
```diff
@@ -541,6 +541,85 @@ console.log(stdout);
...
+	(options: Options<null>): Execa$<Buffer>;
+	(
+		templates: TemplateStringsArray,
+		...expressions: TemplateExpression[]
+	): ExecaChildProcess<StdoutStderrType>;
+
+	/**
+	Same as `$` but synchronous.
+
+	@returns A result `Object` with `stdout` and `stderr` properties.
```
sindresorhus, line 585: The types already describe this. It could instead say:

```suggestion
	@returns The stdout and stderr output.
```

readme.md
```diff
@@ -191,6 +242,28 @@ Execute a file synchronously.
...
+
+### $.sync\`command\`
+
+Same as [$\`command\`](#command) but synchronous like [`execaSync()`](#execasyncfile-arguments-options).
+
+Returns or throws a [`childProcessResult`](#childProcessResult).
+
+### $(options)
+
+Binds options to the [`$`](#command) API. For example, you can use `$(options)` to create a new `$` instance with specific default options, which are then bound to both the asynchronous `` $`command` `` and synchronous `` $.sync`command` `` APIs.
```
ehmicky, line 263:
```suggestion
Binds options to the [`$`](#command) API. For example, you can use `$(options)` to create a new `$` instance with specific default options, which are then bound to both the asynchronous [`` $`command` ``](#command) and synchronous [`` $.sync`command` ``](#synccommand) APIs.
```

index.d.ts
```diff
@@ -541,6 +541,87 @@ console.log(stdout);
...
+	(options: Options<undefined>): Execa$<StdoutStderrType>;
+	(options: Options): Execa$;
+	(options: Options<null>): Execa$<Buffer>;
+	(
+		templates: TemplateStringsArray,
+		...expressions: TemplateExpression[]
+	): ExecaChildProcess<StdoutStderrType>;
+
+	/**
+	Same as `$` but synchronous.
```
ehmicky, line 584: The description of both `$` and `$.sync` in the types documentation should be a copy of the one in the `readme.md`.

index.d.ts
aaronccasanova: Fixed https://github.com/sindresorhus/execa/pull/510/commits/38378d01e16a4258d34728a2d1cd1e96eaaf95cf

index.d.ts
aaronccasanova: Done https://github.com/sindresorhus/execa/pull/510/commits/9e566e582e065529673b073c2dc2ed7e7e1e5741

readme.md
```diff
@@ -39,6 +39,57 @@ console.log(stdout);
...
+```js
+import {$} from 'execa';
+
+const {stdout} = $.sync`echo unicorns`;
+console.log(stdout);
+//=> 'unicorns'
+
+$({stdio: 'inherit'}).sync`echo rainbows`;
+//=> 'rainbows'
+```
```
sindresorhus, line 91: The examples here do not exactly match the ones in the types. I think we should try to make them as similar as possible to ease maintenance.

https://github.com/sindresorhus/execa/pull/510/files#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R608-R625

aaronccasanova: Good call 👍 I think the examples provided in the readme are comprehensive and demonstrate the full API, so I copied those over to the JSDoc types: https://github.com/sindresorhus/execa/pull/510/commits/5bb310ec9a9065265be381eeeab399f2f54ff995

sindresorhus:
Looks great! Thanks so much for working on this, @aaronccasanova 🙏

aaronccasanova:
Of course! Thanks for all your feedback and assistance with the API design, @sindresorhus! And a big shoutout to @ehmicky for the thorough review and support throughout the process 🙏

sindresorhus:
Just FYI. We will hold off from releasing this for another week or so as @ehmicky is writing a blog post about it.

Issue #517, opened 2023-02-10
Output streams do not emit "end" event if execution fails


anomiex:
Consider this code (using node 18.13.0 and execa 6.1.0)
```js
import { execa } from 'execa';

const proc = execa( "does not exist", [], { stdio: [ 'ignore', 'pipe', 'inherit' ], buffer: false } );

proc.stdout.on( 'end', () => console.log( 'stdout ended' ) );
proc.stdout.on( 'close', () => console.log( 'stdout closed' ) );

await proc.catch( e => e );
```
Only "stdout closed" is logged, "stdout ended" never happens.

This does not match the behavior of `child_process`:
```js
import { spawn } from 'node:child_process';

const proc = spawn( "does not exist", [], { stdio: [ 'ignore', 'pipe', 'inherit' ] } );

proc.stdout.on( 'end', () => console.log( 'stdout ended' ) );
proc.stdout.on( 'close', () => console.log( 'stdout closed' ) );

proc.on( 'error', () => console.log( 'proc error' ) );
```
In this case both "stdout ended" and "stdout closed" fire as expected.

anomiex:
FYI, the code I'm trying to make work looks a bit more like this than the examples above:
```js
import { execa } from 'execa';
import readline from 'readline';

async function* listFiles( dir ) {
    const proc = execa( 'git', [ 'ls-files' ], {
        cwd: dir,
        stdio: [ 'ignore', 'pipe', null ],
        buffer: false
    } );

    const rl = readline.createInterface( {
        input: proc.stdout,
        crlfDelay: Infinity,
    } );

    // We can't allow proc to reject, see https://github.com/sindresorhus/execa/issues/516#issuecomment-1426154012
    // So catch the error and save it to rethrow later.
    let err;
    proc.catch( e => { err = e; } );

    // Yield each line of output.
    yield* rl;

    // If proc rejects, throw the error now.
    await proc;
    if ( err ) throw err;
}
```
The failure case is if `git` somehow does not exist in the path.

If you have any better ways to handle this code that don't result in reading the whole output of the command into memory, I'd be happy for the suggestion. The rearrangement suggested by @ehmicky in https://github.com/sindresorhus/execa/issues/516#issuecomment-1425018125 doesn't seem compatible with use of `yield`.

ehmicky:
Hi @anomiex,

Good catch! I have opened #518 to fix this problem when `buffer` is `false`.

mrienstra:
[...]

PR #518, opened 2023-02-11
Emit `end` event on streams when process fails

ehmicky:
Fixes https://github.com/sindresorhus/execa/issues/517

When a process (or one of its standard streams) errors, we return what could have been buffered so far from stdout/stderr as `error.stdout` and `error.stderr` (which are also appended to `error.message`). To do this, we first `destroy()` those streams, then retrieve the buffered value from `get-stream`.

However, calling `destroy()` prevents the `end` event from being emitted on those streams. This behavior differs from how `child_process.spawn()` and might be unexpected.

This PR fixes this for the case where the `buffer` option is `false`. When that's the case, there is no need to destroy the streams since we do not need to do anything with them anymore. We can just let Node.js destroy them naturally as part of the child process exit.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -36,7 +36,7 @@ export const makeAllStream = (spawned, {all}) => {
 
 // On failure, `result.stdout|stderr|all` should contain the currently buffered stream
 const getBufferedData = async (stream, streamPromise) => {
-	if (!stream) {
+	if (!stream || streamPromise === undefined) {
 		return;
 	}
 
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -36,6 +36,7 @@ export const makeAllStream = (spawned, {all}) => {
 
 // On failure, `result.stdout|stderr|all` should contain the currently buffered stream
 const getBufferedData = async (stream, streamPromise) => {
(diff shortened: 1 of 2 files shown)


test/stream.js
```diff
@@ -173,6 +174,11 @@ test('buffer: false > promise rejects when process returns non-zero', async t =>
 	t.is(exitCode, 2);
 });
 
+test('buffer: false > emits end event when promise is rejected', async t => {
+	const subprocess = execa('wrong command', {buffer: false, reject: false});
+	await t.notThrowsAsync(Promise.all([subprocess, pEvent(subprocess.stdout, 'end')]));
```
ehmicky, line 179: Before the code fix above, the `end` event would never be emitted on `subprocess.stdout` in that test case.

lib/stream.js
```diff
@@ -36,7 +36,7 @@ export const makeAllStream = (spawned, {all}) => {
 
 // On failure, `result.stdout|stderr|all` should contain the currently buffered stream
 const getBufferedData = async (stream, streamPromise) => {
-	if (!stream) {
+	if (!stream || streamPromise === undefined) {
```
sindresorhus, line 39: A short code comment here would be great.

ehmicky: Done :+1:

PR #528, opened 2023-03-04
Allow `execa` return values in `$` template expressions

### LINKED ISSUE #523: `${childProcessResult}` shortcut notation
[zx](https://github.com/google/zx/) has a `${childProcessResult}` shortcut notation:

```js
const branch = await $`git branch --show-current`
await $`dep deploy --branch=${branch}`
```

This is equivalent to:

```js
const branch = await $`git branch --show-current`
await $`dep deploy --branch=${branch.stdout.replace(/\n$/, '')}`
```

This is quite handy and would be minimal to implement for us as part of the new `$` API. 
What do you think @sindresorhus @aaronccasanova?

aaronccasanova:
Closes #523 

This PR adds support for interpolating `ExecaReturnValue` and `ExecaSyncReturnValue` in `$` template expressions:

#### Asynchronous

```js
import {$} from 'execa';

const rainbows = await $`echo rainbows`;

await $({stdio: 'inherit'})`echo ${rainbows} unicorns`;
//=> 'rainbows unicorns'
```

#### Synchronous

```js
import {$} from 'execa';

const rainbows = $.sync`echo rainbows`;

$({stdio: 'inherit'}).sync`echo ${rainbows} unicorns`;
//=> 'rainbows unicorns'
```


aaronccasanova:
@sindresorhus @ehmicky I'll update the JSDoc types and README documentation after sign off on the implementation.

index.test-d.ts
```diff
@@ -265,6 +265,12 @@ expectType<ExecaReturnValue>(await $`unicorns ${['foo', 'bar']}`);
 expectType<ExecaSyncReturnValue>($.sync`unicorns ${['foo', 'bar']}`);
 expectType<ExecaReturnValue>(await $`unicorns ${[1, 2]}`);
 expectType<ExecaSyncReturnValue>($.sync`unicorns ${[1, 2]}`);
+expectType<ExecaReturnValue>(await $`unicorns ${await $`echo foo`}`);
+expectError<ExecaReturnValue>(await $`unicorns ${$`echo foo`}`);
```
ehmicky, line 269: This specific line might have an issue: ``$`echo foo` `` returns a promise, not an `ExecaReturnValue`, so that should not work?

lib/command.js
```diff
@@ -40,6 +42,16 @@ export const parseCommand = command => {
 	return tokens;
 };
 
+const parseExpression = expression => {
+	if (typeof expression === 'object' && expression !== null && 'stdout' in expression) {
```
ehmicky, line 46:
```suggestion
	if (typeof expression === 'object' && expression !== null && expression.stdout !== undefined) {
```

I could be wrong, but I believe when `options.stdout` is `inherit`, `expression` would have a `stdout` key, but its value would be `undefined` (as opposed to having no `stdout` key).

lib/command.js
```diff
@@ -40,6 +42,16 @@ export const parseCommand = command => {
...
+		return Buffer.isBuffer(expression.stdout)
+			? expression.stdout.toString()
+			: String(expression.stdout);
```
ehmicky, line 49:
```suggestion
			: expression.stdout;
```

If `stdout` is not `undefined` nor a buffer, it should be guaranteed to be a string.

index.test-d.ts
aaronccasanova: Have the test marked as `expectError`. Should I be handling this case in `parseExpression`? e.g.
```js
if (expression instanceof Promise) throw new TypeError('Invalid expression type')
```

ehmicky: Oops, my bad, I did not notice it was using `expectError`!

ehmicky:
This is looking great @aaronccasanova, thanks! :tada:

lib/command.js
aaronccasanova: Here's my WIP regarding this and discussions [r1125573007](https://github.com/sindresorhus/execa/pull/528#discussion_r1125573007), [r1125572721](https://github.com/sindresorhus/execa/pull/528#discussion_r1125572721):
```js
const parseExpression = expression => {
	if (typeof expression === 'object' && expression !== null) {
		if ('stdout' in expression) {
			if (Buffer.isBuffer(expression.stdout)) {
				return expression.stdout.toString();
			}

			if (typeof expression.stdout === 'string') {
				return expression.stdout;
			}

			throw new TypeError(`Unexpected '${typeof expression.stdout}' stdout in template expression`);
		}

		if (expression instanceof Promise) {
			throw new TypeError('Unexpected promise in template expression');
		}

		throw new TypeError('Unexpected object in template expression');
	}

	return String(expression);
};
```

Although, I'm unsure if this is too exhaustive as we have been coercing all values thus far:
https://github.com/sindresorhus/execa/blob/eae327cc0a49f61b17f699b9199c7953e1497c1a/test/command.js#L144-L147

Or perhaps, we should lean into this, remove the coercion, and additionally validate the standalone `string` and `number` expressions (which would be more representative of the types). Thoughts?

ehmicky: Yes, I was also thinking of the fact that we previously just relied on the coercion instead of failing hard. :thinking: 

Personally, I tend to favor failing hard so that users realize their errors early. In this case, if a user forgets to call `await`, or if a user tries to use this feature with a process with `options.stdout: 'inherit'`, this is a mistake, so it would be helpful to report it so users can fix it. What do you think?

aaronccasanova: Agreed 👍  Here's my first pass at removing invalid template expression coercion and introducing error handling: https://github.com/sindresorhus/execa/pull/528/commits/03e7ab9a79c88e50c306a928a9e59e229d87c2df

test/command.js
```diff
@@ -165,9 +165,29 @@ test('$ handles invalid escape sequence', async t => {
 	t.is(stdout, '\\u');
 });
 
-test('$ coerces all interpolated expressions to strings', async t => {
-	const {stdout} = await $`node test/fixtures/echo.js ${0} ${undefined} ${null}`;
-	t.is(stdout, '0\nundefined\nnull');
+test('$ throws on invalid expression in interpolation', async t => {
```
ehmicky, line 168: We might want to break this into multiple tests.
[Ava macros](https://github.com/avajs/ava/blob/1c567642a31d4bc24c4421ba520b02e25ce7363a/docs/01-writing-tests.md#retrieving-test-metadata) can be used for this purpose too, like we do in https://github.com/sindresorhus/execa/blob/main/test/stdio.js

aaronccasanova: TIL! That cleaned up a ton of duplication and encouraged me to add a few more test cases: https://github.com/sindresorhus/execa/pull/528/commits/602fd8819a3dc425fe633c9405af98ac40e600e3

aaronccasanova: Here's a preview of the test titles:
Screenshot 2023-03-04 at 9 28 22 PM

ehmicky:
This looks good to me so far! :+1:

aaronccasanova:
Great, I'll start on the docs and open the PR shortly!

readme.md
```diff
@@ -90,6 +90,17 @@ $({stdio: 'inherit'}).sync`echo rainbows`;
 //=> 'rainbows'
 ```
 
+#### With results from `$` or `$.sync`
```
aaronccasanova, line 93: Open to suggestions here:
- `` Interpolating results from `$` or `$.sync` ``
- `` Interpolating results from previous `$` or `$.sync` ``
- etc.

ehmicky: This looks good to me :+1:

index.d.ts
```diff
@@ -551,6 +556,8 @@ type Execa$<StdoutStderrType = string> = {
 
 	The `shell` option must be used if the `command` uses shell-specific features (for example, `&&` or `||`), as opposed to being a simple `file` followed by its `arguments`.
 
+	As a convenience, the result from previous `` $`command` `` or `` $.sync`command` `` calls can be used as template expressions in subsequent commands. See the example below `` with results from `$` or `$.sync` `` for more details.
```
ehmicky, line 559: We might want to be explicit that `stdout` (not `stderr`) is being used.

ehmicky (approved):
:rocket:

index.d.ts
aaronccasanova: Good call 👍 How's this: https://github.com/sindresorhus/execa/pull/528/commits/bc64405c7e9f44d5b42571d4b699b0426c7897b1

sindresorhus:
Looks great 👍

PR #531, opened 2023-03-05
Add `pipeStdout()`-related methods

### LINKED ISSUE #525: Piping shortcut notation
## Problem

Redirecting a process's output to two different targets is a common pattern. For example, a user might want to `inherit` stdout/stderr for debugging reasons, while still using `pipe` to get a process's return value.

This is currently possible, but requires 3 statements, which trips up users who are less familiar with process execution. This required us to explicitly add those in our [Tips documentation](https://github.com/sindresorhus/execa#save-and-pipe-output-from-a-child-process).

```js
const subprocess = execa('echo', ['foo']);
subprocess.stdout.pipe(process.stdout);
const {stdout} = await subprocess;
```

This also makes chaining processes rather verbose. For example, we just got an issue #521 which was mostly related to that problem. The solution is not as simple as it could be.

```js
const ffmpegProcess = execa('ffmpeg', ['-ss', '5', '-i', 'input.mp4', '-t', '1', '-vf', 'cropdetect', '-f', 'null', '-'])
const awkProcess = execa('awk', ['/crop/ {print $NF}'])
const tailProcess = execa('tail', ['-1'])

ffmpegProcess.stderr.pipe(awkProcess.stdin)
awkProcess.stdout.pipe(tailProcess.stdin)

const { stdout } = await tailProcess
```

Having a simple syntax for chaining processes is especially important with the new `$` API because users might expect the convenience of the `|` shell syntax. For example, `zx` provides 2 ways of doing this:

```js
await $`cat package.json | grep name`
```

```js
await $`cat package.json`.pipe($`grep name`)
```

## Solution

This problem could be solved by introducing a shortcut notation. For example, the above examples above could be written as:

```js
const {stdout} = await execa('echo', ['foo']).pipeStdout(process.stdout);
```

```js
const {stdout} = await execa('ffmpeg', ['-ss', '5', '-i', 'input.mp4', '-t', '1', '-vf', 'cropdetect', '-f', 'null', '-'])
  .pipeStdout(execa('awk', ['/crop/ {print $NF}']))
  .pipeStdout(execa('tail', ['-1']))
```

```js
await $`cat package.json`.pipeStdout($`grep name`)
```

It would do the same thing, just in a shorter way, so it would be rather minimal to implement and document.

### Return value

When piping to a stream (like `process.stdout`), the return value would be the `childProcessResult`, not the stream, since users would most likely want to retrieve the process's result.

However, when piping to another `childProcessResult` (i.e. to its `stdin`), that other `childProcessResult` would be returned. This allows command chaining. Also, this mimics how piping works in shells. For example, in Bash, `$(cat package.json | grep name)` returns the result after `grep` has been applied. Overall, this is what users would expect.

### Stderr

`pipeStderr()` and `pipeAll()` could be available to redirect stderr, or stdout+stderr.

### Synchronous methods

Those methods only make sense with the async `execa()` methods, since they would otherwise be called once the process has already completed.

### Alternative solution

We could also implement this using the new `$` API with a special pipe symbol, like `zx` does.

```js
await $`cat package.json | grep name`
```

However, I think this solution might not be as good for the following reasons:
  - It does not work with normal `execa()` and `execaCommand()` calls
  - It introduces special tokens to the ``$`...` `` syntax. At the moment, the string argument only has command and arguments, with no shell-like syntax, which is much simpler to document and understand. It would also create some confusion for users expecting the command to be run in a shell by default (which is not the case) due to the usage of the pipe symbol.
  - It does not work well with functional utilities. For example, `pipeStdout()` can use `.bind()` or be assigned to a variable or parameter. Using a string makes this kind of manipulation harder.
  - It would require 3 different tokens to redirect to stdout, stderr or both.
  - Pipe characters would need a way to be escaped.

### Naming

If you have other naming ideas for `pipeStdout()`, `pipeStderr()` and `pipeAll()`, please let me know!
  - `stdout()`, etc. does not work because `childProcess.stdout` is already assigned as a stream
  - `stdoutTo()`, etc. could work, but users might understand the work "pipe" better
  - `pipe()` could work for `pipeStdout()`, but users might not understand it is a different method than `Stream.prototype.pipe()`.

--- 

ehmicky:
Fixes #525.

This adds the following shortcut methods: `pipeStdout()`, `pipeStderr()` and `pipeAll()`.

For Node.js shell-like scripts, this is the equivalent to the `|`, `|&`, `>`, `&>`, or `2>&1` shell characters.

Pipe to stream, before:

```js
const subprocess = $`echo foo`;
subprocess.stdout.pipe(process.stdout);
const {stdout} = await subprocess;
```

After:

```js
const {stdout} = await $`echo foo`.pipeStdout(process.stdout);
```

Pipe to file, before:

```js
const subprocess = $`echo foo`;
subprocess.stdout.pipe(fs.createWriteStream('stdout.txt'))
const {stdout} = await subprocess;
```

After:

```js
const {stdout} = await $`echo foo`.pipeStdout('stdout.txt');
```

Pipe to another process, before:

```js
const ffmpegProcess = $`ffmpeg -ss 5 -i input.mp4 -t 1 -vf cropdetect -f null -`
const awkProcess = $`awk ${'/crop/ {print $NF}'}`
const tailProcess = $`tail -1`

ffmpegProcess.stderr.pipe(awkProcess.stdin)
awkProcess.stdout.pipe(tailProcess.stdin)

const { stdout } = await tailProcess
```

After:

```js
const {stdout} = await $`ffmpeg -ss 5 -i input.mp4 -t 1 -vf cropdetect -f null -`
  .pipeStderr($`awk ${'/crop/ {print $NF}'}`)
  .pipeStdout($`tail -1`)
```

The PR's code and tests are done, but I haven't added the types, type tests nor readme documentation. I'd like to first confirm the overall idea and design make sense to you @sindresorhus. :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -453,29 +453,29 @@ export type ExecaChildPromise<StdoutStderrType extends StdoutStderrAll> = {
 
 	/**
 	[Pipe](https://nodejs.org/api/stream.html#readablepipedestination-options) the child process's `stdout` to `target`, which can be:
-    - Another `execa()` return value
-    - A writable stream
-    - A file path string
+	- Another `execa()` return value
+	- A writable stream
+	- A file path string
 
-  If the `target` is another `execa()` return value, it is returned. Otherwise, the original `execa()` return value is returned. This allows chaining `pipeStdout()` then `await`ing the final result.
+	If the `target` is another `execa()` return value, it is returned. Otherwise, the original `execa()` return value is returned. This allows chaining `pipeStdout()` then `await`ing the final result.
 
-  The `stdout` option] must be kept as `pipe`, its default value.
+	The `stdout` option] must be kept as `pipe`, its default value.
 	*/
 	pipeStdout?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeStdout?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
 
 	/**
-  Like `pipeStdout()` but piping the child process's `stderr` instead.
+	Like `pipeStdout()` but piping the child process's `stderr` instead.
 
-  The `stderr` option must be kept as `pipe`, its default value.
+	The `stderr` option must be kept as `pipe`, its default value.
 	*/
 	pipeStderr?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeStderr?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
 
 	/**
-  Combines both `pipeStdout()` and `pipeStderr()`.
+	Combines both `pipeStdout()` and `pipeStderr()`.
 
-  Either the `stdout` option or the `stderr` option must be kept as `pipe`, their default value. Also, the `all` option must be set to `true`.
+	Either the `stdout` option or the `stderr` option must be kept as `pipe`, their default value. Also, the `all` option must be set to `true`.
 	*/
 	pipeAll?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeAll?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -452,20 +452,31 @@ export type ExecaChildPromise<StdoutStderrType extends StdoutStderrAll> = {
 	cancel(): void;
 
 	/**
+	[Pipe](https://nodejs.org/api/stream.html#readablepipedestination-options) the child process's `stdout` to `target`, which can be:
+    - Another `execa()` return value
+    - A writable stream
+    - A file path string
 
-	 */
+  If the `target` is another `execa()` return value, it is returned. Otherwise, the original `execa()` return value is returned. This allows chaining `pipeStdout()` then `await`ing the final result.
+
+  The `stdout` option] must be kept as `pipe`, its default value.
+	*/
 	pipeStdout?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeStdout?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
 
 	/**
+  Like `pipeStdout()` but piping the child process's `stderr` instead.
 
-	 */
+  The `stderr` option must be kept as `pipe`, its default value.
+	*/
 	pipeStderr?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeStderr?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
 
 	/**
+  Combines both `pipeStdout()` and `pipeStderr()`.
 
-	 */
+  Either the `stdout` option or the `stderr` option must be kept as `pipe`, their default value. Also, the `all` option must be set to `true`.
+	*/
 	pipeAll?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeAll?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
 };
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 import {type Buffer} from 'node:buffer';
 import {type ChildProcess} from 'node:child_process';
-import {type Stream, type Readable as ReadableStream} from 'node:stream';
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -158,7 +160,9 @@ export function execa(file, args, options) {
 
 	spawned.all = makeAllStream(spawned, parsed.options);
 
-	return mergePromise(spawned, handlePromiseOnce);
+	addPipeMethods(spawned);
+	mergePromise(spawned, handlePromiseOnce);
```
ehmicky, line 164: `addPipeMethods()` relies on `spawned` being directly mutated, so we make it clear here.

lib/pipe.js
```diff
@@ -0,0 +1,40 @@
+import {Writable} from 'node:stream';
+import {createWriteStream} from 'node:fs';
+import {ChildProcess} from 'node:child_process';
+
+const isWritableStream = target => target instanceof Writable;
+
+const isExecaChildProcess = target => target instanceof ChildProcess && isWritableStream(target.stdin) && typeof target.then === 'function';
```
ehmicky, line 7: `instanceof` is not cross-realm safe for `Writable` and `ChildProcess`.
However, checking individual properties for existence seems brittle.

We could either use a `Symbol` key on Execa child processes, or a global `WeakSet`. However, that would not work for `instanceof Writable`.

So I opted for only using `instanceof`, but I am not sure whether this is a good choice.

lib/pipe.js
```diff
@@ -0,0 +1,38 @@
+import {createWriteStream} from 'node:fs';
+import {ChildProcess} from 'node:child_process';
+import {isWritableStream} from 'is-stream';
+
+const isExecaChildProcess = target => target instanceof ChildProcess && isWritableStream(target.stdin) && typeof target.then === 'function';
```
ehmicky, line 5: Not sure whether this is the best way to check this, considering `instanceof` is not cross-realm safe. However, checking individual properties for existence seems brittle. We could also either use a `Symbol` key on Execa child processes, or a global `WeakSet`.

sindresorhus: `Symbol` also does not work cross-realm.

lib/pipe.js
```diff
@@ -0,0 +1,38 @@
...
+		spawned[streamName].pipe(target.stdin);
+		return target;
+	}
+
+	throw new TypeError('The second argument must be a string, a stream or an Execa child process.');
+};
+
+export const addPipeMethods = spawned => {
+	if (spawned.stdout !== null) {
+		spawned.pipeStdout = pipeToTarget.bind(null, spawned, 'stdout');
```
sindresorhus, line 28: Nitpick:

```suggestion
		spawned.pipeStdout = pipeToTarget.bind(undefined, spawned, 'stdout');
```

lib/pipe.js
sindresorhus: I don't think cross-realm is a big concern. A lot of stuff don't work cross-realm.

sindresorhus:
This is very nice API sugar 👍

lib/pipe.js
ehmicky: Yes, right, neither a `Symbol` nor a global `WeakSet` would work cross-realm! :man_facepalming: 
The only cross-realm solution would be to detect for properties like `pid` or `stdio`.

In this case, the only situation where users would pass child processes cross-realm would be through `vm`, which seems like an unlikely edge case, so maybe we can ignore it.

index.d.ts
```diff
@@ -439,9 +441,38 @@ export type ExecaChildPromise<StdoutStderrType> = {
...
+    - Another `execa()` return value
+    - A writable stream
+    - A file path string
+
+  If the `target` is another `execa()` return value, it is returned. Otherwise, the original `execa()` return value is returned. This allows chaining `pipeStdout()` then `await`ing the final result.
+
+  The `stdout` option] must be kept as `pipe`, its default value.
+	*/
+	pipeStdout?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
+	pipeStdout?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
```
ehmicky, line 456: Those methods must be marked as optional since they depend on the value of the `stdout`, `stderr`, `stdio`, `all` and `encoding` options.

In the future, we might want to improve the types to infer the value of those options so they can be used to determine those methods' optionality. However, this would complicate the types quite a bit, so I'm leaving it off for this PR.

ehmicky:
I have added types and documentation.. The PR is ready to review :+1:

sindresorhus (approved):

PR #532, opened 2023-03-05
Add "verbose" option

### LINKED ISSUE #524: Verbose mode
## Problem

Projects with many processes or with complex process orchestration (parallel runs, deep/grand-child processes, etc.) can be hard to debug. This is a common pain point in terms of developer experience when working with processes.

This is mostly a problem when executing processes programmatically. In an interactive shell, users run commands one-by-one, can see the history of previous commands by scrolling up, and can see each command being input.

Also, some users of the new `$` API might expect opting in to a shell-like experience in that regard.

## Prior art

A common solution to this problem is to add a verbose mode. Examples include:
  - `bash -x` or `bash -v`
  - `zx` [`verbose` option](https://github.com/google/zx/#verbose)
  - Many build tools like `gulp` print each command being executed

## Solution

Add a `verbose` boolean option to Execa. When `true`, each command is printed before being executed: 
  - `"$ "` would be prepended
  - The command would be escaped like [`escapedCommand`](https://github.com/sindresorhus/execa#escapedcommand) using the [already existing quoting logic](https://github.com/sindresorhus/execa/blob/eae327cc0a49f61b17f699b9199c7953e1497c1a/lib/command.js#L12)

### Default value

Verbosity is mostly meant for debugging, i.e. it should be `false` by default. For example, `zx` currently defaults it `true` but is actually [reverting to `false`](https://github.com/google/zx/issues/569) in their next major release.

### stderr

The verbose mode prints to `stderr`, to separate from the command's normal output (`stdout`), which might be piped to another process or stream.

### NODE_DEBUG

When executing many commands and nesting them, it might be helpful to use an environment variable to set the verbose mode. We could use the core Node.js method [`debuglog()`](https://nodejs.org/api/util.html#utildebuglogsection-callback) for this purpose. Users would then use the `NODE_DEBUG=execa` environment variable.

### Errors

We can expect process failures to throw an exception which would either crash the process then logged, or be handled then logged. Therefore, we do not need to do anything there.

If the [`reject` option](https://github.com/sindresorhus/execa/blob/main/readme.md#reject) is `true`, the user indicates that process failure might be expected. In that case, we would let users print any information they want for debugging.

### Durations

Knowing how long a process took is useful for debugging. Many tools do this, e.g. Gulp.

However, it also comes with some cons. It might unnecessarily too verbose, by adding an additional line for every successful command. Also, when running commands in parallel, it might be hard to know which command ended without labeling or numbering them.

One potential solution would be to timestamp each command as they start, instead of printing their duration as they end. For example, instead of prepending `$ ` to each command, prepending `[HH:mm:ss.sss] `. Printing the date and timezone might not be useful for that purpose, i.e. could be left out.

### Colors

We might want (or not?) using colors. If we do, would the `FORCE_COLOR` environment variable be enough, or would this require a separate `colors` boolean option?

### Stdout/stderr

The verbose mode would not print stdout and stderr to avoid printing it twice. This is because `execa()` cannot know whether the user will `childProcess.stdout.pipe(process.stdout)` after execution. The same applies to failed commands.

Also, this makes the verbose mode more concise, yet still helpful. 

---

While this issue is detailed, the implementation could be rather small, and with a limited API surface (one boolean option), while still significantly improving debugging child processes.

ehmicky:
Fixes #524.

This adds a `verbose` boolean option. It defaults to `false`, unless the [`NODE_DEBUG=execa` environment variable](https://nodejs.org/api/util.html#debuglogenabled) is set. When enabled, it prints any command before executing it. It also prints a timestamp.

The "problem" and "prior art" sections in #524 give some background on why this feature might be a good addition.

For example, the following script:

```js
import {$} from 'execa'

await $`echo one`
await $`echo two`
```

Would print:

```
$ node example.js

$ NODE_DEBUG=execa node example.js
[22:44:57.723] echo one
[22:44:57.731] echo two
```

While the following:

```js
import {$} from 'execa'

const $$ = $({ stdio: 'inherit' })

await $$`echo one`
await $$`echo two`
```

Would print:

```
$ node example.js
one
two

$ NODE_DEBUG=execa node example.js
[22:47:52.056] echo one
one
[22:47:52.060] echo two
two
```

This can also be enabled for individual commands with:

```js
await $({ verbose: true })`echo one`
```

This feature works with any other Execa method too.

```js
await execa('echo', ['one'], { verbose: true })
```

The PR's code and tests are done, but I haven't added the types, type tests nor readme documentation. I'd like to first confirm the overall idea and design make sense to you @sindresorhus. :)

### DIFF
--- a/lib/verbose.js
+++ b/lib/verbose.js
@@ -1,0 +1,19 @@
+import {debuglog} from 'node:util';
+import process from 'node:process';
+
+export const verboseDefault = debuglog('execa').enabled;
+
+const padField = (field, padding) => String(field).padStart(padding, '0');
+
+const getTimestamp = () => {
+	const date = new Date();
+	return `${padField(date.getHours(), 2)}:${padField(date.getMinutes(), 2)}:${padField(date.getSeconds(), 2)}.${padField(date.getMilliseconds(), 3)}`;
+};
+
+export const logCommand = (escapedCommand, {verbose}) => {
+	if (!verbose) {
+		return;
+	}
+
+	process.stderr.write(`[${getTimestamp()}] ${escapedCommand}\n`);
+};
--- a/index.d.ts
+++ b/index.d.ts
@@ -244,6 +244,15 @@ export type CommonOptions<EncodingType> = {
 	@default true
 	*/
 	readonly windowsHide?: boolean;
+
+	/**
+	Print each command on `stderr` before executing it.
+
+  This can also be enabled by setting the `NODE_DEBUG=execa` environment variable in the current process.
+
+	@default false
+	*/
+	readonly verbose?: boolean;
 };
 
 export type Options<EncodingType = string> = {
--- a/index.d.ts
+++ b/index.d.ts
@@ -248,7 +248,7 @@ export type CommonOptions<EncodingType> = {
 	/**
 	Print each command on `stderr` before executing it.
 
-  This can also be enabled by setting the `NODE_DEBUG=execa` environment variable in the current process.
+	This can also be enabled by setting the `NODE_DEBUG=execa` environment variable in the current process.
 
 	@default false
 	*/
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ import {spawnedKill, spawnedCancel, setupTimeout, validateTimeout, setExitHandle
 import {handleInput, getSpawnedResult, makeAllStream, validateInputSync} from './lib/stream.js';
 import {mergePromise, getSpawnedPromise} from './lib/promise.js';
 import {joinCommand, parseCommand, parseTemplates, getEscapedCommand} from './lib/command.js';
+import {logCommand, verboseDefault} from './lib/verbose.js';
 
 const DEFAULT_MAX_BUFFER = 1000 * 1000 * 100;
 
@@ -44,6 +45,7 @@ const handleArguments = (file, args, options = {}) => {
 		cleanup: true,
 		all: false,
 		windowsHide: true,
+		verbose: verboseDefault,
 		...options,
 	};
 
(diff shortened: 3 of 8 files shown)


lib/verbose.js
```diff
@@ -0,0 +1,19 @@
+import {debuglog} from 'node:util';
+import process from 'node:process';
+
+export const verboseDefault = debuglog('execa').enabled;
```
ehmicky, line 4: Using `debuglog()` directly would print `EXECA` followed by the PID, which seems too verbose to me.
So I am only using it to get the `NODE_DEBUG` detection behavior.

lib/verbose.js
```diff
@@ -0,0 +1,19 @@
...
+	const date = new Date();
+	return `${padField(date.getHours(), 2)}:${padField(date.getMinutes(), 2)}:${padField(date.getSeconds(), 2)}.${padField(date.getMilliseconds(), 3)}`;
+};
+
+export const logCommand = (escapedCommand, {verbose}) => {
+	if (!verbose) {
+		return;
+	}
+
+	process.stderr.write(`[${getTimestamp()}] ${escapedCommand}\n`);
```
ehmicky, line 18: Should be colorized, e.g. in bold?

This would look better. However, this would come with the following downsides:
  - Adding a color dependency, like `chalk` or (probably better here) `yoctocolors`
  - Would require color support detection, which can be done with `stderr.hasColors()`. Users can then use the `FORCE_COLOR` environment variable to override the default behavior.

I am not sure whether that's a good idea or not. What do you think?

sindresorhus: I think we should skip colors for now.

sindresorhus:
This looks good to me. Carry on.

lib/verbose.js
ehmicky: Sounds good :+1:

ehmicky:
This is now documented and typed, ready to review :+1:

PR #535, opened 2023-03-06
Add `pipeToStdinFrom()` method

### LINKED ISSUE #534: Add `pipeStdin()` method
#531 added `pipeStdout()`, `pipeStderr()` and `pipeAll()` to redirect a child process's `stdout`/`stderr` to another process, a stream or a file.

For completeness and convenience, we should add a similar `pipeStdin()` which would redirect another process `stdout`, stream or file to a child process `stdin`.

This pattern is currently already documented in [our Tips section](https://github.com/sindresorhus/execa#redirect-input-from-a-file).

```js
const subprocess = execa('cat')
fs.createReadStream('stdin.txt').pipe(subprocess.stdin)
await subprocess
```

This would simplify it to:

```js
await execa('cat').pipeStdin('stdin.txt')
```

ehmicky:
Fixes #534.

This adds the following shortcut method: `pipeToStdinFrom()`.

For Node.js shell-like scripts, this is the equivalent to the `|` and `<` shell characters.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -479,6 +479,13 @@ export type ExecaChildPromise<StdoutStderrType extends StdoutStderrAll> = {
 	*/
 	pipeAll?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
 	pipeAll?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
+
+	/**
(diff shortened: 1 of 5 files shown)


ehmicky:
Closed due to https://github.com/sindresorhus/execa/issues/534#issuecomment-1456970781

PR #538, opened 2023-03-07
Improve documentation of the main return value

ehmicky:
Execa's return value can be confusing:
  - It is a child process, but different from `child_process.ChildProcess`
  - It is also a promise
  - It is different from `ChildProcessResult`, which is the value when awaiting that promise

This PR adds a specific `ChildProcess` mention to describe that return value, and distinguish it clearly from `ChildProcessResult`.


sindresorhus:
Do we need to update the TS docs comments too?

ehmicky:
Good catch! Fixed.

PR #540, opened 2023-03-07
Improve documentation of sync methods

ehmicky:
Just like in core Node.js modules, synchronous methods should mostly be avoided, except in specific situations where they are necessary. This PR improves their documentation based on this.

First, it moves them down in the readme and group them together. This reduces the amount of things to learn for users: they can just focus on the differences on the 4 main invocation methods, and then, as a follow-up, learn that synchronous methods for them exist as well.

Second, their description in both the readme and the type tests were a little out-of-sync. This updates them.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -495,6 +495,7 @@ Think of this as a mix of `child_process.execFile` and `child_process.spawn`.
 @returns An `ExecaChildProcess` that is both:
 	- a `Promise` resolving or rejecting with a `childProcessResult`.
 	- a [`child_process` instance](https://nodejs.org/api/child_process.html#child_process_class_childprocess) with some additional methods and properties.
+@throws A `childProcessResult` error
 
 @example
 ```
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -537,13 +537,11 @@ export function execa(file: string, options?: Options): ExecaChildProcess;
...
 /**
-Execute a file synchronously.
-
-This method throws an `Error` if the command fails.
+Same as `execa()` but synchronous.
 
 @param file - The program/script to execute.
 @param arguments - Arguments to pass to `file` on execution.
-@returns A result `Object` with `stdout` and `stderr` properties.
+@returns A `childProcessResult` object. This is also thrown on error.
```
aaronccasanova, line 544: Wonder if the [`@throws` tag](https://jsdoc.app/tags-throws.html) is appropriate here.

ehmicky: Good point! Fixed. :+1:

aaronccasanova (approved):

PR #541, opened 2023-03-07
Improve documentation of the usage section

ehmicky:
This improves the Usage section of the documentation. That section has been augmented ad-hoc incrementally over time and could use some small improvement. It also removes some information that is not strictly worth documenting, as it helps highlighting the features that are more interesting. 
  - Some of the scripting interface examples are slightly redundant, so I combined them into fewer examples, while keeping the same information.
  - Move the `signal` option's example to the Tips sections, since it is not really a feature anymore now that the option is supported by the `child_process` core module.
  - Remove examples of the synchronous methods, as the async version should usually (although not always) be preferred. The synchronous methods are documented in the API, and are pretty straightforward once you know the related async method.
  - Small description changes

I am hoping those changes help the Usage section to better focus and highlight the major features that Execa does bring.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -68,13 +68,21 @@ console.log(stdout);
 ```js
 import {$} from 'execa';
 
+await $({stdio: 'inherit'})`echo unicorns`;
+//=> 'unicorns'
+```
+
+#### Shared options
+
+```js
+import {$} from 'execa';
+
 const $$ = $({stdio: 'inherit'});
 
 await $$`echo unicorns`;
 //=> 'unicorns'
 
-await $$({shell: true})`echo unicorns && echo rainbows`;
-//=> 'unicorns'
+await $$`echo rainbows`;
 //=> 'rainbows'
 ```
 
--- a/readme.md
+++ b/readme.md
@@ -53,14 +53,19 @@ For more information about Execa scripts, please see [this page](docs/scripts.md
 ```js
 import {$} from 'execa';
 
-const unicorns = await $`echo unicorns`;
-const {stdout} = await $`echo ${unicorns}${'!'}}`;
-console.log(stdout);
-//=> 'unicorns!'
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -44,68 +46,38 @@ console.log(stdout);
...
-// const {stdout} = await $`echo ${['unicorns', 'rainbows']}`;
+const unicorns = await $`echo unicorns`;
+const {stdout} = await $`echo ${unicorns}${'!'}}`;
+console.log(stdout);
+//=> 'unicorns!'
 
+const {stdout} = await $`echo ${['unicorns', 'rainbows']}`;
 console.log(stdout);
-//=> 'unicorns'
+//=> 'unicorns rainbows'
```
aaronccasanova, line 63: This example errors on reassignment of block scoped var `stdout`

readme.md
```diff
@@ -44,68 +46,38 @@ console.log(stdout);
...
-```
-
-#### Synchronous
-
-```js
-import {$} from 'execa';
 
-const {stdout} = $.sync`echo unicorns`;
-console.log(stdout);
+await $$({shell: true})`echo unicorns && echo rainbows`;
```
aaronccasanova, line 76: This consolidation is nice, but removes the basic example of using the imported `$` with options e.g. `` await $(options)`command` ``. I'm not sure if this will lead to confusion, but user might think they have to use the `$$ = $(options)` binding mechanism first.

ehmicky: That's a good point, let's keep them split. Done :+1:

readme.md
ehmicky: Right! Fixed.

sindresorhus (approved):

ehmicky:
Fixed @aaronccasanova, ready for another code review :+1:

aaronccasanova (approved):

PR #542, opened 2023-03-07
Add `inputFile` option

ehmicky:
Discussed in https://github.com/sindresorhus/execa/issues/534#issuecomment-1457802815

This adds an `inputFile` option to pipe a file's contents to the process `stdin`. It works both in sync and async modes.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -260,6 +260,11 @@ export type Options<EncodingType = string> = {
 	Write some input to the `stdin` of your binary.
 	*/
 	readonly input?: string | Buffer | ReadableStream;
+
+	/**
+	Use a file as input to the the `stdin` of your binary.
+	*/
+	readonly inputFile?: string;
 } & CommonOptions<EncodingType>;
 
 export type SyncOptions<EncodingType = string> = {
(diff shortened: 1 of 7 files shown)


index.d.ts
```diff
@@ -260,13 +260,23 @@ export type Options<EncodingType = string> = {
 	Write some input to the `stdin` of your binary.
 	*/
 	readonly input?: string | Buffer | ReadableStream;
+
+	/**
+	Use a file as input to the the `stdin` of your binary.
```
sindresorhus, line 265: This should mention `input` somehow. Maybe using a `@see` annotation.

index.d.ts
```diff
@@ -260,13 +260,23 @@ export type Options<EncodingType = string> = {
 	Write some input to the `stdin` of your binary.
```
sindresorhus, line 260:
```suggestion
	Write some input to the `stdin` of your binary.

	Tip: If the input is a file, use the `inputFile` option instead.
```

sindresorhus: Feel free to rewrite it as you see fit.

index.d.ts
ehmicky: I added a Tip like the `input` option. Would that work?

PR #545, opened 2023-03-10
Improve main method documentation

ehmicky:
Users now have the choice between 4 methods: `execa()`, `execaCommand()`, `execaNode()` and `$`. This PR attempts to simplify the documentation of each of them and focus on their difference and on why users should use one over the other.

It also consolidates their documentation. For example, escaping/quoting was mentioned in different places, with different wording. Each method now includes only the information relevant for it. A dedicated "Shell syntax" section explains the bit about that, since that's a common source of confusion.

A few sections could also be greatly simplified without losing any meaningful information, in my opinion. 

Overall, I'm hoping this makes the main entry points easier to understand for new Execa users.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -504,9 +504,11 @@ ExecaChildPromise<StdoutStderrType> &
 Promise<ExecaReturnValue<StdoutStderrType>>;
 
 /**
-Execute a file.
+Executes a command using `file ...arguments`. `arguments` are specified as an array of strings. Returns a `childProcess`.
 
-Think of this as a mix of `child_process.execFile` and `child_process.spawn`.
+Arguments are automatically escaped. They can contain any character, including spaces.
+
+This is the preferred method when executing single commands.
 
 @param file - The program/script to execute.
 @param arguments - Arguments to pass to `file` on execution.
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -613,72 +615,26 @@ type TemplateExpression =
...
-	@throws A `childProcessResult` error
-
-	@example <caption>Basic</caption>
-	```
-	import {$} from 'execa';
-
-	const {stdout} = await $`echo unicorns`;
-	// const {stdout} = await $`echo ${'unicorns'}`;
-	// const {stdout} = await $`echo ${['unicorns', 'rainbows']}`;
+	Returns a new instance of `$` but with different default `options`. Consecutive calls are merged to previous ones.
```
ehmicky, line 618: Those code comments were accidentally documenting `` $`command` `` instead of `$(options)`.

readme.md
```diff
@@ -300,24 +330,6 @@ Combines both [`pipeStdout()`](#pipestdouttarget) and [`pipeStderr()`](#pipestde
 
 Either the [`stdout` option](#stdout-1) or the [`stderr` option](#stderr-1) must be kept as `pipe`, their default value. Also, the [`all` option](#all-2) must be set to `true`.
 
-### execaSync(file, arguments?, options?)
```
ehmicky, line 303: Those lines were not put in the right place.

sindresorhus:
Looks good

sindresorhus:

ehmicky:
Fixed :+1:

ehmicky:
Fixed :+1: 
Should I merge?

PR #550, opened 2023-03-13
Fix default value of `stdin` with `$`

### LINKED ISSUE #549: Default `stdin` to `inherit` with `$`
By default, `stdin` is `pipe` for all Execa methods, including `$`. This allows `childProcess.stdin.pipe()` to be used, and is also a requirement when using the `input` or `inputFile` option.

However, some commands detect this as "stdin is being piped/used", and hang forever. This was reported for example in https://github.com/sindresorhus/execa/issues/548 for ripgrep. 

Also, requesting user input is not uncommon in shell scripts.

zx fixed that by making `stdin` default to `inherit`, while keeping `stdout`/`stderr` default to `pipe`. 

https://github.com/google/zx/blob/main/src/core.ts#L146

I think we should do the same, with the following exceptions: 
  - Only for `$`, since it makes more sense in "shell-like" scripts. Also to avoid breaking every `execa()` user currently using `childProcess.stdin.pipe()` (which fails when `stdin` is `inherit`).
  - Not if the `input` or `inputFile` option is used.

Note: since `$` just got released, I am not sure whether this would be technically a breaking change or a bug fix. :thinking: 

What do you think?

ehmicky:
Fixes #549 

This makes `stdin` of processes created with `$` default to `inherit` instead of `pipe`, as this is most likely the intended behavior of users in that case.

This is technically a breaking change from `7.1.0`. However, considering this is a very recent release and this is correcting the intended behavior, I am not sure whether we can consider `7.1.0`'s behavior a bug, i.e. this would be a bug fix instead? :thinking:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,7 +27,7 @@ export type CommonOptions<EncodingType> = {
 
 	If you `$ npm install foo`, you can then `execa('foo')`.
 
-	@default `true` with `$`/`$.sync`, `false` otherwise
+	@default `true` with `$`, `false` otherwise
 	*/
 	readonly preferLocal?: boolean;
 
@@ -63,7 +63,7 @@ export type CommonOptions<EncodingType> = {
 	/**
 	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio).
 
-	@default 'pipe'
+	@default `inherit` with `$`, `pipe` otherwise
 	*/
 	readonly stdin?: StdioOption;
 
--- a/index.js
+++ b/index.js
@@ -232,6 +232,16 @@ export function execaSync(file, args, options) {
 	};
 }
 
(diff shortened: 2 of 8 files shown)


index.d.ts
```diff
@@ -27,7 +27,7 @@ export type CommonOptions<EncodingType> = {
 
 	If you `$ npm install foo`, you can then `execa('foo')`.
 
-	@default `true` with `$`/`$.sync`, `false` otherwise
```
ehmicky, line 30: I think `$.sync` is implied by `$`, so I am not sure we need to be explicit there.

sindresorhus:
> This is technically a breaking change from 7.1.0. However, considering this is a very recent release and this is correcting the intended behavior, I am not sure whether we can consider 7.1.0's behavior a bug, i.e. this would be a bug fix instead?

I would see it as a bug-fix.

sindresorhus (approved):

sindresorhus:
Can you fix the merge conflict? Feel free to merge and release afterwards.

PR #551, opened 2023-03-13
Improve validation of `$.sync` options binding

ehmicky:
To bind options with the synchronous `$` API, one must do:

```js
$(options).sync`command`
```

However, users are likely to do the following instead:

```js
$.sync(options)`command`
```

We could fix this either:
  - By supporting that second syntax.
  - By making it fail with a better error message.

This PR implements the second solution. However, I am happy to implement the first solution if you'd prefer it instead, just let me know!

### DIFF
--- a/index.js
+++ b/index.js
@@ -234,15 +234,19 @@ export function execaSync(file, args, options) {
 
 function create$(options) {
 	function $(templatesOrOptions, ...expressions) {
-		if (Array.isArray(templatesOrOptions)) {
-			const [file, ...args] = parseTemplates(templatesOrOptions, expressions);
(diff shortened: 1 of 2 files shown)


aaronccasanova (approved):
> However, users are likely to do the following instead:
> ```js
> $.sync(options)`command`
> ```

The options binding API is well documented and validated with TypeScript, so I'm not really sure how likely this is. However, I do agree with your choice to implement option 2 as a safe guard 👍 (as opposed to introducing another options binding API)

ehmicky:
I ran into the mistake myself, when writing automated tests, and it took me a little bit to realize what I was doing wrong :)

PR #553, opened 2023-03-13
Fix argument concatenation with `$`

ehmicky:
The `$` method concatenates arguments incorrectly currently. For example:

```js
const bar = '/bar'
$`mkdir /foo${bar}`
```

Currently behaves the same way as:

```js
const bar = '/bar'
$`mkdir /foo ${bar}`
```

I.e. it evaluates to `mkdir /foo /bar`. Instead, it should result in `mkdir /foo/bar`. Concatenation is a very common use case, especially (but not limited to) file paths.

This PR fixes this problem.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -76,21 +76,41 @@ const parseExpression = expression => {
 	throw new TypeError(`Unexpected "${typeOfExpression}" in template expression`);
 };
 
-const parseTemplate = (template, index, templates, expressions) => {
-	const templateString = template ?? templates.raw[index];
-	const templateTokens = templateString.split(SPACES_REGEXP).filter(Boolean);
+const concatTokens = (tokens, nextTokens, isNew) => isNew || tokens.length === 0 || nextTokens.length === 0
+	? [...tokens, ...nextTokens]
+	: [
+		...tokens.slice(0, -1),
+		`${tokens[tokens.length - 1]}${nextTokens[0]}`,
+		...nextTokens.slice(1),
+	];
 
-	if (index === expressions.length) {
-		return templateTokens;
+export const parseTemplates = (templates, expressions) => {
+	let tokens = [];
+
+	for (const [index, template] of templates.entries()) {
+		const templateString = template ?? templates.raw[index];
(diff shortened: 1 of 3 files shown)


lib/command.js
```diff
@@ -76,21 +76,45 @@ const parseExpression = expression => {
...
+		? expression.map(expression => parseExpression(expression))
+		: [parseExpression(expression)];
+	return concatTokens(
+		newTokens,
+		expressionTokens,
+		templateString.endsWith(' '),
+	);
+};
+
+export const parseTemplates = (templates, expressions) => {
```
ehmicky, line 111: We could use `Array.reduce()` too, but it is forbidden by xo, so I opted for a `for` loop instead.

test/command.js
```diff
@@ -171,6 +176,41 @@ test('$ handles invalid escape sequence', async t => {
...
+
+test('$ does not concatenate at the start of tokens with a space', async t => {
+	const {stdout} = await $`echo.js ${'foo'} bar`;
+	t.is(stdout, 'foo\nbar');
+});
+
+test('$ can concatenate at the start of tokens followed by an array', async t => {
+	const {stdout} = await $`echo.js ${['foo', 'bar']}foo`;
+	t.is(stdout, 'foo\nbarfoo');
+});
```
aaronccasanova, line 207: Tested locally, but this might be another good one to include:

```js
test('$ can concatenate at the start and end of tokens surrounding an array', async t => {
	const {stdout} = await $`echo.js foo${['bar', 'baz']}foo`;
	t.is(stdout, 'foobar\nbazfoo');
});
```

aaronccasanova (approved):
Nice catch @ehmicky! Totally agree this is the behavior I would expect 👍

sindresorhus (approved):

test/command.js
ehmicky: Good idea! Added in 366e6fcfe74bab8f5b074c9c0b8c70d9d974f88e

ehmicky:
Looks like this bug fix was quite necessary, we are getting some bug report about it: https://github.com/sindresorhus/execa/issues/554

PR #568, opened 2023-08-04
isCanceled should be true after abort in README

otakustay:


otakustay:
I amended my commit to include this change.

PR #569, opened 2023-08-04
Drop support for Node 14

ehmicky:
This drops support for Node 14 (breaking change) since it is not maintained anymore.
This also adds CI tests for Node 20.

This also upgrades some dependencies with major releases.


package.json
```diff
@@ -13,7 +13,7 @@
 	"type": "module",
 	"exports": "./index.js",
 	"engines": {
-		"node": "^14.18.0 || ^16.14.0 || >=18.0.0"
+		"node": "^16.14.0 || >=18.0.0 || >=20.0.0"
```
ehmicky, line 16: I am not sure the exact version range you'd prefer. Please let me know, and I can update it.

lib/command.js
```diff
@@ -10,14 +10,13 @@ const normalizeArgs = (file, args = []) => {
...
 const NO_ESCAPE_REGEXP = /^[\w.-]+$/;
-const DOUBLE_QUOTES_REGEXP = /"/g;
 
 const escapeArg = arg => {
 	if (typeof arg !== 'string' || NO_ESCAPE_REGEXP.test(arg)) {
 		return arg;
 	}
 
-	return `"${arg.replace(DOUBLE_QUOTES_REGEXP, '\\"')}"`;
+	return `"${arg.replaceAll('"', '\\"')}"`;
```
ehmicky, line 19: Linting-related updates due to `xo` upgrade.

package.json
```diff
@@ -44,7 +44,7 @@
 	"dependencies": {
 		"cross-spawn": "^7.0.3",
 		"get-stream": "^6.0.1",
```
ehmicky, line 46: Upgrading `get-stream` is not as straightforward, so I'll do it in a separate PR. This probably should be done before making the major release.

PR #570, opened 2023-08-06
Upgrade `signal-exit` to v4

ehmicky:
This upgrades `signal-exit` to a new major version.

---

Note: we should still wait until we upgrade `get-stream` before making a major release. That upgrade is currently being blocked by an issue in `get-stream` I have opened here: https://github.com/sindresorhus/get-stream/issues/56

### DIFF
--- a/test/kill.js
+++ b/test/kill.js
@@ -177,17 +177,17 @@ test('spawnAndKill cleanup detached SIGKILL', spawnAndKill, ['SIGKILL', true, tr
 // See #128
 test('removes exit handler on exit', async t => {
 	// @todo this relies on `signal-exit` internals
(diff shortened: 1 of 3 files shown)


test/kill.js
```diff
@@ -177,17 +177,17 @@ test('spawnAndKill cleanup detached SIGKILL', spawnAndKill, ['SIGKILL', true, tr
 // See #128
 test('removes exit handler on exit', async t => {
 	// @todo this relies on `signal-exit` internals
-	const emitter = process.__signal_exit_emitter__;
+	const emitter = globalThis[Symbol.for('signal-exit emitter')];
```
ehmicky, line 180: `signal-exit` v4 changes how it exposes its global event listener.

PR #571, opened 2023-08-17
Upgrade `get-stream`

ehmicky:
This upgrades `get-stream` to its latest version.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,6 +1,7 @@
 import {createReadStream, readFileSync} from 'node:fs';
+import {setTimeout} from 'node:timers/promises';
 import {isStream} from 'is-stream';
-import getStream from 'get-stream';
+import getStream, {getStreamAsBuffer} from 'get-stream';
 import mergeStream from 'merge-stream';
 
 const validateInputOptions = input => {
@@ -79,6 +80,9 @@ const getBufferedData = async (stream, streamPromise) => {
 		return;
 	}
 
+	// Wait for the `all` stream to receive the last chunk before destroying the stream
+	await setTimeout(0);
+
 	stream.destroy();
 
 	try {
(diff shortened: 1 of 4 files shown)


test/stream.js
```diff
@@ -17,6 +17,25 @@ test('buffer', async t => {
 	t.is(stdout.toString(), 'foo');
 });
 
+const checkEncoding = async (t, encoding) => {
+	const encodedString = '\u1000.';
```
ehmicky, line 21: The string was chosen to give different outputs with each encoding type.

lib/stream.js
```diff
@@ -93,11 +97,21 @@ const getStreamPromise = (stream, {encoding, buffer, maxBuffer}) => {
 		return;
 	}
 
-	if (encoding) {
-		return getStream(stream, {encoding, maxBuffer});
+	// eslint-disable-next-line unicorn/text-encoding-identifier-case
+	if (encoding === 'utf8' || encoding === 'utf-8') {
```
ehmicky, line 101: Allows both `utf8` and `utf-8` to mimic Node.js behavior.

lib/stream.js
```diff
@@ -93,11 +97,21 @@ const getStreamPromise = (stream, {encoding, buffer, maxBuffer}) => {
...
+		return getStreamAsBuffer(stream, {maxBuffer});
 	}
 
-	return getStream.buffer(stream, {maxBuffer});
+	return applyEncoding(stream, maxBuffer, encoding);
+};
+
+const applyEncoding = async (stream, maxBuffer, encoding) => {
+	const buffer = await getStreamAsBuffer(stream, {maxBuffer});
+	return buffer.toString(encoding);
```
ehmicky, line 114: There is one subtle change of behavior here (that is quite of an edge case): when using both `maxBuffer` and an unusual `encoding` (that is neither `utf8` nor `null`, such as `hex` or `base64`), then the number of bytes being limited by `maxBuffer` is stdout/stderr as a `Buffer`, not as the final value of `childProcess.stdout/stderr`, which is an hex/base64/etc.-encoded string.

test/stream.js
```diff
@@ -17,6 +17,25 @@ test('buffer', async t => {
...
+test('can pass encoding "utf16le"', checkEncoding, 'utf16le');
+test('can pass encoding "utf-16le"', checkEncoding, 'utf16le');
+test('can pass encoding "ucs2"', checkEncoding, 'utf16le');
+test('can pass encoding "ucs-2"', checkEncoding, 'utf16le');
+test('can pass encoding "latin1"', checkEncoding, 'latin1');
+test('can pass encoding "binary"', checkEncoding, 'latin1');
+test('can pass encoding "ascii"', checkEncoding, 'ascii');
+test('can pass encoding "hex"', checkEncoding, 'hex');
+test('can pass encoding "base64"', checkEncoding, 'base64');
+test('can pass encoding "base64url"', checkEncoding, 'base64url');
```
ehmicky, line 37: Ensure we support all the `encoding` option's value that `childProcess.exec()` supports.

test/stream.js
sindresorhus: I would write that as a short code comment as it's definitely not clear when just reading the code.

ehmicky: Done in cff527d6698016392dbc92a2cac6832e82a5fb66 :+1:

sindresorhus:
Yeah, go ahead 👍

ehmicky:
Done in [`8.0.0`](https://github.com/sindresorhus/execa/releases/tag/v8.0.0).

PR #572, opened 2023-08-19
Support `encoding: 'buffer'`

### LINKED ISSUE #480: Change the way you indicate you want a Buffer
Setting `{encoding: null}` is such a weird API I should never have inherited from `childProcess`.

We can introduce a new way while still supporting the old way.

Some possible solutions:

1. `{encoding: 'buffer'}` *(It's not technically an encoding, but who cares?)*
2. `{returnValue: 'string' | 'buffer'}`

In Got, we went with `{buffer: Boolean}`, but we already use that option name for something else.

ehmicky:
Fixes #480.

Before [`8.0.0`](https://github.com/sindresorhus/execa/releases/tag/v8.0.0), the `encoding: 'buffer'` option was supported like in `child_process`, albeit it was undocumented. This PR re-adds support for it, documents it, adds some tests, and improve the related types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,27 @@ export type StdioOption =
 	| number
 	| undefined;
 
-export type CommonOptions<EncodingType> = {
+type EncodingOption =
+  | 'utf8'
+  // eslint-disable-next-line unicorn/text-encoding-identifier-case
+  | 'utf-8'
+  | 'utf16le'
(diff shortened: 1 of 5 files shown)


ehmicky:
Published in [`8.0.1`](https://github.com/sindresorhus/execa/releases/tag/v8.0.1).

PR #575, opened 2023-09-13
minor fix: Scripts Examples

### LINKED ISSUE #574: Some problems according to examples
I tried to learn new Node.js scripts functions so I faced two problems:
1. I think [pipeAll](https://github.com/sindresorhus/execa/blob/main/docs/scripts.md#piping-stdout-and-stderr-to-another-command) example should be ```await $({all: true})`echo example`.pipeAll($`cat`);```

2. ```await $`echo tmp`.pipeStdout($`cat`);``` have following error:
```
throw new TypeError('The target child process\'s stdin must be available.');
TypeError: The target child process's stdin must be available.
```
which If we use ```await $`echo tmp`.pipeStdout(execa('cat'));``` it will be OK.

amammad:
Fixes #574

### DIFF
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -619,7 +619,7 @@ await Promise.all([echo, cat]);
 
 ```js
 // Execa
-await $({all: true})`echo example`.pipeAll($({ stdin: 'pipe' })`cat`);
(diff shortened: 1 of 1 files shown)


ehmicky (approved):
This looks good to me!
Thanks for creating this PR @amammad.

CI tests are failing due to an unrelated problem, see #576.

Issue #580, opened 2023-10-27
TypeScript: pipeStdout (and other pipeXxx) is optional property 


koshic:
What is the reason to keep it optional?

```ts
pipeStdout?<Target extends ExecaChildPromise<StdoutStderrAll>>(target: Target): Target;
pipeStdout?(target: WritableStream | string): ExecaChildProcess<StdoutStderrType>;
```

It can't be user without extra check or '!`:
<img width="667" alt="image" src="https://github.com/sindresorhus/execa/assets/20106607/f52eb6b3-43ed-4631-b69f-896ae0097bcc">

ehmicky:
Hi @koshic,

`pipeStdout()` is `undefined` if the `stdout` option is `'ignore'`, `'ipc'`, a file descriptor integer, or a stream.
The same applies to `pipeStderr()`, but using the `stderr` option.

`pipeAll()` is `undefined` if either: 
  - Both the `stdout` and `stderr` options are `'ignore'`, `'ipc'`, a file descriptor integer, or a stream.
  - The `all` option is not `true`.

https://github.com/sindresorhus/execa/blob/834e3726f0c7b7f23827b6e09b69b66dedbddd47/lib/pipe.js#L31-L33

Ideally the types would take this into account. Would you be interested in submitting a PR to fix them?

koshic:
@ehmicky yeah, let me experiment a bit )

koshic:
As you can see, if we will use current strategy - modify options via generic parameter, and provide all possible combinations as execa definitions, we have to multiply all definitions by 3 (new modifiers: stdout + stdin + all) which makes code unmaintainable:

<img width="606" alt="image" src="https://github.com/sindresorhus/execa/assets/20106607/86723f46-3aa1-49ba-9d13-b6e6f78d0a68">

I see only one way - keep modifiers for options, but produce exact ExecaChildProcess type by additional conditional type wrapper which will work like switch-case statement:
```ts
O extends { stdin: 'pipe' /* + default value check */} ? { pipeStdout: () => {} } : { }
```

@ehmicky what do you think about it?

ehmicky:
Yes, I think what you're highlighting was the reason I did not implement stricter typing in the first place when adding those pipe methods.

I agree with you. I'm not sure if this is what you are describing, but instead of method overloading, type inference should be used. I.e. only once `execa()` definition where `Options<EncodingType>` is being inferred from the parameters passed as input. Something like:

```ts
[...]

PR #583, opened 2023-11-01
Use type inference instead of multiple declarations

koshic:
See #580.

Change types from

```ts
export function execa(file: string, options?: Options): ExecaChildProcess;
export function execa(file: string, options?: Options<BufferEncodingOption>): ExecaChildProcess<Buffer>;
```

to

```ts
export function execa<EncodingType extends EncodingOption = DefaultEncodingOption>(
	file: string, options?: Options<EncodingType>
): ExecaChildProcess<EncodingType extends DefaultEncodingOption ? string : Buffer>;
```

It's a prerequisite to make smarter return type which allow us to properly infer `pipeXxx` types.


koshic:
@ehmicky could you plz take a look? there is a failed test, but it doesn't related to .d.ts changes

index.d.ts
```diff
@@ -32,7 +32,7 @@ type EncodingOption =
 type DefaultEncodingOption = 'utf8';
 type BufferEncodingOption = 'buffer' | null;
 
-export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingOption> = {
```
ehmicky, line 35: Why remove all the default values for the generic types? This makes it easier to use the types for the common use case (where `Buffer`s are not used).

koshic: It helps to be ensured that all variants of execa definition will have explicit type parameter. But I've forget about users who may use exported types in their apps, my bad. 

I'll restore those defaults, thx!

index.d.ts
```diff
@@ -637,18 +637,14 @@ setTimeout(() => {
...
-): ExecaChildProcess;
-export function execa(
-	file: string,
-	arguments?: readonly string[],
-	options?: Options<BufferEncodingOption>
-): ExecaChildProcess<Buffer>;
-export function execa(file: string, options?: Options): ExecaChildProcess;
-export function execa(file: string, options?: Options<BufferEncodingOption>): ExecaChildProcess<Buffer>;
+	options?: Options<EncodingType>
+): ExecaChildProcess<EncodingType extends DefaultEncodingOption ? string : Buffer>;
```
ehmicky, line 644: Should we create a type for `EncodingType extends DefaultEncodingOption ? string : Buffer` since it is used multiple times?

koshic: I thought about that but decided to keep that PR simple. At the next step we already should create complicated type set to infer results. But if you want, I can add something like `GetStdoutStderrType<EncodingType extends EncodingOption>`

ehmicky: > But if you want, I can add something like GetStdoutStderrType<EncodingType extends EncodingOption>

Yes, that sounds good. I think using such a type would actually simplify the PR, by making it shorter and with fewer `extends ? :` statements.

koshic: Ok, it shouldn't take too much time. But plz be ready to re-run those unstable tests )

ehmicky (approved):
This is looking good to me. Thanks a lot @koshic!

PR #586, opened 2023-11-19
`buffer` encoding option now returns `Uint8Array` instead of `Buffer`

### LINKED ISSUE #584: Return stdout/stderr/all as `Uint8Array` instead of `Buffer`?
https://sindresorhus.com/blog/goodbye-nodejs-buffer

We could use this small helper:

```js
export function bufferToUint8Array(value) {
    return new Uint8Array(value.buffer, value.byteOffset, value.byteLength);
}
```

GODLiangCY:
Fixes #584.

Allow `stdout` and `stderr` to be an ArrayBuffer to provide a more fine-grained control to users

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,10 +41,9 @@ type EncodingOption =
   | 'base64'
   | 'base64url'
   | 'buffer'
-  | null
   | undefined;
 type DefaultEncodingOption = 'utf8';
-type BufferEncodingOption = 'buffer' | null;
+type BufferEncodingOption = 'buffer';
 
 type GetStdoutStderrType<EncodingType extends EncodingOption> =
   EncodingType extends DefaultEncodingOption ? string : Buffer;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,3 @@
-import {type Buffer} from 'node:buffer';
 import {type ChildProcess} from 'node:child_process';
 import {type Stream, type Readable, type Writable} from 'node:stream';
 
@@ -46,7 +45,7 @@ type DefaultEncodingOption = 'utf8';
 type BufferEncodingOption = 'buffer';
 
 type GetStdoutStderrType<EncodingType extends EncodingOption> =
-  EncodingType extends DefaultEncodingOption ? string : Buffer;
+  EncodingType extends DefaultEncodingOption ? string : Uint8Array;
 
 export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingOption> = {
 	/**
@@ -218,7 +217,7 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	readonly shell?: boolean | string;
 
 	/**
-	Specify the character encoding used to decode the `stdout` and `stderr` output. If set to `'buffer'` or `null`, then `stdout` and `stderr` will be a `Buffer` instead of a string.
+	Specify the character encoding used to decode the `stdout` and `stderr` output. If set to `'buffer'`, then `stdout` and `stderr` will be a `Uint8Array` instead of a string.
 
 	@default 'utf8'
 	*/
@@ -343,7 +342,7 @@ export type NodeOptions<EncodingType extends EncodingOption = DefaultEncodingOpt
 	readonly nodeOptions?: string[];
 } & Options<EncodingType>;
 
-type StdoutStderrAll = string | Buffer | undefined;
+type StdoutStderrAll = string | Uint8Array | undefined;
 
 export type ExecaReturnBase<StdoutStderrType extends StdoutStderrAll> = {
 	/**
@@ -790,9 +789,9 @@ export function execaCommandSync<EncodingType extends EncodingOption = DefaultEn
 type TemplateExpression =
 	| string
 	| number
-	| ExecaReturnValue<string | Buffer>
-	| ExecaSyncReturnValue<string | Buffer>
-	| Array<string | number | ExecaReturnValue<string | Buffer> | ExecaSyncReturnValue<string | Buffer>>;
+	| ExecaReturnValue<string | Uint8Array>
+	| ExecaSyncReturnValue<string | Uint8Array>
+	| Array<string | number | ExecaReturnValue<string | Uint8Array> | ExecaSyncReturnValue<string | Uint8Array>>;
 
 type Execa$<StdoutStderrType extends StdoutStderrAll = string> = {
 	/**
@@ -820,7 +819,7 @@ type Execa$<StdoutStderrType extends StdoutStderrAll = string> = {
 	*/
 	(options: Options<undefined>): Execa$<StdoutStderrType>;
 	(options: Options): Execa$;
-	(options: Options<BufferEncodingOption>): Execa$<Buffer>;
+	(options: Options<BufferEncodingOption>): Execa$<Uint8Array>;
 	(
 		templates: TemplateStringsArray,
 		...expressions: TemplateExpression[]
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+import {Buffer} from 'node:buffer';
 import path from 'node:path';
 import childProcess from 'node:child_process';
 import process from 'node:process';
@@ -59,6 +60,14 @@ const handleArguments = (file, args, options = {}) => {
 	}
 
 	return {file, args, options};
+};
+
+const handleOutputSync = (options, value, error) => {
+	if (options.encoding === 'buffer' && Buffer.isBuffer(value)) {
+		value = new Uint8Array(value.buffer, value.byteOffset, value.byteLength);
+	}
+
+	return handleOutput(options, value, error);
 };
 
 const handleOutput = (options, value, error) => {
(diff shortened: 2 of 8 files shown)


ehmicky:
Hi @GODLiangCY,

Thanks a lot for this PR! It's great that you got the documentation and types fixed too :+1: 

I also like that you kept `encoding: 'buffer'` and `encoding: null` as `Buffer`, as it keeps Execa compatible with the underlying `child_process` options. I.e. it remains a superset of it.

We actually discussed this in https://github.com/sindresorhus/execa/issues/584 and the idea was to use `Uint8Array` instead of `ArrayBuffer`. I think this makes sense, although one can easily convert from one to the other. Could you please update it to `Uint8Array` instead?

Also related: https://github.com/sindresorhus/execa/issues/480

GODLiangCY:
> I think this makes sense, although one can easily convert from one to the other. Could you please update it to `Uint8Array` instead?

Thanks for your kind and quick reply. This looks way more reasonable too.

By the way, could you please help take a glance at these CI results?
https://github.com/sindresorhus/execa/actions/runs/6921001511
https://github.com/sindresorhus/execa/actions/runs/6921089138
These are a little confusing. Was it caused by some unstable factors?

sindresorhus:
The plan is to do it without a new option. Just Uint8Array instead of Buffer.

ehmicky:
@sindresorhus This means Execa would not be a superset of `child_process.spawn()` anymore, but it does simplify the API and generally makes sense, so this sounds good to me. This might also help you @GODLiangCY for this PR too, as the idea would then to replace `Buffer` by `Uint8Array` everywhere, so you can search any occurrences of it in the code base.

@GODLiangCY The CI tests are unstable, so if only specific combinations fail, please ignore it. I will restart them after code review to make it all green. Ideally we should fix those. We have done so in the past, but new unstable CI tests keep appearing. I think this is due to a combination of the amount of tests run in parallel, together with the nature of those tests (some of the features require several layers of spawning processes, with timeouts and potential race conditions).

sindresorhus:
> This means Execa would not be a superset of child_process.spawn() anymore, but it does simplify the API and generally makes sense, so this sounds good to me.

:+1:

> Should we keep the current value 'buffer' for the encoding option, or rename it?

Keep it.

> Also, if we're going that road, we should remove encoding: null

:+1:

index.js
```diff
@@ -63,13 +62,18 @@ const handleArguments = (file, args, options = {}) => {
...
 	}
 
 	if (options.stripFinalNewline) {
-		return stripFinalNewline(value);
+		if (typeof value === 'string') {
+			return stripFinalNewline(value);
+		}
+
+		const rawStripedString = stripFinalNewline(new TextDecoder().decode(value));
+		return new TextEncoder().encode(rawStripedString);
```
ehmicky, line 76: Decoding and re-encoding the `Uint8Array` to/from a string might be rather slow on big outputs.

Since `\n` and `\r` are always single UTF-8 bytes, we should be able to strip them by directly manipulating the `Uint8Array`.

Actually, it seems like [the code from `stripFinalNewline()`](https://github.com/sindresorhus/strip-final-newline/blob/main/index.js) should work with `Uint8Array`, not only `Buffer` (although I am not sure this is intentional).

However, that code does call `.slice()` once (twice on Windows), which requires copying the whole underlying `ArrayBuffer`, which can be slow too. 

A more efficient approach that does not involve any copying could be:

```js
const stripTypedArrayNewline = (value) => value[value.length - 1] === LF
  ? value.subarray(0, value[value.length - 2] === CR ? -2 : -1)
  : value

const LF = '\n'.charCodeAt();
const CR = '\r'.charCodeAt();
```

What do you think?

ehmicky: Quick benchmarks running 1e6 iterations, comparing:
  1. Using `.subarray()`
  2. Using `.slice()` with `strip-final-newline`
  3. Using `TextEncoder`/`TextDecoder` + `strip-final-newline`

For:
  - Unix newlines on a small input: 32ms (1), 40ms (2), 452ms (3)
  - Windows newlines on a small input: 31ms (1), 77ms (2), 477ms (3)
  - Unix newlines on a big input: 32ms (1), 1120ms (2), 2233ms (3)
  - Windows newlines on a big input: 31ms (1), 2378ms (2), 2705ms (3)

Small input is `hello\n` or `hello\n\r`, running 1e6 iterations.
Big input is the same but repeated 1e6 times, with 1e3 iterations.

ehmicky: Update: I have added a PR to `strip-final-newline` directly: https://github.com/sindresorhus/strip-final-newline/pull/5

lib/stream.js
```diff
@@ -92,8 +93,9 @@ const getStreamPromise = (stream, {encoding, buffer, maxBuffer}) => {
 		return getStream(stream, {maxBuffer});
 	}
 
-	if (encoding === null || encoding === 'buffer') {
-		return getStreamAsBuffer(stream, {maxBuffer});
+	if (encoding === 'buffer') {
+		const streamBufferPromise = getStreamAsBuffer(stream, {maxBuffer});
+		return streamBufferPromise.then(value => bufferToUint8Array(value));
```
ehmicky, line 98: We can use [`getStreamAsArrayBuffer()`](https://github.com/sindresorhus/get-stream#getstreamasarraybufferstream-options) here to completely remove the dependency on `Buffer`. It still requires converting to an `Uint8Array` though.

index.js
ehmicky: If you rebase onto `main`, you should be able to use the new `strip-final-newline` (https://github.com/sindresorhus/execa/pull/589), i.e. revert those changed lines.

index.d.ts
```diff
@@ -27,13 +27,12 @@ type EncodingOption =
   | 'base64'
```
ehmicky, line 27: The first line of this file (importing `Buffer` type) can be removed.

readme.md
```diff
@@ -679,10 +679,10 @@ We recommend against using this option since it is:
 
 #### encoding
 
-Type: `string | null`\
+Type: `string`\
 Default: `utf8`
 
-Specify the character encoding used to decode the `stdout` and `stderr` output. If set to `'buffer'` or `null`, then `stdout` and `stderr` will be a `Buffer` instead of a string.
```
ehmicky, line 685: The description should be updated in `index.d.ts` as well.

test/stream.js
```diff
@@ -16,9 +16,9 @@ const pExec = promisify(exec);
 setFixtureDir();
 
 test('buffer', async t => {
```
ehmicky, line 18: There are two tests in this file which currently checks whether the `input` option can be a `Buffer`.
The first one actually passes a `string` (which is incorrect) while the second passes a `Buffer`.
Maybe we should test passing an `Uint8Array` in both cases instead?

ehmicky:
Thanks a lot for all the work @GODLiangCY, I think we're almost there!

GODLiangCY:
I appreciate a lot for your patience and responsible code reviews! @ehmicky 

I'm not sure if I modified things correctly, can you please review them again?

sindresorhus:

test/stream.js
```diff
@@ -185,7 +189,7 @@ test('input option cannot be a String when stdio is set', t => {
 });
 
 test('input option can be a Buffer', async t => {
-	const {stdout} = await execa('stdin.js', {input: 'testing12'});
+	const {stdout} = await execa('stdin.js', {input: Buffer.from('testing12')});
 	t.is(stdout, 'testing12');
 });
```
ehmicky, line 194: We can remove this test.

test/stream.js
```diff
@@ -255,6 +259,11 @@ test('input option can be used with $.sync', t => {
 	t.is(stdout, 'foobar');
 });
 
+test('input option can be a Uint8Array - sync', t => {
+	const {stdout} = execaSync('stdin.js', {input: Uint8Array.from('foo', c => c.codePointAt(0))});
+	t.is(stdout, 'foo');
+});
+
 test('input option can be a Buffer - sync', t => {
 	const {stdout} = execaSync('stdin.js', {input: Buffer.from('testing12', 'utf8')});
```
ehmicky, line 268: We can remove this test as well.

test/stream.js
```diff
@@ -16,9 +16,9 @@ const pExec = promisify(exec);
 setFixtureDir();
 
 test('buffer', async t => {
-	const {stdout} = await execa('noop.js', ['foo'], {encoding: null});
-	t.true(Buffer.isBuffer(stdout));
-	t.is(stdout.toString(), 'foo');
+	const {stdout} = await execa('noop.js', ['foo'], {encoding: 'buffer'});
+	t.true(ArrayBuffer.isView(stdout));
+	t.is(new TextDecoder().decode(stdout), 'foo');
 });
```
ehmicky, line 22: We should add a similar test but for `execaSync()` with `encoding: 'buffer'`.

ehmicky:
Thanks again @GODLiangCY!
It appears there might be a bug currently with `execaSync()` (see above comment).

index.test-d.ts
```diff
@@ -149,7 +148,6 @@ expectError(execa('unicorns', {encoding: 'unknownEncoding'}));
 execa('unicorns', {execPath: '/path'});
 execa('unicorns', {buffer: false});
 execa('unicorns', {input: ''});
-execa('unicorns', {input: Buffer.from('')});
```
ehmicky, line 152: We should replace this type test with one with `Uint8Array`.

ehmicky (approved):
This looks good to me! (except for two small things)

ehmicky:
Thanks again @GODLiangCY!

PR #588, opened 2023-12-13
Fix processes not exiting on `inputFile` validation

ehmicky:
When both the `input` and `inputFile` options are specified, Execa throws. However, when doing so, the spawned process is not terminated, and keeps running.

This bug was not previously reported because Ava v3 terminates the test process even if some other processes are hanging. Ava v4 now captures this bug, so this PR upgrades Ava as well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ import {makeError} from './lib/error.js';
 import {normalizeStdio, normalizeStdioNode} from './lib/stdio.js';
 import {spawnedKill, spawnedCancel, setupTimeout, validateTimeout, setExitHandler} from './lib/kill.js';
 import {addPipeMethods} from './lib/pipe.js';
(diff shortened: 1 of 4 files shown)


test/node.js
```diff
@@ -100,4 +100,6 @@ test('node\'s forked script has a communication channel', async t => {
 
 	const message = await pEvent(subprocess, 'message');
 	t.is(message, 'pong');
+
+	subprocess.kill();
```
ehmicky, line 104: Another fix for a test which did not terminate its process properly.

index.js
```diff
@@ -82,6 +82,7 @@ export function execa(file, args, options) {
 	logCommand(escapedCommand, parsed.options);
 
 	validateTimeout(parsed.options);
+	validateInputOptions(parsed.options);
```
ehmicky, line 85: That's the main change: `input`/`inputFile` option validation now happens before process spawning, not after.

PR #594, opened 2023-12-14
Add `$.s` as an alias for `$.sync`

### LINKED ISSUE #581: feature request: shorter $.sync
Hello, and thanks for the library.

I love the `$.sync` command, but it is too long. I want to make scripts that are nice and short.

e.g.

```
// Bad
await $`foo`;
await $`bar`;
await $`baz`;

// Also bad
$.sync`foo`;
$.sync`bar`;
$.sync`baz`;
```

I propose a default alias for `$.sync` as `s`.

```
// Good!
s`foo`;
s`bar`;
s`baz`;
```

I know that I can do this for my own projects by wrapping `$.sync` and exporting it, but I think this feature should be for all users!

ehmicky:
Fixes #581.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -852,6 +852,56 @@ type Execa$<StdoutStderrType extends StdoutStderrAll = string> = {
 		templates: TemplateStringsArray,
 		...expressions: TemplateExpression[]
 	): ExecaSyncReturnValue<StdoutStderrType>;
+
+	/**
+	Same as $\`command\` but synchronous.
+
+	@returns A `childProcessResult` object
+	@throws A `childProcessResult` error
+
+	@example <caption>Basic</caption>
+	```
+	import {$} from 'execa';
+
+	const branch = $.s`git branch --show-current`;
+	$.s`dep deploy --branch=${branch}`;
+	```
+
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -852,6 +852,56 @@ type Execa$<StdoutStderrType extends StdoutStderrAll = string> = {
...
+
+	const $$ = $({stdio: 'inherit'});
+
+	$$.s`echo unicorns`;
+	//=> 'unicorns'
+
+	$$.s`echo rainbows`;
+	//=> 'rainbows'
+	```
+	*/
```
ehmicky, line 900: Not sure if there is a way to re-use (instead of duplicating) those documentation comments from the `sync()` method. :thinking:

index.js
```diff
@@ -263,6 +263,8 @@ function create$(options) {
 		return execaSync(file, args, normalizeScriptOptions(options));
 	};
 
+	$.s = $.sync;
```
ehmicky, line 266: This method does not use `this`, so no need to `.bind()`.

index.d.ts
sindresorhus: There's no way to do this, AFAIK.

PR #604, opened 2023-12-15
Allow `stdin` to be a sync/async iterable

### LINKED ISSUE #593: Add `stdin: Iterable` option
We could allow the `stdin` option to be a sync/async iterable (of strings or byte integers).

This would replace the `input` option, since `stdin: [string]` could be used instead.

Augmenting the `stdin` option instead of having a separate `input` option is more in line with the designs proposed in #591 and #592.

Also, being able to pass any iterable to `stdin` can have multiple use cases. For example, this would allow using a `function *()` generator to pass inputs to a process.

ehmicky:
Fixes #593.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,7 +86,9 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	readonly buffer?: boolean;
 
 	/**
-	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio) except it can also be an [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) or [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols).
+	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio).
+
+	It can also be an [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) or an [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols), providing neither [`execaSync()`](#execasyncfile-arguments-options), the [`input` option](#input) nor the [`inputFile` option](#inputfile) is used.
 
 	@default `inherit` with `$`, `pipe` otherwise
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,6 +12,8 @@ export type StdioOption =
 	| number
 	| undefined;
 
+type StdinOption = StdioOption | Iterable<string | Uint8Array> | AsyncIterable<string | Uint8Array>;
+
 type EncodingOption =
   | 'utf8'
   // eslint-disable-next-line unicorn/text-encoding-identifier-case
@@ -84,11 +86,11 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	readonly buffer?: boolean;
 
 	/**
-	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio).
+	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio) except it can also be an [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) or [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols).
 
 	@default `inherit` with `$`, `pipe` otherwise
 	*/
-	readonly stdin?: StdioOption;
+	readonly stdin?: StdinOption;
 
 	/**
 	Same options as [`stdio`](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_options_stdio).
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -52,14 +52,15 @@ const handleArguments = (file, args, options = {}) => {
 
 	options.env = getEnv(options);
 
-	options.stdio = normalizeStdio(options);
+	const stdio = normalizeStdio(options);
+	options.stdio = transformStdio(stdio);
```
ehmicky, line 56: We cannot directly convert iterables to a stream because `child_process.spawn()` only allows streams that have a file descriptor.

Instead, we pass `stdin: 'pipe'`. Once the process is spawned, we create a stream from the iterable and pipe it onto `spawned.sdin`.

`options.stdio` is the actual value passed to `child_process.spawn()` (i.e. `pipe`) while `parsed.stdio` keeps track of the original iterable passed as argument.

lib/stdio.js
```diff
@@ -1,5 +1,26 @@
+import {isStream} from 'is-stream';
+
 const aliases = ['stdin', 'stdout', 'stderr'];
 
+const isIterableStdin = stdinOption => typeof stdinOption === 'object'
+	&& stdinOption !== null
+	&& !isStream(stdinOption)
+	&& (typeof stdinOption[Symbol.asyncIterator] === 'function' || typeof stdinOption[Symbol.iterator] === 'function');
```
ehmicky, line 8: We explicitly exclude `string` and `Stream` since those already have a meaning with the `stdin` option.

lib/stream.js
```diff
@@ -1,17 +1,33 @@
...
+			throw new TypeError('The `stdin` option cannot be an iterable when the `inputFile` option is set.');
+		}
+	}
 };
 
 // `input` and `inputFile` option in sync mode
-export const handleInputSync = ({input, inputFile}) => {
+export const handleInputSync = ({input, inputFile}, stdio) => {
+	if (stdioHasIterableStdin(stdio)) {
+		throw new TypeError('The `stdin` option cannot be an iterable in sync mode');
```
ehmicky, line 28: Due to [the above](https://github.com/sindresorhus/execa/pull/604/files#r1427496220), this PR's logic requires stream piping, which cannot work in sync mode.

sindresorhus: This should be noted in the docs.

lib/stream.js
```diff
@@ -1,17 +1,33 @@
...
 	}
+
+	if (stdioHasIterableStdin(stdio)) {
+		if (input !== undefined) {
+			throw new TypeError('The `stdin` option cannot be an iterable when the `input` option is set.');
+		}
+
+		if (inputFile !== undefined) {
+			throw new TypeError('The `stdin` option cannot be an iterable when the `inputFile` option is set.');
+		}
```
sindresorhus, line 21: This should be noted in the docs.

index.d.ts
```diff
@@ -12,6 +12,8 @@ export type StdioOption =
 	| number
 	| undefined;
 
+type StdinOption = StdioOption | Iterable<string | Uint8Array> | AsyncIterable<string | Uint8Array>;
```
sindresorhus, line 15:
```suggestion
export type StdinOption = StdioOption | Iterable<string | Uint8Array> | AsyncIterable<string | Uint8Array>;
```

sindresorhus (changes requested):

lib/stream.js
ehmicky: Fixed. :+1:

lib/stream.js
ehmicky: Fixed :+1:

ehmicky:
This should be fixed.
I added proper error handling for when the iterable throws, which requires refactoring the whole PR.

lib/stream.js
```diff
@@ -86,14 +106,19 @@ const applyEncoding = async (stream, maxBuffer, encoding) => {
 	return buffer.toString(encoding);
 };
 
+// Handle any errors thrown by the iterable passed to the `stdin` option, if any.
+// We do not consume nor wait on that stream though, since it could potentially be infinite (like `process.stdin` in an interactive TTY).
+const throwOnStreamsError = streams => streams.filter(Boolean).map(stream => once(stream, ''));
```
ehmicky, line 111: `once()` throws on `error` event, regardless of the event passed as second argument (in that case, a dummy empty string).

Another way of writing this (longer, but maybe clearer?) would be:

```js
const [error] = await once(stream, 'error')
throw error
```

lib/stream.js
```diff
@@ -86,14 +106,24 @@ const applyEncoding = async (stream, maxBuffer, encoding) => {
...
 };
 
+// Handle any errors thrown by the iterable passed to the `stdin` option, if any.
+// We do not consume nor wait on that stream though, since it could potentially be infinite (like `process.stdin` in an interactive TTY).
+const throwOnStreamsError = streams => streams.filter(Boolean).map(stream => throwOnStreamError(stream));
+
+const throwOnStreamError = async stream => {
+	const [error] = await once(stream, 'error');
+	throw error;
+};
```
ehmicky, line 116: This could technically be just:

```js
const throwOnStreamError = stream => once(stream, '');
```

Since `once()` throws on `error` events, and stream events are never an empty string. But this would make the code harder to understand.

lib/stream.js
```diff
@@ -86,14 +106,24 @@ const applyEncoding = async (stream, maxBuffer, encoding) => {
...
+
 // Retrieve result of child process: exit code, signal, error, streams (stdout/stderr/all)
-export const getSpawnedResult = async (spawned, {encoding, buffer, maxBuffer}, processDone) => {
+export const getSpawnedResult = async (spawned, {encoding, buffer, maxBuffer}, {stdinIterableStream}, processDone) => {
 	const stdoutPromise = getStreamPromise(spawned.stdout, {encoding, buffer, maxBuffer});
 	const stderrPromise = getStreamPromise(spawned.stderr, {encoding, buffer, maxBuffer});
 	const allPromise = getStreamPromise(spawned.all, {encoding, buffer, maxBuffer: maxBuffer * 2});
+	const processDoneOrStreamsError = Promise.race([processDone, ...throwOnStreamsError([stdinIterableStream])]);
```
ehmicky, line 123: Once we had support for web streams to the `stdin`/`stdout`/`stderr` options, we probably will need to listen for their `error` events too. I.e. this line is written so we can add more streams in the future.

PR #607, opened 2023-12-15
Ensure process exits on stream errors

### LINKED ISSUE #606: Errors on `stdout`/`stderr` leave the process hanging
When the `stdout`/`stderr` streams of the child process error, the child process is not terminated by Execa.
In most cases, those errors indicate an issue which should also make the child process crash, but that's not always the case.
We should properly exit the process when either on those streams error.

ehmicky:
Fixes #606.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -87,19 +87,20 @@ const applyEncoding = async (stream, maxBuffer, encoding) => {
 };
 
 // Retrieve result of child process: exit code, signal, error, streams (stdout/stderr/all)
-export const getSpawnedResult = async ({stdout, stderr, all}, {encoding, buffer, maxBuffer}, processDone) => {
-	const stdoutPromise = getStreamPromise(stdout, {encoding, buffer, maxBuffer});
-	const stderrPromise = getStreamPromise(stderr, {encoding, buffer, maxBuffer});
-	const allPromise = getStreamPromise(all, {encoding, buffer, maxBuffer: maxBuffer * 2});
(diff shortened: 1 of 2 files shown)


lib/stream.js
```diff
@@ -87,19 +87,20 @@ const applyEncoding = async (stream, maxBuffer, encoding) => {
...
-	const allPromise = getStreamPromise(all, {encoding, buffer, maxBuffer: maxBuffer * 2});
+export const getSpawnedResult = async (spawned, {encoding, buffer, maxBuffer}, processDone) => {
+	const stdoutPromise = getStreamPromise(spawned.stdout, {encoding, buffer, maxBuffer});
+	const stderrPromise = getStreamPromise(spawned.stderr, {encoding, buffer, maxBuffer});
+	const allPromise = getStreamPromise(spawned.all, {encoding, buffer, maxBuffer: maxBuffer * 2});
 
 	try {
 		return await Promise.all([processDone, stdoutPromise, stderrPromise, allPromise]);
 	} catch (error) {
+		spawned.kill();
```
ehmicky, line 98: `.kill()` never throws. If the process already exited, it is a noop.

test/stream.js
```diff
@@ -298,3 +298,9 @@ if (process.platform !== 'win32') {
 		t.true(timedOut);
 	});
 }
+
+test('Errors on streams should make the process exit', async t => {
+	const childProcess = execa('forever');
+	childProcess.stdout.destroy();
+	await t.throwsAsync(childProcess, {code: 'ERR_STREAM_PREMATURE_CLOSE'});
+});
```
ehmicky, line 306: This test fails without the code change above.

PR #609, opened 2023-12-16
Handle `inputFile` missing file

### LINKED ISSUE #608: The `inputFile` option crashes the process when the file path is missing
When passing a missing file path to the `inputFile` option, the process crashes.
This is because the `error` events from the streams created by the `inputFile` option are not handled.

ehmicky:
Fixes #608.

To fix this, the `inputFile` option is re-using the error handling from the `stdin: iterable` option.
However, those two options were handled in different parts of the code.

Therefore, this PR is doing some rather big refactoring, merging the `inputFile`, `input` and `stdin: iterable` options. This should make it easier to maintain this part of the code.

It does not change behavior, except for the stream created by `inputFile` now being properly handled.

Additionally, this PR now validates that users are not passing both the `input`/`inputFile` options and the `stdin` option (except for `null`, `undefined`, `pipe` or `overlapped`) since those contradict each other and it always indicates a user mistake. For example, a user doing `execa(..., { input: '...', stdin: 'inherit' })` might not realize that those options cannot be used together. At the moment, they get some strange error message: this PR gives a more explicit description on how to fix the problem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,10 +7,10 @@ import stripFinalNewline from 'strip-final-newline';
 import {npmRunPathEnv} from 'npm-run-path';
 import onetime from 'onetime';
 import {makeError} from './lib/error.js';
-import {handleStdioOption, normalizeStdioNode} from './lib/stdio.js';
+import {handleStdioOption, handleInputOption, pipeStdioOption, normalizeStdioNode} from './lib/stdio.js';
 import {spawnedKill, spawnedCancel, setupTimeout, validateTimeout, setExitHandler} from './lib/kill.js';
 import {addPipeMethods} from './lib/pipe.js';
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -52,14 +52,12 @@ const handleArguments = (file, args, options = {}) => {
...
 	options.env = getEnv(options);
 
-	const stdioStreams = handleStdioOption(options);
-
 	if (process.platform === 'win32' && path.basename(file, '.exe') === 'cmd') {
 		// #116
 		args.unshift('/q');
 	}
 
-	return {file, args, options, parsed, stdioStreams};
```
ehmicky, line 62: `parsed.parsed` was dead code.

PR #614, opened 2023-12-16
Allow `stdin` to be a file path

ehmicky:
Part of #592. See also #610.

This allows the `stdin` option to be a file path string. I will send another PR for `stdout`/`stderr`.

It works with both Unix and Windows file paths. 
File paths can be either absolute or relative. It requires relative file paths to start with `.`, which prevents any ambiguity with the already existing string values of the `stdin` option.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,12 @@ export type StdioOption =
 	| number
 	| undefined;
 
-export type StdinOption = StdioOption | Iterable<string | Uint8Array> | AsyncIterable<string | Uint8Array> | URL;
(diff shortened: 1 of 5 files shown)


sindresorhus:
Can you make it throw an error if it's a string, does not start with `.` and is not one of

```
| 'pipe'
| 'overlapped'
| 'ipc'
| 'ignore'
| 'inherit'
```

That could catch some mistakes.

ehmicky:
Right! This should be fixed now (605930b5048c3f3f436f4f20837be009649d6bb4).

PR #621, opened 2023-12-18
Add `stdout|stderr: filePath | fileURL` option

### LINKED ISSUE #592: Allow `stdin`/`stdout`/`stderr` options to be a file URL

ehmicky:
Fixes #592.
Follow-up after #610 and #614.

This allows the `stdout`/`stderr` options to be a file path or a file URL.


ehmicky:
Ready for review. :+1:

PR #623, opened 2023-12-19
Refactor `stdio` options

ehmicky:
This is a big refactoring of the recent `stdio`-related features. As new PRs got added one after the other, it started getting messy.

This PR ensures that `execa()` and `execaSync()`  methods share as much logic as possible.
It also breaks down the main file into smaller ones.
Finally, the general cyclomatic complexity has been reduced.

I am hopeful this will make maintaining that part of the code easier in the future.

Although most methods have been rewritten, the behavior is the same, so the tests still pass and did not need to be updated (except for some error messages that slightly changed). We have some pretty good coverage for that part of the code base now, which should give us some confidence.

The test coverage slightly decreases (see Codecov PR comments), making the Codecov CI actions fail, but that's fixed by #622.


lib/stdio/normalize.js
```diff
@@ -0,0 +1,50 @@
+// Add support for `stdin`/`stdout`/`stderr` as an alias for `stdio`
+export const normalizeStdio = options => {
```
ehmicky, line 2: Note: this whole file has not changed.

PR #628, opened 2023-12-21
Refactor tests

ehmicky:
This PR refactors the tests related to `stdin`/`stdout`/`stderr`/`stdio`/`input`/`inputFile`. 
It does not change the logic in itself but instead:
  - Split into several smaller files
  - Remove duplicate code by using Ava test macros
  - Turn some sync I/O into async I/O instead, for better parallelism
  - Ensure every test is run on both `execa()` and `execaSync()`


sindresorhus:
Can you fix the conflict?

ehmicky:
Fixed the merged conflict, and made sure I kept the new tests added by the `Buffer -> Uint8Array` PR. :+1:

PR #629, opened 2023-12-21
Split `input`/`inputFile` options to their own file

ehmicky:
This is small refactoring splitting the logic for the `input` and `inputFile` options to its own file.

### DIFF
--- a/lib/stdio/input.js
+++ b/lib/stdio/input.js
@@ -1,52 +1,7 @@
 import {isStream as isNodeStream} from 'is-stream';
-import {getStdioOptionType, isRegularUrl, isUnknownStdioString} from './type.js';
-import {normalizeStdio} from './normalize.js';
-
-// Handle `input`, `inputFile`, `stdin`, `stdout` and `stderr` options, before spawning, in async/sync mode
-export const handleInput = (addProperties, options) => {
-	const stdio = normalizeStdio(options);
-	const stdioArray = arrifyStdio(stdio);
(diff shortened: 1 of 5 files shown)


lib/stdio/input.js
```diff
@@ -69,7 +24,6 @@ const handleInputFileOption = (stdioStream, index, inputFile, input) => {
 
 	const optionName = 'inputFile';
 	validateInputOption(stdioStream.value, optionName);
-	validateRegularUrl(inputFile, optionName);
```
ehmicky, line 72: This is now handled by the `validateFileStdio()` method above, which has been slightly reworked.

lib/stdio/type.js
```diff
@@ -29,7 +29,8 @@ export const isRegularUrl = stdioOption => isUrlInstance(stdioOption) && !hasFil
 
 const stringIsFilePath = stdioOption => stdioOption.startsWith('.') || isAbsolute(stdioOption);
 const isFilePath = stdioOption => typeof stdioOption === 'string' && stringIsFilePath(stdioOption);
-export const isUnknownStdioString = stdioOption => typeof stdioOption === 'string' && !stringIsFilePath(stdioOption) && !KNOWN_STDIO_STRINGS.has(stdioOption);
+
+export const isUnknownStdioString = (type, stdioOption) => type === 'native' && typeof stdioOption === 'string' && !KNOWN_STDIO_STRINGS.has(stdioOption);
```
ehmicky, line 33: `native` type is when the `stdin`/`stdout`/`stderr` value is passed directly to `child_process.spawn()`, as opposed to being handled by Execa. In that case, only few specific strings are allowed.

Issue #458, opened 2021-05-03
Support `URL` where paths are supported?


sindresorhus:
Node.js is adopting it for a lot of APIs.

- `file` argument.
- `cwd` option.
- `localDir` option.
- `execPath` option.
- `nodePath` option.

@ehmicky Thoughts?

---

If we decide to do this, we should probably open a Node.js issue about supporting `URL` for things in `childProcess` too.

ehmicky:
Yes, this makes sense :+1:

ehmicky:
Done for:
  - `cwd` option
  - `localDir` option 
  - `stdin`/`stdout`/`stderr` options

Not done for:
  - `execPath` option
  - `nodePath` option
  - `inputFile` option 
  - `file` argument

ehmicky:
`inputFile` done at #630.

ehmicky:
Only one missing: the `shell` option.

PR #631, opened 2023-12-21
Allow the first argument of `execa()` to be a file URL

ehmicky:
Part of https://github.com/sindresorhus/execa/issues/458

This allows the first argument to `execa()`, `execaSync()` and `execaNode()` to be  a file URL.

This does not add this to `execaCommand()` nor `$`, since those command expects both the command and the argument to be a single string. Therefore, using a file URL does not make sense there.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ import {Buffer} from 'node:buffer';
 import path from 'node:path';
 import childProcess from 'node:child_process';
 import process from 'node:process';
+import {fileURLToPath} from 'node:url';
 import crossSpawn from 'cross-spawn';
 import stripFinalNewline from 'strip-final-newline';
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -29,6 +30,18 @@ const getEnv = ({env: envOption, extendEnv, preferLocal, localDir, execPath}) =>
...
 };
 
+const getFilePath = file => {
+	if (file instanceof URL) {
+		return fileURLToPath(file);
+	}
+
+	if (typeof file !== 'string') {
+		throw new TypeError('First argument must be a string or a file URL.');
+	}
```
ehmicky, line 40: Catches potential mistake: `execa(['command', ...args])` instead of `execa('command', [...args])`.

PR #635, opened 2023-12-21
Allow `shell` option to be a file URL

### LINKED ISSUE #458: Support `URL` where paths are supported?

ehmicky:
Fixes #458.

This allows the `shell` option to be a file URL.


package.json
```diff
@@ -65,6 +65,7 @@
 		"path-key": "^4.0.0",
 		"tempfile": "^5.0.0",
 		"tsd": "^0.29.0",
+		"which": "^4.0.0",
```
ehmicky, line 68: I am not a huge fan of that module, but it is only used for a single test.

PR #638, opened 2023-12-22
Refactor `stdio` normalization logic

ehmicky:
When a user passes `stdio: undefined` or `stdio: 'pipe'`, we currently keep it as is.
This PR normalizes it instead to `stdio: [undefined, undefined, undefined]` and `stdio: ['pipe', 'pipe', 'pipe']`.

According to [Node.js documentation](https://nodejs.org/api/child_process.html#optionsstdio), this is the same:

> For convenience, options.stdio may be one of the following strings:
>    'pipe': equivalent to ['pipe', 'pipe', 'pipe'] (the default)

I also checked the Node.js source code. Our tests also pass.

Finally, users cannot access the actual options being passed, so if the behavior is the same, there is no difference from a user standpoint.

Normalizing `stdio` to be always an array is helping working with that option without having duplicate logic (one for when it is an array, one for when it is not). 

I.e. this PR does not change any behavior, it is just meant to simplify working on the `stdio` option.

### DIFF
--- a/lib/stdio/normalize.js
+++ b/lib/stdio/normalize.js
@@ -1,7 +1,7 @@
 // Add support for `stdin`/`stdout`/`stderr` as an alias for `stdio`
 export const normalizeStdio = options => {
 	if (!options) {
-		return;
+		return [undefined, undefined, undefined];
 	}
 
 	const {stdio} = options;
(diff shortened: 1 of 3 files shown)


lib/stdio/normalize.js
```diff
@@ -33,18 +33,5 @@ const aliases = ['stdin', 'stdout', 'stderr'];
...
-
-	if (stdio === undefined || typeof stdio === 'string') {
-		return [stdio, stdio, stdio, 'ipc'];
-	}
-
-	if (stdio.includes('ipc')) {
-		return stdio;
-	}
-
-	return [...stdio, 'ipc'];
```
ehmicky, line 49: All of this can now be summarized in a single line.

Issue #620, opened 2023-12-18
Allow `stdin`/`stdout`/`stderr` options to be an array


ehmicky:
Allowing the `stdin`/`stdout`/`stderr` options to be an array of values would be a small change from an API complexity standpoint, but offer some great expressiveness to users and handle a few important use cases.

---

A common use case that's been brought up multiple times:

```js
const { stdout } = await execa(..., { stdout: ['inherit', 'pipe'] })
```

I have needed this myself in production code. We used to document it in our `readme.md`. This is discussed in 2 long-standing open issues (which would then be solved): https://github.com/sindresorhus/execa/issues/121, https://github.com/sindresorhus/execa/issues/25.

Another use case would be:

```js
await execa(..., { stdin: ['inherit', 'pipe'] })
```

This was discussed when we changed the default value for the `stdin` option with Execa `$` since both `inherit` and `pipe` had benefits.

Other use case: logging errors on-disk while processing them programatically.

```js
const { stderr } = await execa(..., { stderr: [ 'pipe', './errors.txt'] })
```

---

Implementation-wise, all combinations should work (with a few small tweaks). When using `inherit` + something else, we would need to actually use `pipe` and emulate `inherit` with some piping logic (which mostly already exists due to the recent work around those options). This would mean the child process would not run in a TTY, so we should document that change of behavior.

What do you think?

sindresorhus:
👍

PR #639, opened 2023-12-23
Refactor `stdio` option logic

ehmicky:
This PR refactors the logic related to the `stdin`/`stdout`/`stderr`/`stdio` options.
This refactoring will help implementing #620, in another PR.
This does not change any behavior, only moves logic around.

### DIFF
--- a/lib/stdio/direction.js
+++ b/lib/stdio/direction.js
@@ -1,0 +1,42 @@
+import {
+	isStream as isNodeStream,
+	isReadableStream as isNodeReadableStream,
+	isWritableStream as isNodeWritableStream,
+} from 'is-stream';
+import {isWritableStream} from './type.js';
+
+// For `stdio[index]` beyond stdin/stdout/stderr, we need to guess whether the value passed is intended for inputs or outputs.
+// This allows us to know whether to pipe _into_ or _from_ the stream.
+export const addStreamDirection = stdioStream => {
+	const direction = getStreamDirection(stdioStream);
+	return addDirection(stdioStream, direction);
+};
+
+const getStreamDirection = stdioStream => KNOWN_DIRECTIONS[stdioStream.index] ?? guessStreamDirection[stdioStream.type](stdioStream.value);
+
+// `stdin`/`stdout`/`stderr` have a known direction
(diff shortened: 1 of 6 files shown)


lib/stdio/direction.js
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		return 'output';
+	},
+	native(stdioOption) {
+		if (isNodeStream(stdioOption)) {
+			return guessStreamDirection.nodeStream(stdioOption);
+		}
+	},
+};
```
ehmicky, line 37: The "direction" is whether the `stdio` option value is meant for input or output. For example, a readable vs a writable stream.

Some directions are ambiguous, e.g. a file path can be used both for reads or writes. In that case, we use `undefined`, which eventually defaults to `output`.

lib/stdio/handle.js
```diff
@@ -1,43 +1,34 @@
...
 import {handleInputOption, handleInputFileOption} from './input.js';
 
 // Handle `input`, `inputFile`, `stdin`, `stdout` and `stderr` options, before spawning, in async/sync mode
 export const handleInput = (addProperties, options) => {
 	const stdio = normalizeStdio(options);
-	const stdioStreams = stdio.map((stdioOption, index) => getStdioStream(stdioOption, index, addProperties, options));
+	const stdioStreams = stdio
+		.map((stdioOption, index) => getStdioStream(stdioOption, index, options))
+		.map(stdioStream => addStreamDirection(stdioStream))
+		.map(stdioStream => addStreamProperties(stdioStream, addProperties));
```
ehmicky, line 12: The main refactoring we are doing is to break down the `stdio` option normalization logic into three successive steps:
  1. Main step: retrieve type of value (stream, file path, etc.) and handle `input`/`inputFile` options
  2. Get the value's "direction"
  3. Add "stream properties", i.e. information that is both type-specific and direction-specific. For example, if the value is a file path and its direction is "input", we create a stream using `fs.createReadStream()`. Or, if the value is an iterable and its direction is "output", we fail since iterables are only allowed as inputs.

Breaking it into three steps is going to help implement #620 in the next PR.

PR #640, opened 2023-12-23
Document workaround for proper `all` interleaving

### LINKED ISSUE #502: `childProcess.all` interleaved result might be out-of-order
## Reproduction step

```console
$ cat noop-132.mjs                                     
#!/usr/bin/env node

process.stdout.write('1');
process.stderr.write('3');
process.stdout.write('2');
$ cat test.mjs    
import {execa} from 'execa';

const { all } = await execa('./noop-132.mjs', {all: true });
console.log(all);
$ node test.mjs
123
```

## Expect behavior

`node test.mjs` should output `132` as the README says:

> [Get interleaved output](https://github.com/sindresorhus/execa#all) from stdout and stderr similar to what is printed on the terminal. [(Async only)](https://github.com/sindresorhus/execa#execasyncfile-arguments-options)

```console
$ ./noop-132.mjs         
132
```

## Actual behavior

`node test.mjs`  output `123`

## Environment Info

```console
$ node -v      
v18.2.0
$ npm ls execa  
test@1.0.0 /Users/feng/test
└── execa@6.1.0
```

ehmicky:
Fixes #502.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -333,7 +333,7 @@ Can be disabled with `false`.
 
 Type: `ReadableStream | undefined`
 
-Stream combining/interleaving [`stdout`](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout) and [`stderr`](https://nodejs.org/api/child_process.html#child_process_subprocess_stderr).
+Stream [combining/interleaving](#ensuring-all-output-is-interleaved) [`stdout`](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout) and [`stderr`](https://nodejs.org/api/child_process.html#child_process_subprocess_stderr).
 
 This is `undefined` if either:
 - the [`all` option](#all-2) is `false` (the default value)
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -840,19 +840,51 @@ try {
...
+console.error('2'); // writes to stderr
+console.log('3'); // writes to stdout
+```
+
+This can be worked around by using `process.nextTick()` or `setTimeout()`.
+
+```js
+console.log('1');
+console.error('2');
+process.nextTick(() => {
```
sindresorhus, line 879: [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) is generally a better choice than `nextTick`.

ehmicky: Right!
However, it turns out both `queueMicrotask()` and `process.nextTick()` still group both `console.log()` together. But `setTimeout()` does work, so I switched to this instead.

ehmicky:
Fixed. :+1:

PR #643, opened 2023-12-25
Allow `stdin`/`stdout`/`stderr` to be an array

### LINKED ISSUE #25: Stdio `tap` option
For `stderr` and `stdout` the choice between `inherit` and `pipe`, has tradeoffs. 

Specifically, if you choose `inherit`, you loose the ability to observe the stream, it's piped straight to the parents stdout/stderr.

I propose we add a `tap` option. It will be converted internally to the `pipe` option, but we will grab the pipe, split it into two streams, one stream we pipe directly to `stdout/stderr`, the other we make available for observation.

### LINKED ISSUE #620: Allow `stdin`/`stdout`/`stderr` options to be an array

ehmicky:
Fixes #620
Fixes #25

This allows the `stdin`/`stdout`/`stderr` to be an array of values instead of a single one.

Some interesting use cases: "taping" (mentioned in #25):

```js
const { stdout } = await execa(..., { stdout: ['inherit', 'pipe'] })
```

Or redirecting to both terminal and file:

```js
const { stdout } = await execa(..., { stdout: ['inherit', './path.txt'] })
```

The way this works is: we pass `pipe` to `child_process.spawn()`, and emulate the options ourselves through manual piping. 
One caveat (which is documented by this PR): when using `inherit` + another value, even if the current `stdin`/`stdout`/`stderr` is an interactive TTY, the child process will not be interactive. This contrasts with using `inherit` on its own.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -125,7 +125,7 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- web [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 	- [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) or [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols)
 
-	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+	This can be an [array of values](https://github.com/sindresorhus/execa#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
 	@default `inherit` with `$`, `pipe` otherwise
 	*/
@@ -146,7 +146,7 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- file URL.
 	- web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
 
-	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+	This can be an [array of values](https://github.com/sindresorhus/execa#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
 	@default 'pipe'
 	*/
@@ -167,7 +167,7 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- file URL.
 	- web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
 
-	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+	This can be an [array of values](https://github.com/sindresorhus/execa#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
 	@default 'pipe'
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,9 +25,15 @@ type OutputStdioOption =
 	| Writable
 	| WritableStream;
 
-export type StdinOption = CommonStdioOption | InputStdioOption;
-export type StdoutStderrOption = CommonStdioOption | OutputStdioOption;
-export type StdioOption = CommonStdioOption | InputStdioOption | OutputStdioOption;
+export type StdinOption =
+	CommonStdioOption | InputStdioOption
+	| Array<CommonStdioOption | InputStdioOption>;
+export type StdoutStderrOption =
+	CommonStdioOption | OutputStdioOption
+	| Array<CommonStdioOption | OutputStdioOption>;
+export type StdioOption =
+	CommonStdioOption | InputStdioOption | OutputStdioOption
+	| Array<CommonStdioOption | InputStdioOption | OutputStdioOption>;
 
 type StdioOptions =
 	| BaseStdioOption
@@ -119,6 +125,8 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- web [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 	- [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) or [`AsyncIterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols)
 
+	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+
 	@default `inherit` with `$`, `pipe` otherwise
 	*/
 	readonly stdin?: StdinOption;
@@ -138,6 +146,8 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- file URL.
 	- web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
 
+	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+
 	@default 'pipe'
 	*/
 	readonly stdout?: StdoutStderrOption;
@@ -156,6 +166,8 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- file path. If relative, it must start with `.`.
 	- file URL.
 	- web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
+
+	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
 	@default 'pipe'
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -174,6 +174,17 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	readonly stderr?: StdoutStderrOption;
 
 	/**
+	Like the `stdin`, `stdout` and `stderr` options but for all file descriptors at once. For example, `{stdio: ['ignore', 'pipe', 'pipe']}` is the same as `{stdin: 'ignore', stdout: 'pipe', stderr: 'pipe'}`.
+
+	A single string can be used as a shortcut. For example, `{stdio: 'pipe'}` is the same as `{stdin: 'pipe', stdout: 'pipe', stderr: 'pipe'}`.
+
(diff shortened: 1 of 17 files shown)


lib/stdio/native.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	if (isNodeStream(value)) {
+		return {...stdioStream, type: 'nodeStream'};
+	}
+
+	return stdioStream;
+};
+
+// Node.js does not allow to easily retrieve file descriptors beyond stdin/stdout/stderr as streams
```
ehmicky, line 33: I discovered this when doing this PR. 
While Node.js has `childProcess.stdio[index]` to access any file descriptor as a stream, it does not have a way to do so on the current process, beyond file descriptors 0/1/2? 
I looked into the `fs` module, but although many methods can take a file descriptor as argument, and although `FileHandle` exists, there does not seem to be a way to open a file descriptor as a Node.js stream (or am I missing something?)

test/stdio/array.js
```diff
@@ -0,0 +1,253 @@
+import {readFile, writeFile, rm} from 'node:fs/promises';
```
ehmicky, line 1: Lots of tests! :)

readme.md
```diff
@@ -803,6 +809,18 @@ List of [CLI options](https://nodejs.org/api/cli.html#cli_options) passed to the
 
 ## Tips
 
+### Redirect stdin/stdout/stderr to multiple destinations
+
+The [`stdin`](#stdin), [`stdout`](#stdout-1) and [`stderr`](#stderr-1) options can be an array of values.
+The following example redirects `stdout` to both the terminal and an `output.txt` file, while also retrieving its value programmatically.
+
+```js
+const {stdout} = await execa('npm', ['install'], {stdout:['inherit', './output.txt', 'pipe']})
```
sindresorhus, line 818:
```suggestion
const {stdout} = await execa('npm', ['install'], {stdout: ['inherit', './output.txt', 'pipe']})
```

lib/stdio/native.js
sindresorhus: Did you try `fs.createReadStream('', {fd: 4})`?

index.d.ts
```diff
@@ -138,6 +146,8 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- file URL.
 	- web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
 
+	This can be an array of values such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+
```
sindresorhus, line 150: Maybe also link to the tip here: https://github.com/sindresorhus/execa/pull/643/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R812

sindresorhus (changes requested):

index.d.ts
ehmicky: Fixed. :+1:

lib/stdio/native.js
ehmicky: That was a very good tip, I forgot this existed!

Unfortunately, there is the following problem.

The stream created by `fs.createReadStream()` uses `fs.read(fd, ...)` under the hood. However, when the file descriptor points to an interactive TTY (or any other character device with blocking reads), `fs.read()` might never resolve, which makes the process hang. Even after calling both `stream.destroy()` and `fs.close(fd)`, `fs.read()` keeps going and holding the process.

https://github.com/nodejs/node/blob/345f15e493162e27fd8b543b7bc995aa45469639/lib/internal/fs/streams.js#L261

This can be reproduced by running this in an interactive TTY. The process hangs.

```js
const stream = createReadStream('', {fd: 0});
stream.read()
setTimeout(() => {
  stream.destroy();
}, 0);
```

The Node.js documentation for [`fs.createReadStream()`](https://nodejs.org/api/fs.html#fscreatereadstreampath-options) warns about this:

> If `fd` points to a character device that only supports blocking reads (such as keyboard or sound card), read operations do not finish until data is available. This can prevent the process from exiting and the stream from closing naturally.

I have been digging into the Node.js source code for `fs.createReadStream()` to find a solution, but I am not sure there is one. :thinking: If you think of anything, please let me know!

sindresorhus: Maybe ChatGPT can help:


Your analysis of the issue with `fs.createReadStream()` and blocking reads on character devices like TTYs is accurate. The stream's underlying `fs.read()` call can cause the process to hang since it waits for data to be available. Unfortunately, Node.js does not provide a built-in, straightforward way to handle this scenario with its standard APIs.

However, you can approach this problem using non-blocking techniques. One potential solution is to use `net.Socket` with the file descriptor. This allows you to handle TTY or other character devices in a non-blocking manner. The `net.Socket` class can be used to wrap a file descriptor and convert it into a stream that behaves more predictably with character devices.

Here's an example of how you might implement this:

```javascript
import { Socket } from 'net';

// Wrap the file descriptor in a net.Socket
const fd = 0; // Standard input (stdin)
const socket = new Socket({ fd, readable: true, writable: false });

// Read data from the socket
socket.on('data', (data) => {
    console.log('Received:', data.toString());
});

// Handle any errors
socket.on('error', (err) => {
    console.error('Socket error:', err);
});

// Example of cleaning up
setTimeout(() => {
    socket.destroy();
}, 5000); // Destroys the socket after 5 seconds
```

In this code, we create a `Socket` object and pass the file descriptor along with readability and writability options. This socket acts as a stream and provides the typical stream events like `data`, `error`, and `end`.

Using `net.Socket` is a more advanced technique and requires careful handling, especially regarding the lifetime and cleanup of the socket. Be aware that this approach may have nuances depending on the specifics of the file descriptor and the environment in which your Node.js application is running.


Another solution to handle non-blocking reads from a file descriptor in Node.js, especially for character devices like TTYs, is to use `polling` with the `fs.read` method. The idea is to repeatedly check if there's data available on the file descriptor without blocking the event loop.

Here's an approach using `setImmediate` or `setTimeout` to periodically poll the file descriptor:

1. **Use `fs.read` in Non-blocking Mode**: Call `fs.read` in a loop, but ensure it doesn't block the event loop. You can do this by using `setImmediate` or `setTimeout`.

2. **Manage the Read Buffer**: Handle the buffer management yourself.

3. **Proper Cleanup**: Ensure to clean up and close the file descriptor when done.

Here's a basic example to illustrate this:

```javascript
import { open, read, close } from 'fs';

// Open the file descriptor for reading (e.g., fd 0 for stdin)
const fd = 0; // Using stdin as an example

// Function to perform a non-blocking read
function nonBlockingRead() {
    const buffer = Buffer.alloc(1024);
    read(fd, buffer, 0, buffer.length, null, (err, bytesRead, buffer) => {
        if (err) {
            console.error('Error reading:', err);
            return;
        }

        if (bytesRead > 0) {
            console.log('Read:', buffer.slice(0, bytesRead).toString());
        }

        // Continue polling
        setImmediate(nonBlockingRead);
    });
}

// Start the non-blocking read loop
nonBlockingRead();

// Example cleanup after some time
setTimeout(() => {
    close(fd, (err) => {
        if (err) {
            console.error('Error closing file descriptor:', err);
        } else {
            console.log('File descriptor closed successfully.');
        }
    });
}, 10000); // 10 seconds
```

In this example, `nonBlockingRead` is a function that reads from the file descriptor in a non-blocking way. It uses `setImmediate` to continuously poll the file descriptor. The `setTimeout` is used just as an example to stop the polling and close the file descriptor after a certain period.

This approach gives you more control over how you read from the file descriptor, but it requires careful management of the buffer and the read loop. Additionally, this method is somewhat low-level and may not be as convenient as using built-in stream abstractions. Keep in mind that for very high-frequency data, this approach might still introduce performance overhead due to the repeated scheduling of callbacks.

ehmicky: I don't think the polling method ChatGPT is suggesting works. I do not think `fs.read()` has a polling option, and the example suggested uses the same API that is used by `fs.createReadStream()`. In other words, `fs.read()` will not return right away if there is no data available, it will block.

Now, the `Socket` suggestion would work. However, opening and managing a TCP socket seems quite heavy for the edge case we're covering here.

We could also detect whether the file descriptor points to a character device, and only throw in that case. However, it would lead to consumers to have their code fail when redirecting to an interactive terminal, but not in other cases, which will most likely cause confusion.

The case we're trying to cover is the following: `execa(..., { stdio: [..., ..., ..., ['inherit', ...])`. This requires the combination of all of:
  1. A custom file descriptor is used (beyond `stdin`/`stdout`/`stderr`)
  2. That custom file descriptor uses an array of multiple values, such as `['inherit', 'pipe']`
  3. One of those values is either `'inherit'`, or a file descriptor integer that is not 0/1/2

This is a very unlikely edge case, which might not justify adding too much complexity to the implementation. I think we might be good with just keeping it as is, and throwing a helpful error message. What do you think?

sindresorhus: > and throwing a helpful error message. I do not even think this might be worth documenting in the readme.md, as almost no users would run into that use case, and it might confuse them more than help.

Agreed

ehmicky: This should be done now. I have also added some code comments to summarize the above.

readme.md
```diff
@@ -606,9 +608,11 @@ Unless either [synchronous methods](#execasyncfile-arguments-options), the value
 - file URL.
 - web [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream).
 
+This can be an [array of values](#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
+
 #### stderr
 
-Type: `string | number | stream.Writable | WritableStream | URL`\
+Type: `string | number | stream.Writable | WritableStream | URL | StderrOption[]`\
```
sindresorhus, line 615: `StderrOption` doesn't mean anything to a user. It should either be expanded or linked to the types or something.

ehmicky: Could we something like this (to avoid the repetition, which might make the option look more complicated than it is)?

```
Type: `string | number | stream.Writable | WritableStream | URL`, or an `Array` of those types\
```

sindresorhus: 👍

ehmicky: Done in 222ebdfe7142afb0979cd8314b021a9588a2e077

PR #644, opened 2023-12-27
Document how to pass a Node.js stream to stdin/stdout/stderr

### LINKED ISSUE #617: Remove file descriptor restriction on the `stdin`/`stdout`/`stderr` option
Node.js requires the streams passed to the `stdio` option to have an underlying file descriptor.

https://github.com/nodejs/node/blob/0afe731d357d4df9a65aeb06429bf663d8dc2975/lib/internal/child_process.js#L1060
https://github.com/nodejs/node/blob/0afe731d357d4df9a65aeb06429bf663d8dc2975/lib/internal/child_process.js#L340

Execa could remove that restriction: if a stream without a file descriptor is passed to the `stdio` option(s), we could pipe it to the child process after it spawned. This would re-use the code that does the same thing for web streams (#615).

This was actually already discussed in #81, although in different terms.

However, Node.js detects whether there is an underlying file descriptor using the undocumented `stream._handle` and `stream.handle` properties. Also, those must be instances of `Pipe`, `TTY`, `TCP` or `UDP` which are internal classes not accessible userland. I do not think we should build detection logic based on this, as any Node.js release might break it.

Alternatively, we could make users opt-in to that behavior. For example, wrapping the stream into an array or a plain object instead of passing `stream` directly, to the `stdio` option(s).

Or, users could turn the Node.js stream into a web stream which we now support (#603), using `Readable.toWeb()`/`Writable.toWeb()`. I.e. we would just add some documentation about this.

ehmicky:
Fixes #617.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -830,7 +830,9 @@ When passing a Node.js stream to the [`stdin`](#stdin), [`stdout`](#stdout-1) or
 TypeError [ERR_INVALID_ARG_VALUE]: The argument 'stdio' is invalid.
 ```
 
-This limitation can be worked around by passing `[nodeStream, 'pipe']` instead of `nodeStream`.
+This limitation can be worked around by passing either:
+  - a web stream ([`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) or [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream))
+  - `[nodeStream, 'pipe']` instead of `nodeStream`
 
 ```diff
 - await execa(..., { stdout: nodeStream })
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -822,6 +822,21 @@ console.log(stdout);
...
+```
+TypeError [ERR_INVALID_ARG_VALUE]: The argument 'stdio' is invalid.
+```
+
+This limitation can be worked around by passing `[nodeStream, 'pipe']` instead of `nodeStream`.
+
+```diff
+- await execa(..., { stdout: nodeStream })
++ await execa(..., { stdout: [nodeStream, 'pipe'] })
+```
```
sindresorhus, line 838: Maybe also mention that they can just pass a web stream in instead?

ehmicky: Done :+1:

PR #647, opened 2024-01-05
Improve error handling

ehmicky:
This improves how the child process and its streams are cleaned up both on success and during errors.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,5 +1,4 @@
 import {once} from 'node:events';
-import {setTimeout} from 'node:timers/promises';
 import {finished} from 'node:stream/promises';
 import process from 'node:process';
 import getStream, {getStreamAsBuffer, getStreamAsArrayBuffer} from 'get-stream';
@@ -23,16 +22,13 @@ export const makeAllStream = (spawned, {all}) => {
 };
 
 // On failure, `result.stdout|stderr|all` should contain the currently buffered stream
+// They are automatically closed and flushed by Node.js when the child process exits
+// We guarantee this by calling `childProcess.kill()`
 const getBufferedData = async (stream, streamPromise) => {
 	// When `buffer` is `false`, `streamPromise` is `undefined` and there is no buffered data to retrieve
 	if (!stream || streamPromise === undefined) {
 		return;
 	}
-
-	// Wait for the `all` stream to receive the last chunk before destroying the stream
-	await setTimeout(0);
-
-	stream.destroy();
 
 	try {
 		return await streamPromise;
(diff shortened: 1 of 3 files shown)


lib/stream.js
```diff
@@ -23,17 +22,14 @@ export const makeAllStream = (spawned, {all}) => {
...
 const getBufferedData = async (stream, streamPromise) => {
 	// When `buffer` is `false`, `streamPromise` is `undefined` and there is no buffered data to retrieve
 	if (!stream || streamPromise === undefined) {
 		return;
 	}
 
-	// Wait for the `all` stream to receive the last chunk before destroying the stream
-	await setTimeout(0);
-
-	stream.destroy();
```
ehmicky, line 35: `childProcess.spawn()` automatically destroys `childProcess.stdin|stdout|stderr` on exit. Destroying it manually prevents proper flushing.

`await setTimeout(0)` was a hack to "fix" that improper flushing, which manifested itself more with `childProcess.all` since that stream involves an additional stream, therefore an additional delay.

The destroying logic originates from 8 years ago, in the following PR: https://github.com/sindresorhus/execa/pull/68, which itself originates from the following issue: https://github.com/sindresorhus/execa/issues/32

> Right now we aren't destroying those streams. Probably should.

We now have 600+ tests, which work properly without that logic. From checking Node.js current code, that cleanup logic is not necessary, and prevents proper flushing.

lib/stream.js
```diff
@@ -114,14 +113,14 @@ export const getSpawnedResult = async (
 			...stdioStreams.map(stdioStream => waitForStreamEnd(stdioStream, processDone)),
 		]);
 	} catch (error) {
-		return Promise.all([
+		spawned.kill();
```
ehmicky, line 116: For `childProcess.stdin|stdout|stderr` to be destroyed, we do need to ensure the child process is terminated on errors, i.e. this needs to happen before we wait for those streams' completion.

lib/stream.js
```diff
@@ -114,14 +113,14 @@ export const getSpawnedResult = async (
...
+		const results = await Promise.all([
 			{error, signal: error.signal, timedOut: error.timedOut},
 			getBufferedData(spawned.stdout, stdoutPromise),
 			getBufferedData(spawned.stderr, stderrPromise),
 			getBufferedData(spawned.all, allPromise),
 		]);
-	} finally {
-		cleanupStdioStreams(stdioStreams);
-		spawned.kill();
+		cleanupStdioStreams(stdioStreams, error);
```
ehmicky, line 123: This is only needed on errors. On success, the `end: true` option of `.pipe()` automatically closes those streams.

PR #652, opened 2024-01-06
Fix `error.stdout|stderr|all` type when `encoding` option is used

ehmicky:
Right now, the `encoding` option is not correctly applied to `error.stdout|stderr|all`. Namely:
  - When `encoding` is `buffer`, `error.stdout|stderr|all` is an `ArrayBuffer` instead of an `Uint8Array`
  - When `encoding` is something else like `hex`, `error.stdout|stderr|all` is a Node.js `Buffer` instead of a string
  - In both cases, our output normalization logic eventually transforms this to `undefined`

This is because the encoding logic is applied after `get-stream` has succeeded, but it is not applied on `error.bufferedData`, i.e. when `get-stream` has failed or been aborted.

This PR fixes this.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,7 +1,8 @@
+import {Buffer} from 'node:buffer';
 import {once} from 'node:events';
 import {finished} from 'node:stream/promises';
 import process from 'node:process';
-import getStream, {getStreamAsBuffer, getStreamAsArrayBuffer} from 'get-stream';
+import getStream, {getStreamAsArrayBuffer} from 'get-stream';
 import mergeStreams from '@sindresorhus/merge-streams';
 
 // `all` interleaves `stdout` and `stderr`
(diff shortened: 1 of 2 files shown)


lib/stream.js
```diff
@@ -24,40 +25,40 @@ export const makeAllStream = (spawned, {all}) => {
...
-		return;
-	}
-
+// When `buffer` is `false`, `streamPromise` is `undefined` and there is no buffered data to retrieve
+const getBufferedData = async (streamPromise, encoding) => {
 	try {
 		return await streamPromise;
 	} catch (error) {
-		return error.bufferedData;
+		return error.bufferedData === undefined ? undefined : applyEncoding(error.bufferedData, encoding);
```
ehmicky, line 33: `error.bufferedData` is `undefined` when the `encoding` option has an invalid value.

PR #653, opened 2024-01-06
Refactor execa promise to use `async/await`

ehmicky:
This PR refactors the child process `exit`/`error` events logic to use `async`/`await`. This does not change the behavior.

### DIFF
--- a/lib/promise.js
+++ b/lib/promise.js
@@ -1,3 +1,5 @@
+import {once} from 'node:events';
+
 // eslint-disable-next-line unicorn/prefer-top-level-await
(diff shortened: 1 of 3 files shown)


lib/promise.js
```diff
@@ -19,18 +21,7 @@ export const mergePromise = (spawned, promise) => {
...
-
-	spawned.on('error', error => {
-		reject(error);
-	});
-
-	if (spawned.stdin) {
-		spawned.stdin.on('error', error => {
-			reject(error);
-		});
-	}
```
ehmicky, line 35: Move this to where all the other streams errors are now being handled.

PR #654, opened 2024-01-06
Refactoring: move inner function to outer scope

ehmicky:
This is a small refactoring that moves some inner function to the outer scope. This splits the main function to distinguish in a clearer way what Execa does: 
  - Synchronously (`childProcess` spawned and returned right away) 
  - Asynchronously (waiting for `childProcess` exit and buffering its streams)

### DIFF
--- a/index.js
+++ b/index.js
@@ -136,49 +136,7 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	spawned.kill = spawnedKill.bind(null, spawned.kill.bind(spawned));
 	spawned.cancel = spawnedCancel.bind(null, spawned, context);
 
-	const handlePromise = async () => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -189,6 +147,54 @@ export function execa(rawFile, rawArgs, rawOptions) {
...
+		stdoutResult,
+		stderrResult,
+		allResult,
+	] = await getSpawnedResult(spawned, options, stdioStreams, processDone);
+	const stdout = handleOutput(options, stdoutResult);
+	const stderr = handleOutput(options, stderrResult);
+	const all = handleOutput(options, allResult);
+
+	if (error || exitCode !== 0 || signal !== null) {
+		const isCanceled = context.isCanceled || Boolean(options.signal?.aborted);
```
ehmicky, line 162: This line was also slightly reworked/simplified.

PR #659, opened 2024-01-07
Use `async/await` instead of direct `Promise` manipulation

ehmicky:
This PR refactors the logic to avoid passing promises around. Instead, it uses `async/await`.
This only refactors it, the behavior is not changed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,10 +10,10 @@ import {makeError} from './lib/error.js';
 import {handleInputAsync, pipeOutputAsync} from './lib/stdio/async.js';
 import {handleInputSync, pipeOutputSync} from './lib/stdio/sync.js';
 import {normalizeStdioNode} from './lib/stdio/normalize.js';
-import {spawnedKill, spawnedCancel, setupTimeout, validateTimeout, setExitHandler} from './lib/kill.js';
+import {spawnedKill, spawnedCancel, validateTimeout} from './lib/kill.js';
 import {addPipeMethods} from './lib/pipe.js';
 import {getSpawnedResult, makeAllStream} from './lib/stream.js';
-import {mergePromise, getSpawnedPromise} from './lib/promise.js';
+import {mergePromise} from './lib/promise.js';
 import {joinCommand, parseCommand, parseTemplates, getEscapedCommand} from './lib/command.js';
 import {logCommand, verboseDefault} from './lib/verbose.js';
 
@@ -62,6 +62,7 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
 		all: false,
 		windowsHide: true,
 		verbose: verboseDefault,
+		killSignal: 'SIGTERM',
 		...initialOptions,
 		shell: normalizeFileUrl(initialOptions.shell),
 	};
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -143,16 +144,12 @@ export function execa(rawFile, rawArgs, rawOptions) {
 }
 
 const handlePromise = async ({spawned, options, context, stdioStreams, command, escapedCommand}) => {
-	const spawnedPromise = getSpawnedPromise(spawned);
-	const timedPromise = setupTimeout(spawned, options, spawnedPromise);
-	const processDone = setExitHandler(spawned, options, timedPromise);
```
ehmicky, line 148: Instead of manipulating promises like this, put all this logic in the same function using `Promise.race()`/`Promise.all()` (see below).

lib/kill.js
```diff
@@ -51,27 +51,22 @@ export const spawnedCancel = (spawned, context) => {
 	}
 };
 
-const killAfterTimeout = async (spawned, timeout, killSignal, controller) => {
+const killAfterTimeout = async ({spawned, timeout, killSignal, context, controller}) => {
 	await pSetTimeout(timeout, undefined, {ref: false, signal: controller.signal});
 	spawned.kill(killSignal);
-	throw Object.assign(new Error('Timed out'), {timedOut: true, signal: killSignal});
+	Object.assign(context, {timedOut: true, signal: killSignal});
```
ehmicky, line 57: Pass this information using the `context` object instead of attaching it to the thrown error. 
Like this, if some library adds some unrelated `error.signal` property, we will ignore it.
The final error thrown by Execa uses `makeError()` which re-adds those properties.

lib/stream.js
```diff
@@ -90,43 +93,55 @@ const throwOnStreamError = async stream => {
...
 // Retrieve result of child process: exit code, signal, error, streams (stdout/stderr/all)
 export const getSpawnedResult = async (
 	spawned,
-	{encoding, buffer, maxBuffer},
+	{encoding, buffer, maxBuffer, timeout, killSignal, cleanup, detached},
+	context,
 	stdioStreams,
-	processDone,
 ) => {
+	const finalizers = [];
```
ehmicky, line 111: Array of functions to call after completion (success or error). At the moment, this cleans up the `timeout` and `cleanup` option listeners.

lib/stream.js
```diff
@@ -90,43 +93,55 @@ const throwOnStreamError = async stream => {
...
+			Promise.all([
+				once(spawned, 'exit'),
+				stdoutPromise,
+				stderrPromise,
+				allPromise,
+				...waitForCustomStreamsEnd(customStreams),
+			]),
+			...throwOnCustomStreamsError(customStreams),
+			...throwIfStreamError(spawned.stdin),
+			...throwOnTimeout({spawned, timeout, killSignal, context, finalizers}),
```
ehmicky, line 130: `Promise.all()` waits for each argument and returns it: 
  - child process exit code and signal
  - stdout/stderr/all output
  - wait for any custom stream to complete, but do not use their output. For example, the stream created by using `{ stdout: './path.txt' }`.

`Promise.race()` short-circuits it by throwing if either:
  - custom streams error
  - `stdin` errors
  - on timeout, with the `timeout` option

The `throw*(...)` methods never resolve, they only throw.

PR #663, opened 2024-01-08
Simplify `stdio` logic

ehmicky:
This PR refactors the `stdio` logic to simplify it. It does not change any behavior.

### DIFF
--- a/lib/stdio/direction.js
+++ b/lib/stdio/direction.js
@@ -10,24 +10,15 @@ import {isWritableStream} from './type.js';
 // This allows us to know whether to pipe _into_ or _from_ the stream.
 // When `stdio[index]` is a single value, this guess is fairly straightforward.
 // However, when it is an array instead, we also need to make sure the different values are not incompatible with each other.
-export const addStreamDirection = stdioStream => Array.isArray(stdioStream)
-	? addStreamArrayDirection(stdioStream)
-	: addStreamSingleDirection(stdioStream);
-
-const addStreamSingleDirection = stdioStream => {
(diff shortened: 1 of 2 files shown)


lib/stdio/direction.js
```diff
@@ -10,24 +10,15 @@ import {isWritableStream} from './type.js';
 // This allows us to know whether to pipe _into_ or _from_ the stream.
 // When `stdio[index]` is a single value, this guess is fairly straightforward.
 // However, when it is an array instead, we also need to make sure the different values are not incompatible with each other.
-export const addStreamDirection = stdioStream => Array.isArray(stdioStream)
-	? addStreamArrayDirection(stdioStream)
-	: addStreamSingleDirection(stdioStream);
```
ehmicky, line 15: Users can pass either a single value or an array of values to the `std*` option. At the moment, the logic handles both cases, for example in this line of code.

The main thing this PR does is to normalize single values to arrays (with a single item). This enables simplifying logic like this one.

lib/stdio/handle.js
```diff
@@ -7,45 +7,39 @@ import {handleNativeStream} from './native.js';
 // Handle `input`, `inputFile`, `stdin`, `stdout` and `stderr` options, before spawning, in async/sync mode
 export const handleInput = (addProperties, options) => {
 	const stdio = normalizeStdio(options);
-	const stdioStreams = stdio
+	const stdioStreamsGroups = stdio
```
ehmicky, line 10: `stdioStreamsGroups` is an array of arrays.

PR #668, opened 2024-01-09
Validate `input` option's type

ehmicky:
This PR improves the validation of the `input` option. Currently, if the wrong type is passed, the error message is confusing. This enforces that the `input` option is a `string`, `Uint8Array` or `Readable` Node.js stream.

### DIFF
--- a/lib/stdio/sync.js
+++ b/lib/stdio/sync.js
@@ -2,6 +2,7 @@ import {readFileSync, writeFileSync} from 'node:fs';
 import {isStream as isNodeStream} from 'is-stream';
 import {handleInput} from './handle.js';
 import {TYPE_TO_MESSAGE} from './type.js';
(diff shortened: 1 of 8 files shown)


lib/stdio/sync.js
```diff
@@ -24,7 +25,7 @@ const forbiddenIfSync = ({type, optionName}) => {
 
 const addPropertiesSync = {
 	input: {
-		filePath: ({value}) => ({value: readFileSync(value, 'utf8'), type: 'stringOrBuffer'}),
+		filePath: ({value}) => ({value: bufferToUint8Array(readFileSync(value)), type: 'uint8Array'}),
```
ehmicky, line 28: Performance optimization. Reading files as buffers instead of serializing to UTF-8 strings is faster.

PR #669, opened 2024-01-09
Improve documentation of synchronous methods

ehmicky:
The synchronous methods have multiple limitations. At the moment, those are documented in each method or option that has the limitation. 

This PR lists all the limitations directly in the documentation of the synchronous methods instead. This makes the documentation of the methods and options simpler. This also makes it more discouraging to use the synchronous methods by listing their limitations upfront.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -118,12 +118,10 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- `'inherit'`: Re-use the current process' `stdin`.
 	- an integer: Re-use a specific file descriptor from the current process.
 	- a Node.js `Readable` stream.
-
-	Unless the synchronous methods are used, the value can also be a:
-	- file path. If relative, it must start with `.`.
-	- file URL.
-	- web [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -118,12 +118,10 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
 	- `'inherit'`: Re-use the current process' `stdin`.
 	- an integer: Re-use a specific file descriptor from the current process.
 	- a Node.js `Readable` stream.
-
-	Unless the synchronous methods are used, the value can also be a:
-	- file path. If relative, it must start with `.`.
-	- file URL.
```
ehmicky, line 124: This was a mistake: file paths and URLs are actually allowed with the synchronous methods.

index.d.ts
```diff
@@ -737,6 +728,8 @@ export function execa<EncodingType extends EncodingOption = DefaultEncodingOptio
 /**
 Same as `execa()` but synchronous.
 
+Cannot use the following options: `all`, `cleanup`, `buffer`, `detached`, `serialization` and `signal`. Also, the `stdin`, `stdout`, `stderr`, `stdio` and `input` options cannot be a stream nor an iterable.
```
ehmicky, line 731: The limitations with the `cleanup`, `buffer`, `detached`, `serialization` and `signal` was not previously documented.

PR #671, opened 2024-01-10
Improve syntax of `std*: filePath` option

### LINKED ISSUE #667: Improved syntax for the `stdin|stdout|stderr: filePath` option
Background in https://github.com/sindresorhus/execa/issues/656#issuecomment-1882593657

Instead of passing the `stdin: './path.txt'` option, users should pass `stdin: { file: './path.txt' }` instead. This applies also to the `stdout` and `stderr` options. This does not apply to file `URLs`.

The reason is: those options can also be a string such as `inherit` or `pipe`. This creates the current problems:
  - This requires relative file paths to start with `.`
  - This could be a security problem when the file path is dynamic, as the user providing that file path could make it `inherit`, `pipe`, etc.

The new syntax is a plain object with a single `file` property, which value is a string.

ehmicky:
Fixes #667.

This replaces the `{ stdin: './file/path' }` option syntax by `{ stdin: { file: './file/path' } }` syntax instead, in order to avoid typing conflict with `{ stdin: string }` such as `{ stdin: 'inherit' }`. 

This PR also applies to the `stdout`, `stderr` and `stdio` options.

Since passing a file path to the `std*` options has not been released yet, this is not a breaking change.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,7 +13,7 @@ type CommonStdioOption =
 	| number
 	| undefined
 	| URL
-	| string;
+	| {file: string};
 
 type InputStdioOption =
 	| Iterable<string | Uint8Array>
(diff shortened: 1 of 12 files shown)


sindresorhus:
Conflict

ehmicky:
Fixed. :+1:

PR #676, opened 2024-01-11
Add `result.stdio` and `error.stdio`

### LINKED ISSUE #665: `result.stdio` and `error.stdio`.
We currently return the child process' `stdout` and `stderr`, as a property to the result and error objects.
We could do the same for the other file descriptors (3, 4, etc.), using [`childProcess.stdio`](https://nodejs.org/api/child_process.html#subprocessstdio). This would useful to any users writing to custom file descriptors.

It would behave just like `stdout` and `stderr` when it comes to buffering and encoding. It would in fact mostly use the same existing underlying logic, so the implementation would be fairly lightweight.

Just like `stdout` and `stderr`, it would be `undefined` when using `inherit`, `ignore`, `ipc`, etc. since those do leave `childProcess.stdio[*]`  as `null`.

`childProcess.spawnSync()` actually [does exactly this](https://nodejs.org/api/child_process.html#optionsdetached).

ehmicky:
Fixes #665.

This allows users to retrieve the output of other file descriptors than `stdout` and `stderr`, just like `child_process.spawnSync()` does.
Input streams like `stdin` are left as `undefined`.
It behaves exactly like `result.stdout` and `error.stdout`. For example, `result.stdio[1] === result.stdout` always.

Moreover, many of our tests are repeated on `stdin`, `stdout`, `stderr` and `stdio[3]`. This new property allowed simplifying many and removing some code duplication. This led to this PR having some rather big refactoring of our tests. I tried to split this into a separate PR, but this was difficult. Overall, this PR is only either refactoring existing tests, or adding new ones for `result.stdio` and `error.stdio`.

This leads this PR to be rather big, sorry for the review!

### DIFF
--- a/index.js
+++ b/index.js
@@ -80,32 +80,23 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
 	return {file, args, command, escapedCommand, options};
 };
 
-const handleOutputSync = (options, value, error) => {
+const handleOutput = (options, value) => {
+	if (typeof value !== 'string' && !ArrayBuffer.isView(value)) {
+		return;
+	}
+
 	if (Buffer.isBuffer(value)) {
 		value = bufferToUint8Array(value);
 	}
 
-	return handleOutput(options, value, error);
-};
-
-const handleOutput = (options, value, error) => {
-	if (typeof value !== 'string' && !ArrayBuffer.isView(value)) {
-		// When `execaSync()` errors, we normalize it to '' to mimic `execa()`
-		return error === undefined ? undefined : '';
-	}
-
-	if (options.stripFinalNewline) {
-		return stripFinalNewline(value);
-	}
-
-	return value;
+	return options.stripFinalNewline ? stripFinalNewline(value) : value;
 };
 
 export function execa(rawFile, rawArgs, rawOptions) {
 	const {file, args, command, escapedCommand, options} = handleArguments(rawFile, rawArgs, rawOptions);
 	validateTimeout(options);
 
-	const stdioStreams = handleInputAsync(options);
+	const {stdioStreams, stdioLength} = handleInputAsync(options);
 
 	let spawned;
 	try {
@@ -115,9 +106,8 @@ export function execa(rawFile, rawArgs, rawOptions) {
 		const dummySpawned = new childProcess.ChildProcess();
 		const errorPromise = Promise.reject(makeError({
 			error,
-			stdout: '',
-			stderr: '',
-			all: '',
+			stdio: Array.from({length: stdioLength}),
+			all: options.all ? '' : undefined,
 			command,
 			escapedCommand,
 			options,
@@ -147,12 +137,10 @@ export function execa(rawFile, rawArgs, rawOptions) {
 const handlePromise = async ({spawned, options, context, stdioStreams, command, escapedCommand}) => {
 	const [
 		[exitCode, signal, error],
-		stdoutResult,
-		stderrResult,
+		stdioResults,
 		allResult,
 	] = await getSpawnedResult(spawned, options, context, stdioStreams);
-	const stdout = handleOutput(options, stdoutResult);
-	const stderr = handleOutput(options, stderrResult);
+	const stdio = stdioResults.map(stdioResult => handleOutput(options, stdioResult));
 	const all = handleOutput(options, allResult);
 
 	if (error || exitCode !== 0 || signal !== null) {
@@ -161,8 +149,7 @@ const handlePromise = async ({spawned, options, context, stdioStreams, command, 
 			error,
 			exitCode,
 			signal,
-			stdout,
-			stderr,
+			stdio,
 			all,
 			command,
 			escapedCommand,
@@ -182,8 +169,9 @@ const handlePromise = async ({spawned, options, context, stdioStreams, command, 
 		command,
 		escapedCommand,
 		exitCode: 0,
-		stdout,
-		stderr,
+		stdio,
+		stdout: stdio[1],
+		stderr: stdio[2],
 		all,
 		failed: false,
 		timedOut: false,
@@ -195,7 +183,7 @@ const handlePromise = async ({spawned, options, context, stdioStreams, command, 
 export function execaSync(rawFile, rawArgs, rawOptions) {
 	const {file, args, command, escapedCommand, options} = handleArguments(rawFile, rawArgs, rawOptions);
 
-	const stdioStreams = handleInputSync(options);
+	const {stdioStreams, stdioLength} = handleInputSync(options);
 
 	let result;
 	try {
@@ -203,9 +191,7 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 	} catch (error) {
 		throw makeError({
 			error,
-			stdout: '',
-			stderr: '',
-			all: '',
+			stdio: Array.from({length: stdioLength}),
 			command,
 			escapedCommand,
 			options,
@@ -215,13 +201,13 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 	}
 
 	pipeOutputSync(stdioStreams, result);
-	const stdout = handleOutputSync(options, result.stdout, result.error);
-	const stderr = handleOutputSync(options, result.stderr, result.error);
+
+	const output = result.output || [undefined, undefined, undefined];
+	const stdio = output.map(stdioOutput => handleOutput(options, stdioOutput));
 
 	if (result.error || result.status !== 0 || result.signal !== null) {
 		const error = makeError({
-			stdout,
-			stderr,
+			stdio,
 			error: result.error,
 			signal: result.signal,
 			exitCode: result.status,
@@ -243,8 +229,9 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 		command,
 		escapedCommand,
 		exitCode: 0,
-		stdout,
-		stderr,
+		stdio,
+		stdout: stdio[1],
+		stderr: stdio[2],
 		failed: false,
 		timedOut: false,
 		isCanceled: false,
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -35,6 +35,11 @@ const getBufferedData = async (streamPromise, encoding) => {
 	}
 };
 
+const getStdioPromise = ({stream, index, stdioStreams, encoding, buffer, maxBuffer}) => {
+	const stdioStream = stdioStreams.find(stdioStream => stdioStream.index === index);
+	return stdioStream?.direction === 'output' ? getStreamPromise(stream, {encoding, buffer, maxBuffer}) : undefined;
+};
+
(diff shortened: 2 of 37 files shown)


index.js
```diff
@@ -80,32 +80,23 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
...
 
-const handleOutputSync = (options, value, error) => {
-	if (Buffer.isBuffer(value)) {
-		value = bufferToUint8Array(value);
-	}
-
-	return handleOutput(options, value, error);
-};
-
-const handleOutput = (options, value, error) => {
```
ehmicky, line 91: Merging those two functions does not change behavior.

index.js
```diff
@@ -115,9 +106,8 @@ export function execa(rawFile, rawArgs, rawOptions) {
 		const dummySpawned = new childProcess.ChildProcess();
 		const errorPromise = Promise.reject(makeError({
 			error,
-			stdout: '',
-			stderr: '',
-			all: '',
+			stdio: Array.from({length: stdioLength}),
+			all: options.all ? '' : undefined,
```
ehmicky, line 110: Small improvement. When the `all` option is `false`, `error.all` is `undefined`. However, in this specific case, it was left as an empty string instead, which was inconsistent.

test/encoding.js
```diff
@@ -0,0 +1,121 @@
+import {Buffer} from 'node:buffer';
```
ehmicky, line 1: File split from `test/stream.js`.

test/fixtures/noop-err.js
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env node
-import process from 'node:process';
-
-console.error(process.argv[2]);
```
ehmicky, line 4: Fixture replaced by `noop-fd.js`.

test/fixtures/noop-fd3.js
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env node
```
ehmicky, line 1: Fixture replaced by `noop-fd.js`.

test/fixtures/noop-no-newline.js
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env node
```
ehmicky, line 1: Fixture replaced by `noop-fd.js`.

test/fixtures/noop-throw.js
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env node
```
ehmicky, line 1: Fixture replaced by `fail.js`.

test/fixtures/stdin-fd3.js
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env node
-import process from 'node:process';
-import {readFileSync} from 'node:fs';
-
-const fileDescriptorIndex = Number(process.argv[2] || 3);
-console.log(readFileSync(fileDescriptorIndex, {encoding: 'utf8'}));
```
ehmicky, line 6: Fixture replaced by `stdin-fd.js`.

test/helpers/run.js
```diff
@@ -0,0 +1,6 @@
+import {execa, execaSync, $} from '../../index.js';
+
+export const runExeca = (file, options) => execa(file, options);
+export const runExecaSync = (file, options) => execaSync(file, options);
+export const runScript = (file, options) => $(options)`${file}`;
+export const runScriptSync = (file, options) => $(options).sync`${file}`;
```
ehmicky, line 6: Allows repeating the same tests across `execa()` vs `$` and async vs sync.

test/helpers/stdio.js
```diff
@@ -1,10 +1,17 @@
...
+
+export const getStdio = (indexOrName, stdioOption) => {
+	if (typeof indexOrName === 'string') {
+		return {[indexOrName]: stdioOption};
+	}
+
+	const stdio = ['pipe', 'pipe', 'pipe'];
+	stdio[indexOrName] = stdioOption;
+	return {stdio};
+};
```
ehmicky, line 13: Replaced `getStdinOption()`, and so on. This is more generic and results in simpler tests.

test/pipe.js
```diff
@@ -9,44 +9,44 @@ import {setFixtureDir} from './helpers/fixtures-dir.js';
...
+test('pipeStderr() can pipe to streams', pipeToStream, 2, 'pipeStderr');
+test('pipeAll() can pipe stdout to streams', pipeToStream, 1, 'pipeAll');
+test('pipeAll() can pipe stderr to streams', pipeToStream, 2, 'pipeAll');
 
-const pipeToFile = async (t, fixtureName, funcName, streamName) => {
+const pipeToFile = async (t, index, funcName) => {
 	const file = tempfile();
-	const result = await execa(fixtureName, ['test'], {all: true})[funcName](file);
-	t.is(result[streamName], 'test');
-	t.is(await readFile(file, 'utf8'), 'test\n');
```
ehmicky, line 38: Many tests are now using `noop-fd.js` instead of `noop.js`. That fixture does not add a newline, which explains why some tests now do not output that newline.

test/pipe.js
```diff
@@ -9,44 +9,44 @@ import {setFixtureDir} from './helpers/fixtures-dir.js';
...
-test.serial('pipeAll() can pipe stdout to files', pipeToFile, 'noop.js', 'pipeAll', 'stdout');
-test.serial('pipeAll() can pipe stderr to files', pipeToFile, 'noop-err.js', 'pipeAll', 'stderr');
+test.serial('pipeStdout() can pipe to files', pipeToFile, 1, 'pipeStdout');
+test.serial('pipeStderr() can pipe to files', pipeToFile, 2, 'pipeStderr');
+test.serial('pipeAll() can pipe stdout to files', pipeToFile, 1, 'pipeAll');
+test.serial('pipeAll() can pipe stderr to files', pipeToFile, 2, 'pipeAll');
 
 const invalidTarget = (t, funcName, getTarget) => {
-	t.throws(() => execa('noop.js', {all: true})[funcName](getTarget()), {
+	t.throws(() => execa('empty.js', {all: true})[funcName](getTarget()), {
```
ehmicky, line 49: Many tests which only validate that specific arguments/options should throw now use the `empty.js` fixture instead of the `noop.js` fixture, since the fixture is never actually run then.

test/stream.js
```diff
@@ -1,61 +1,15 @@
...
-const BUFFER_TO_ENCODE = Buffer.from(STRING_TO_ENCODE);
-
-test('can pass encoding "utf8"', checkEncoding, 'utf8');
-test('can pass encoding "utf-8"', checkEncoding, 'utf8');
-test('can pass encoding "utf16le"', checkEncoding, 'utf16le');
-test('can pass encoding "utf-16le"', checkEncoding, 'utf16le');
-test('can pass encoding "ucs2"', checkEncoding, 'utf16le');
-test('can pass encoding "ucs-2"', checkEncoding, 'utf16le');
-test('can pass encoding "latin1"', checkEncoding, 'latin1');
-test('can pass encoding "binary"', checkEncoding, 'latin1');
```
ehmicky, line 36: Those tests were incorrect. For example, `binary` should be `binary` not `latin1`. And `ucs2` should be `ucs2` not `utf16le`. Technically those values are the same, but the point of the test is to try each alternative name of each encoding.

test/test.js
```diff
@@ -125,34 +136,16 @@ const testExecPath = async (t, mapPath) => {
 test.serial('execPath option', testExecPath, identity);
 test.serial('execPath option can be a file URL', testExecPath, pathToFileURL);
 
-const emitStdinError = async subprocess => {
-	await setTimeout(0);
-	subprocess.stdin.emit('error', new Error('test'));
-};
-
-test('stdin errors are handled', async t => {
```
ehmicky, line 133: Moved to `test/stream.js`.

test/test.js
```diff
@@ -125,34 +136,16 @@ const testExecPath = async (t, mapPath) => {
...
 	const subprocess = execa('noop.js');
 	subprocess.emit('error', new Error('test'));
 	await t.throwsAsync(subprocess, {message: /test/});
 });
 
-test('child process errors rejects promise right away', async t => {
-	const subprocess = execa('noop.js');
-	subprocess.emit('error', new Error('test'));
-	await t.throwsAsync(subprocess, {message: /test/});
-});
```
ehmicky, line 151: That test is identical to the previous one.

lib/stream.js
```diff
@@ -112,16 +117,14 @@ export const getSpawnedResult = async (
...
+	const stdioPromises = spawned.stdio.map((stream, index) => getStdioPromise({stream, index, stdioStreams, encoding, buffer, maxBuffer}));
 	const allPromise = getStreamPromise(spawned.all, {encoding, buffer, maxBuffer: maxBuffer * 2});
 
 	try {
 		return await Promise.race([
 			Promise.all([
 				once(spawned, 'exit'),
-				stdoutPromise,
-				stderrPromise,
+				Promise.all(stdioPromises),
```
sindresorhus, line 127:
```suggestion
				...stdioPromises,
```

?

lib/stream.js
```diff
@@ -133,8 +136,7 @@ export const getSpawnedResult = async (
 		spawned.kill();
 		const results = await Promise.all([
 			[undefined, context.signal, error],
-			getBufferedData(stdoutPromise, encoding),
-			getBufferedData(stderrPromise, encoding),
+			Promise.all(stdioPromises.map(stdioPromise => getBufferedData(stdioPromise, encoding))),
```
sindresorhus, line 139: Same as above.

sindresorhus (changes requested):

lib/stream.js
ehmicky: The problem is that there are two variadic arguments: `stdioPromises` and `customStreams`, i.e. the length of those arrays is variable. But we still want to retrieve the resolved values of `stdioPromises` (not `customStreams`). 

I.e. we are doing this:

```js
const [
  [exitCode, signal, error],
  stdioResults,
  allResult,
] = Promise.all([
  once(spawned, 'exit'),
  Promise.all(stdioPromises), // length of `stdioPromises` is variable
  allPromise,
  ...customStreams.map(customStream => ...), // length of `customStreams` is variable
])
```

If we do this instead:

```js
const [
  [exitCode, signal, error],
  allResult,
  ...stdioResults,
] = Promise.all([
  once(spawned, 'exit'),
  allPromise,
  ...stdioPromises,
  ...customStreams.map(customStream => ...), 
])
```

`stdioResults` will get the results from both `stdioPromises` and `customStreams`.

sindresorhus: You are right. I didn't consider that.

ehmicky: I think it's because it is a little hard to see what's going on in there.

But I believe when we were previously passing promises around, we were obfuscating that complexity. Having a single function that shows the full flow with `async`/`await` is a little easier to wrap our head around, although it's still inherently complex.

ehmicky:
Fixed :+1:

PR #677, opened 2024-01-11
Fix type of `inputFile` option with `execaSync()`

ehmicky:
This fixes the type of the `inputFile` option, which incorrectly does not currently allows file URLs when using `execaSync()`.


ehmicky:
Closing in favor of #678.

PR #678, opened 2024-01-11
Improve types of synchronous methods

ehmicky:
This improves the types of synchronous methods:
  - Forbid streams and iterables with the `stdin`, `stdout`, `stderr`, `stdio` and `input` options
  - Forbid the `cleanup`, `buffer`, `all`, `serialization`, `detached` and `signal` options
  - Allows the `inputFile` option to be a file URL
  - Returns the `isCanceled` option (it is always `false` when synchronous, but it is present)
  - Add many type tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,30 +15,32 @@ type CommonStdioOption =
 	| URL
 	| {file: string};
 
-type InputStdioOption =
-	| Iterable<string | Uint8Array>
+type InputStdioOption<IsSync extends boolean = boolean> = IsSync extends true
+	? Uint8Array
+	: Iterable<string | Uint8Array>
 	| AsyncIterable<string | Uint8Array>
 	| Uint8Array
 	| Readable
 	| ReadableStream;
 
-type OutputStdioOption =
-	| Writable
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -343,39 +290,76 @@ export type CommonOptions<EncodingType extends EncodingOption = DefaultEncodingO
...
+	- `json`: Uses `JSON.stringify()` and `JSON.parse()`.
+	- `advanced`: Uses [`v8.serialize()`](https://nodejs.org/api/v8.html#v8_v8_serialize_value)
+
+	[More info.](https://nodejs.org/api/child_process.html#child_process_advanced_serialization)
+
+	@default 'json'
+	*/
+	readonly serialization?: 'json' | 'advanced';
 
-export type SyncOptions<EncodingType extends EncodingOption = DefaultEncodingOption> = {
```
sindresorhus, line 364: I would still keep this as a type alias

```js
export type SyncOptions<EncodingType extends EncodingOption = DefaultEncodingOption> = Options<true, EncodingType>;
```

PR #681, opened 2024-01-12
Improve types of `result.stdout|stderr|all`

### LINKED ISSUE #673: Invalid types for `result.stdout|stderr|all`
`result.stdout` is `undefined` when either:
  - the `buffer` option is `false`
  - the `stdout` option is `inherit`, `ipc`, `ignore`, a Node.js stream or an integer. Or an array with a single item with one of those values.

At the moment, `result.stdout` is always typed as either `string` or `Uint8Array` instead.

This problem applies also to `result.stderr` and `result.all`.

ehmicky:
Fixes #673.

Most of the time, we know that `result.stdout|stderr|all` is defined. However, we currently type it as possibly `undefined`. For example, the following fails becomes `stdout` is typed as `string | undefined` even though we know it cannot be `undefined`.

```js
const { stdout } = await execa('echo')
console.log(stdout.trim())
```

The reverse is true: we sometimes know that `result.stdout|stderr|all` is `undefined`. Typing it as `undefined` instead of `string | undefined` would help TypeScript users realize they cannot use that property. For example, when using the `buffer: false` option or the `stdout: 'inherit'` option, users cannot retrieve `result.stdout`. The type should then be `undefined`, not `string | undefined`.

This PR implements this. The stream output is now correctly either `undefined`, `string` or `Uint8Array`.

This PR works with all the following cases:
  - New `result.stdio` property
  - `stdio: string` option
  - `stdio: array` option
  - `stdio` option with more than 3 items
  - `buffer: false` option
  - `result.all` property

Example from the type tests:

```ts
const { stdout, stderr, all, stdio } = await execa('unicorns', {stdio: ['pipe', 'ignore', 'pipe', 'pipe', new Uint8Array()]});
expectType<undefined>(stdout);
expectType<string>(stderr);
expectType<string | undefined>(all);
expectType<undefined>(stdio[0]);
expectType<undefined>(stdio[1]);
expectType<string>(stdio[2]);
expectType<string>(stdio[3]);
expectType<undefined>(stdio[4]);
// stdio[5] fails
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,15 @@
 import {type ChildProcess} from 'node:child_process';
 import {type Readable, type Writable} from 'node:stream';
 
+type NoOutputStdioOption =
+	| 'ignore'
+	| 'inherit'
+	| 'ipc'
+	| number
+	| Readable
+	| Writable
+	| [NoOutputStdioOption];
+
 type BaseStdioOption =
 	| 'pipe'
 	| 'overlapped'
@@ -38,9 +47,14 @@ export type StdioOption<IsSync extends boolean = boolean> =
 	CommonStdioOption | InputStdioOption | OutputStdioOption<IsSync>
 	| Array<CommonStdioOption | InputStdioOption | OutputStdioOption<IsSync>>;
 
-type StdioOptions<IsSync extends boolean = boolean> =
-	| BaseStdioOption
-	| readonly [StdinOption<IsSync>, StdoutStderrOption<IsSync>, StdoutStderrOption<IsSync>, ...Array<StdioOption<IsSync>>];
+type StdioOptionsArray<IsSync extends boolean = boolean> = readonly [
+	StdinOption<IsSync>,
+	StdoutStderrOption<IsSync>,
+	StdoutStderrOption<IsSync>,
+	...Array<StdioOption<IsSync>>,
+];
+
+type StdioOptions<IsSync extends boolean = boolean> = BaseStdioOption | StdioOptionsArray<IsSync>;
 
 type EncodingOption =
   | 'utf8'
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,6 +1,15 @@
...
 import {type Readable, type Writable} from 'node:stream';
 
+type NoOutputStdioOption =
+	| 'ignore'
+	| 'inherit'
+	| 'ipc'
+	| number
+	| Readable
+	| Writable
+	| [NoOutputStdioOption];
```
ehmicky, line 11: `stdio` options which prevent Node.js from creating `childProcess.std*`, i.e. prevent us from retrieving the output.
Please note that when this does not apply when passing multiple values to this option. For example, `result.stdout` is `undefined` with `{ stdout: 'inherit' }` but not with `{ stdout: ['pipe', 'inherit'] }`. This PR accounts for this, and has a type test for it.

index.d.ts
```diff
@@ -61,10 +75,72 @@ type EncodingOption =
...
+type TupleItem<
+	Tuple extends readonly unknown[],
+	Index extends string,
+> = Tuple[Index extends keyof Tuple ? Index : number];
+
+// Whether `stdout|stderr: 'ignore', 'inherit'`, etc. is used
+type IgnoresStreamResult<
+	StreamIndex extends string,
+	OptionsType extends Options = Options,
+> = StreamIndex extends '0' ? false
```
ehmicky, line 87: We never return `stdin`.

index.d.ts
```diff
@@ -61,10 +75,72 @@ type EncodingOption =
...
+
+type MapStdioOptions<
+	StdioOptionsArrayType extends StdioOptionsArray,
+	OptionsType extends Options = Options,
+> = {
+	[StreamIndex in keyof StdioOptionsArrayType & string]: StreamIndex extends '0'
+		? undefined
+		: StreamIndex extends '1' | '2'
+			? StdioOutput<StreamIndex, OptionsType>
+			: StdioOutput<StreamIndex, OptionsType> | undefined
```
ehmicky, line 140: Unlike `result.stdio[1]` and `result.stdio[2]`, `result.stdio[3]` (and later ones) can be possible `undefined` when the file descriptor is used for input instead of output. In other words, when it behaves like `stdin` not like `stdout`/`stderr`. For example, this happens when passing an `Uint8Array` as input.

index.d.ts
```diff
@@ -419,21 +493,21 @@ export type ExecaReturnValue<IsSync extends boolean, StdoutStderrType extends St
 
 	This is `undefined` if the `stdout` option is set to [`'inherit'`, `'ipc'`, `'ignore'`, `Stream` or `integer`](https://nodejs.org/api/child_process.html#child_process_options_stdio).
 	*/
-	stdout: StdoutStderrType;
+	stdout: StdioOutput<'1', OptionsType>;
```
ehmicky, line 496: Due to how tuples are typed in TypeScript, we need to pass `'1'` not `1`.

PR #682, opened 2024-01-13
Limit breaking changes with types

ehmicky:
The `ExecaSyncReturnValue`, `ExecaSyncError` and `SyncOptions` types are for sync methods.

The `ExecaReturnValue`, `ExecaError` and `Options` types used to be for async methods. However, with #678, they are now for what's common between the async and sync methods. Based on your comment at https://github.com/sindresorhus/execa/pull/678#discussion_r1449871532, I reintroduced the sync methods types. This PR does the second step: to ensure those types remain for the async methods only, so that's it's not a breaking change. Also, it makes more sense, since most users need either the async types or the sync ones, not both.


ehmicky:
Fixed conflict.

Issue #675, opened 2024-01-11
Stricter typing for `childProcess.all`


ehmicky:
`childProcess.all` is currently always typed as `Readable | undefined`. 
We could make the types stricter to make it either `Readable` or `undefined` depending on Execa's arguments. Namely, it should be `undefined` only if either:
  - the `all` option is `false` (the default value)
  - both the `stdout` and `stderr` option is `inherit`, `ipc`, `ignore`, a Node.js stream or an integer. Or an array with a single item with one of those values.

ehmicky:
Started at #684.

ehmicky:
Finished at #689.

PR #684, opened 2024-01-13
Fix type of `result.all`

ehmicky:
Part of #675.

This ensures `result.all` is not typed as possibly `undefined` when `all: true` is used, and is typed as `undefined` otherwise.


sindresorhus:
Conflict

ehmicky:
Fixed.

ehmicky:
Fixed new conflict.

PR #686, opened 2024-01-14
Fix inconsistency between `result.stdout` and `error.stdout`

### LINKED ISSUE #674: Inconsistency between `result.stdout` and `error.stdout`.
We've been trying to keep the shape of an Execa's successful `result` and an `error` as close to each other as possible.

At the moment, `result.stdout` is:
  - `undefined` when: 
     - `childProcess.stdout` is `undefined`, e.g. when using `stdout: 'ignore'`
     - `buffer` is `false`
  - an empty string when there is an `stdout` but it is empty

Distinguishing between those two is useful. However, `error.stdout` does not do this: instead of `undefined`, it is an empty string.
We should fix this and make `error.stdout` behave like `result.stdout`.

This also applies to `result.stderr` and `result.all`.

We should also fix its types as suggested in #673.

ehmicky:
Fixes #674.

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,7 +81,7 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
 };
 
 const handleOutput = (options, value) => {
-	if (typeof value !== 'string' && !ArrayBuffer.isView(value)) {
+	if (value === undefined || value === null) {
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -81,7 +81,7 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
 };
 
 const handleOutput = (options, value) => {
-	if (typeof value !== 'string' && !ArrayBuffer.isView(value)) {
+	if (value === undefined || value === null) {
```
ehmicky, line 84: `value` is `null` only when using `execaSync()`, as this is what `child_process.spawnSync()` returns for `result.stdout|stderr` and `result.output[*]`.

PR #687, opened 2024-01-14
Improve types of `childProcess.stdout` and `childProcess.stderr`

ehmicky:
Follow-up on #681.

While #681 fixed the types of `result.stdout|stderr`, this PR fixes the types of `childProcess.stdout|stderr`.

`childProcess.stdout|stderr` is `null` when using options like `stdout: 'inherit'`. Otherwise it is a `Readable` stream.

This PR does not fix `childProcess.all` yet.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -617,7 +617,16 @@ export type KillOptions = {
 	forceKillAfterTimeout?: number | false;
 };
 
+type StreamUnlessIgnored<
+	StreamIndex extends string,
+	OptionsType extends Options = Options,
+> = IgnoresStreamResult<StreamIndex, OptionsType> extends true ? null : Readable;
+
 export type ExecaChildPromise<OptionsType extends Options = Options> = {
+	stdout: StreamUnlessIgnored<'1', OptionsType>;
+
+	stderr: StreamUnlessIgnored<'2', OptionsType>;
+
 	/**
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -617,7 +617,16 @@ export type KillOptions = {
...
 
+type StreamUnlessIgnored<
+	StreamIndex extends string,
+	OptionsType extends Options = Options,
+> = IgnoresStreamResult<StreamIndex, OptionsType> extends true ? null : Readable;
+
 export type ExecaChildPromise<OptionsType extends Options = Options> = {
+	stdout: StreamUnlessIgnored<'1', OptionsType>;
+
+	stderr: StreamUnlessIgnored<'2', OptionsType>;
```
ehmicky, line 628: Leaving the documentation blank as we do not document this property.
Since the final type uses `& ChildProcess`, it turns out `childProcess.stdout|stderr` still shows the TypeScript documentation for it coming from Node.js' `ChildProcess`.

index.d.ts
```diff
@@ -617,7 +617,16 @@ export type KillOptions = {
 	forceKillAfterTimeout?: number | false;
 };
 
+type StreamUnlessIgnored<
+	StreamIndex extends string,
+	OptionsType extends Options = Options,
+> = IgnoresStreamResult<StreamIndex, OptionsType> extends true ? null : Readable;
```
ehmicky, line 623: Note: `ChildProcess["stdout"]` type is `Readable | null`.

PR #693, opened 2024-01-16
Allow passing generators to the `std*` options

### LINKED ISSUE #20: Observable support with customizable through filters

### LINKED ISSUE #21: Allow transform streams convert output to an array of any objects, instead of a single string
We need to think through the ideal API for this and #20 as well, but something like:

``` js
execa('./cli', [], {
  transforms: {
     // shortcut to a built-in transform we provide
    stdout: 'line',

    stderr: through2(function (chunk, encoding, callback) {
      // custom processing here
    })
  }
}).then(function (result) {
  result.stdout;
  // => an array of lines

  result.stderr;
  // => an array of whatever the custom transform emits
});
```

### LINKED ISSUE #121: Ability to prefix child process stdio lines with a custom string
Proposal: add a new option `prefix` (optional, string). If given, the child's stdout and stderr streams would be transformed so **each line** is prefixed with the given string.

Use case: when running several child processes in parallel and using `inherit` to pipe each child's stdio to the parent, prefixing would make it easier to see what's going on:

```js
Promise.all(jobs.map((job, i) => execa(job.cmd, job.args, {
  stdio: 'inherit',
  prefix: `[job ${i + 1} of jobs.length] `,
}))).then(() => {
  // ...
});
```

```
[job 1 of 4] blah blah
[job 3 of 4] blah blah
[job 3 of 4] blah blah blah
[job 2 of 4] blah
...
```

I think the best API would be to have a `prefix` option which would apply to both stdout and stderr streams, and also `prefixStdout` and `prefixStderr` in case you want to override them individually.

Happy to implement this if interested.

ehmicky:
Fixes #20 (second most upvoted issue). Also fixes #21 and fixes #121 (third most upvoted issue).

This allows transforming a child process' input and ouput by passing an async generator to the `stdin`/`stdout`/`stderr`/`stdio` option.

```js
const transform = async function * (chunks) {
		yield chunk.toUpperCase();
	}
};

const {stdout} = await execa('echo', ['hello'], {stdout: transform});
console.log(stdout); // HELLO
```

Using generators is much simpler for users than creating web or Node.js `Transform` streams (which are used under-the-hood). Unlike regular functions, it still enables init and flushing logic (before and after the `for await` loop, which can be particularly helpful (and necessary for issues like #121).

IMHO There are many use cases for this. I have been needed this feature myself in production several times. For example, I've been needed to use `stdout: 'inherit'` but filtering secret values. This is now easy by using the `{ stdout: [filterSecretTransform, 'inherit'] }` option. I think the fact that those are among our most upvoted issues is telling.

This PR intentionally currently leaves out two points for the moment:
  - Splitting chunks line-wise. At the moment, a chunk is basically a single write, except for chunks >16KB which are split. Doing so would be required for #121.
  - Allowing generators to return or take as input other types than string and `Uint8Array`. Especially objects. Basically for the underlying stream to use `objectMode: true`. This is the issue at #21.

For a fuller description of the feature from a user standpoint, I have created the [following documentation](https://github.com/sindresorhus/execa/blob/transforms/docs/transform.md).

### DIFF
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -28,6 +28,8 @@ The transform can `yield` either a `string` or an `Uint8Array`, regardless of th
 `yield` can be called 0, 1 or multiple times. Not calling `yield` enables filtering a specific chunk.
 
 ```js
+import {execa} from 'execa';
+
 const transform = async function * (chunks) {
 	for await (const chunk of chunks) {
 		if (!chunk.includes('secret')) {
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -1,0 +1,55 @@
+# Transforms
+
+## Summary
+
+Transforms map or filter the input or output of a child process. They are defined by passing an [async generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*) to the [`stdin`](../readme.md#stdin), [`stdout`](../readme.md#stdout-1), [`stderr`](../readme.md#stderr-1) or [`stdio`](../readme.md#stdio-1) option.
+
+```js
+import {execa} from 'execa';
+
+const transform = async function * (chunks) {
+	for await (const chunk of chunks) {
+		yield chunk.toUpperCase();
+	}
+};
+
+const {stdout} = await execa('echo', ['hello'], {stdout: transform});
+console.log(stdout); // HELLO
+```
+
+## Encoding
+
+The `chunks` argument passed to the transform is an [`AsyncIterable<string>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols). If the [`encoding`](../readme.md#encoding) option is `buffer`, it is an `AsyncIterable<Uint8Array>` instead.
+
+The transform can `yield` either a `string` or an `Uint8Array`, regardless of the `chunks` argument's type.
+
+## Filtering
+
+`yield` can be called 0, 1 or multiple times. Not calling `yield` enables filtering a specific chunk.
+
+```js
+const transform = async function * (chunks) {
+	for await (const chunk of chunks) {
+		if (!chunk.includes('secret')) {
+			yield chunk;
+		}
+	}
+};
+
+const {stdout} = await execa('echo', ['This is a secret.'], {stdout: transform});
+console.log(stdout); // ''
+```
+
+## Combining
+
+The [`stdin`](../readme.md#stdin), [`stdout`](../readme.md#stdout-1), [`stderr`](../readme.md#stderr-1) and [`stdio`](../readme.md#stdio-1) options can accept an array of values. While this is not specific to transforms, this can be useful with them too. For example, the following transform impacts the value printed by `inherit`.
+
+```js
+await execa('echo', ['hello'], {stdout: [transform, 'inherit']});
+```
+
+This also allows using multiple transforms.
+
+```js
+await execa('echo', ['hello'], {stdout: [transform, otherTransform]});
+```
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,6 +25,8 @@ type CommonStdioOption<IsSync extends boolean = boolean> =
 	| undefined
 	| URL
 	| {file: string}
+	// TODO: Use either `Iterable<string>` or `Iterable<Uint8Array>` based on whether `encoding: 'buffer'` is used.
+	// See https://github.com/sindresorhus/execa/issues/694
 	| IfAsync<IsSync, ((chunks: Iterable<string | Uint8Array>) => AsyncGenerator<string | Uint8Array, void, void>)>;
 
 type InputStdioOption<IsSync extends boolean = boolean> =
--- a/index.d.ts
+++ b/index.d.ts
@@ -244,7 +244,7 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 
 	This can be an [array of values](https://github.com/sindresorhus/execa#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
-	This can also be an [async generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*) to transform the input. Please see the [full documentation here](https://github.com/sindresorhus/execa/tree/main/docs/transform.md).
+	This can also be an async generator function to transform the input. [Learn more.](https://github.com/sindresorhus/execa/tree/main/docs/transform.md)
 
 	@default `inherit` with `$`, `pipe` otherwise
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,13 +18,14 @@ type BaseStdioOption =
 	| 'ignore'
 	| 'inherit';
 
-type CommonStdioOption =
+type CommonStdioOption<IsSync extends boolean = boolean> =
 	| BaseStdioOption
 	| 'ipc'
 	| number
 	| undefined
 	| URL
-	| {file: string};
+	| {file: string}
+	| IfAsync<IsSync, ((chunks: Iterable<string | Uint8Array>) => AsyncGenerator<string | Uint8Array, void, void>)>;
 
 type InputStdioOption<IsSync extends boolean = boolean> =
 	| Uint8Array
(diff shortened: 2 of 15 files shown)


index.d.ts
```diff
@@ -18,13 +18,14 @@ type BaseStdioOption =
...
-type CommonStdioOption =
+type CommonStdioOption<IsSync extends boolean = boolean> =
 	| BaseStdioOption
 	| 'ipc'
 	| number
 	| undefined
 	| URL
-	| {file: string};
+	| {file: string}
+	| IfAsync<IsSync, ((chunks: Iterable<string | Uint8Array>) => AsyncGenerator<string | Uint8Array, void, void>)>;
```
ehmicky, line 28: Technically, `chunks` is `Iterable<Uint8Array>` if the `encoding` option is `buffer`, and `Iterable<string>`.

It turns out typing this was quite difficult, so I left it with this weaker typing for the time being.

lib/stdio/generator.js
```diff
@@ -0,0 +1,86 @@
...
+	const objectMode = false;
+	const highWaterMark = getDefaultHighWaterMark(objectMode);
+	const passThrough = new PassThrough({objectMode, highWaterMark, destroy: destroyPassThrough});
+	const iterable = passThrough.iterator();
+	const encodedIterable = applyEncoding(iterable, encoding);
+	const mappedIterable = value(encodedIterable);
+	const readableStream = Readable.from(mappedIterable, {objectMode, highWaterMark});
+	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
+	return {value: duplexStream};
+};
```
ehmicky, line 26: Description of what's going in here, because it is rather complex.

We return a `Duplex`, created by `Duplex.from()` made of a writable stream and a readable stream, piped to each other.

The writable stream is a simple `PassThrough`, so it does not do anything except forwarding data to the readable part.

The `PassThrough` is read as an iterable using `passThrough.iterator()`. 

This iterable is transformed to another iterable, by applying the encoding generators. Those convert the chunk type from `Buffer` to either `string` or `Uint8Array` depending on the `encoding` option.

This new iterable is transformed again to another one, this time by applying the user-supplied generator.

Finally, `Readable.from()` is used to convert this final iterable to a `Readable` stream.

I have carefully checked the following points, and added tests for it:
  - Error handling: the right errors are propagated. Streams end/abort/errors are propagated to other streams. Streams are properly destroyed.
  - Buffering with `highWaterMark`

index.d.ts
sindresorhus: Can you add a TODO comment so it won't be forgotten?

index.d.ts
```diff
@@ -241,6 +242,8 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 
 	This can be an [array of values](https://github.com/sindresorhus/execa#redirect-stdinstdoutstderr-to-multiple-destinations) such as `['inherit', 'pipe']` or `[filePath, 'pipe']`.
 
+	This can also be an [async generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*) to transform the input. Please see the [full documentation here](https://github.com/sindresorhus/execa/tree/main/docs/transform.md).
```
sindresorhus, line 245:
```suggestion
	This can also be an async generator function to transform the input. [Learn more.](https://github.com/sindresorhus/execa/tree/main/docs/transform.md)
```

?

Good to keep it short. We can also link to "async generator function" in the full docs.

lib/stdio/generator.js
```diff
@@ -0,0 +1,86 @@
...
+// Generators have a simple syntax, yet allows all of the following:
+//  - Sharing state between chunks, by using logic before the `for` loop
+//  - Flushing logic, by using logic after the `for` loop
+//  - Asynchronous logic
+//  - Emitting multiple chunks from a single source chunk, even if spaced in time, by using multiple `yield`
+//  - Filtering, by using no `yield`
+// Therefore, there is no need to allow Node.js or web transform streams.
+// The `highWaterMark` is kept as the default value, since this is what `childProcess.std*` uses.
+// We ensure `objectMode` is `false` for better buffering.
+// Chunks are currently processed serially. We could add a `concurrency` option to parallelize in the future.
```
sindresorhus, line 15: I would prefer a `/* */` comment as it makes it easier to edit it.

lib/stdio/generator.js
sindresorhus: The description here would make a code code comment for future readers of the code.

lib/stdio/generator.js
```diff
@@ -0,0 +1,86 @@
...
+	return {value: duplexStream};
+};
+
+// When an error is thrown in a generator, the PassThrough is aborted.
+// This creates a race condition for which error is propagated, due to the Duplex throwing twice:
+//  - The writable side is aborted (PassThrough)
+//  - The readable side propagate the generator's error
+// In order for the later to win that race, we need to wait one microtask.
+const destroyPassThrough = callbackify(async error => {
+	await setTimeout(0);
```
sindresorhus, line 34: Shouldn't this use `queueMicrotask`?

docs/transform.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Encoding
+
+The `chunks` argument passed to the transform is an [`AsyncIterable<string>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols). If the [`encoding`](../readme.md#encoding) option is `buffer`, it is an `AsyncIterable<Uint8Array>` instead.
+
+The transform can `yield` either a `string` or an `Uint8Array`, regardless of the `chunks` argument's type.
+
+## Filtering
+
```
sindresorhus, line 27: Users must be careful hereful here. Their secret could potentially be split over multiple chunks. This example would work better if the chunk was a line.

docs/transform.md
```diff
@@ -0,0 +1,55 @@
...
+`yield` can be called 0, 1 or multiple times. Not calling `yield` enables filtering a specific chunk.
+
+```js
+const transform = async function * (chunks) {
```
sindresorhus, line 31:
```suggestion
import {execa} from 'execa';

const transform = async function * (chunks) {
```

sindresorhus:
This is super cool and very useful 🎉

sindresorhus (changes requested):

lib/stdio/generator.js
ehmicky: Done in 71055df5c889f0370e55de9eb397529bb0bfb0cf :+1:

index.d.ts
ehmicky: Done in b760d9efdf2748e70683102028e5bbace66ce9e8 and also as a new issue in https://github.com/sindresorhus/execa/issues/694

docs/transform.md
ehmicky: Yes, absolutely, great catch!
I did think of it as well. I have an upcoming PR which will provide with line-wise splitting for the generators `chunks`.

lib/stdio/generator.js
ehmicky: Done in daf5b9c4185a473a74edf18e35c20d77b7410434

index.d.ts
ehmicky: Fixed the other instances in ff1d6785d3d45f302f97d014bb627efc2828febc

lib/stdio/generator.js
```diff
@@ -0,0 +1,100 @@
+import {Duplex, Readable, PassThrough, getDefaultHighWaterMark} from 'node:stream';
+import {callbackify} from 'node:util';
+import {setTimeout} from 'node:timers/promises';
+
+/*
+ * Generators can be used to transform/filter standard streams.
```
sindresorhus, line 6: Without the `*` prefixes.

lib/stdio/generator.js
ehmicky: Waiting for one microtask (`queueMicrotask()`) works.

I am waiting for one macrotask instead (`setTimeout()`) to be on the safe side, in case the Node.js core logic changes in the future. The race condition this is solving relies on multiple core Node.js functions' implementation, specifically: 
  - [`Readable.from()`](https://github.com/nodejs/node/blob/e133e5115a829aa7a445cfd8754df3a3dd586b0a/lib/internal/streams/from.js#L200)'s error handling
  - [`readable.iterator()`](https://github.com/nodejs/node/blob/e133e5115a829aa7a445cfd8754df3a3dd586b0a/lib/internal/streams/readable.js#L1394-L1403)'s error handling
  - [stream destroyer](https://github.com/nodejs/node/blob/e133e5115a829aa7a445cfd8754df3a3dd586b0a/lib/internal/streams/destroy.js#L325)

The first two rely on `async`/`await`, so if core Node.js added an additional `await` in that logic, it is possible that it would impact that race condition.

To clarify the underlying issue: when the user-supplied generator throws, two errors are propagated in a race condition:
  - The `Duplex` writable part (the `PassThrough`) is aborted, i.e. throw an `AbortError` that is very generic and unhelpful for users
  - The `Duplex` readable part (wrapped by `Readable.from()`) propagated the original error

So we are delaying the first part by adding some delay to the `PassThrough`'s `destroy()` method. This delays destroying the generator's underlying `PassThrough` slightly longer, but that should not be an issue.

I did add a unit test for this specific case. However, I think we might want to be safe and wait for one macrotask instead of a microtask? What do you think?

lib/stdio/generator.js
ehmicky: Fixed in 295f63a6922061e72db981cac2dacc3ff177b05c

ehmicky:
Ready for another review! :+1: 
That PR was lots of fun, and taught me a few things :)

lib/stdio/generator.js
sindresorhus: > In order for the later to win that race, we need to wait one microtask.

The code comment is incorrect then, since it says "microtask". I would expand a little bit on the comment (as succinctly as possible) there on why a macrosecond is used.

sindresorhus:
https://github.com/sindresorhus/execa/pull/693#discussion_r1453895599

ehmicky:
Fixed :+1:

PR #697, opened 2024-01-16
Add passing the same transform multiple times

ehmicky:
The `stdin`/`stdout`/`stderr`/`stdio` options allow passing an array of values. We skip duplicate values, since there is no good reason to pass duplicates there. For example, `{ stdout: [fileUrl, fileUrl] }` does not make sense, and would result in `stdout` being piped twice to the same file.

However, there is a potentially legitimate use case for duplicates: transforms. For example, the following one, if `thirdPartyTransform` might add secret values, but should not read any:

```js
await execa('...', { stdout: [filterSecretsTransform, thirdPartyTransform, filterSecretsTransform] })
```

This PR allows duplicates with transforms.

### DIFF
--- a/lib/stdio/handle.js
+++ b/lib/stdio/handle.js
@@ -21,17 +21,28 @@ export const handleInput = (addProperties, options) => {
 // For example, `stdout: ['ignore']` behaves the same as `stdout: 'ignore'`.
 const getStdioStreams = (stdioOption, index) => {
 	const optionName = getOptionName(index);
-	const stdioOptions = Array.isArray(stdioOption) ? [...new Set(stdioOption)] : [stdioOption];
-	const isStdioArray = stdioOptions.length > 1;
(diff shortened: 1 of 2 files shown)


lib/stdio/handle.js
```diff
@@ -21,40 +21,45 @@ export const handleInput = (addProperties, options) => {
 // For example, `stdout: ['ignore']` behaves the same as `stdout: 'ignore'`.
 const getStdioStreams = (stdioOption, index) => {
 	const optionName = getOptionName(index);
-	const stdioOptions = Array.isArray(stdioOption) ? [...new Set(stdioOption)] : [stdioOption];
```
ehmicky, line 24: This PR basically changes `[...new Set(stdioOption)]` to the `filterDuplicates()` method, which takes into account generators.

Since `filterDuplicates()` requires knowing the type of each `stdioOptions` item, the logic had to be shuffled around a little. It does not change the behavior otherwise.

PR #698, opened 2024-01-16
Separate generator -> Duplex utility

ehmicky:
This PR splits the following logic into their own file:
  - Utility converting an array of generators into a `Duplex`
  - Everything related to the `encoding` option

It does not change the code. It just moves lines of code around.

### DIFF
--- a/lib/stdio/duplex.js
+++ b/lib/stdio/duplex.js
@@ -1,0 +1,42 @@
+import {Duplex, Readable, PassThrough, getDefaultHighWaterMark} from 'node:stream';
+
+/*
+Transform an array of generator functions into a `Duplex`.
+The `Duplex` is created by `Duplex.from()` made of a writable stream and a readable stream, piped to each other.
+- The writable stream is a simple `PassThrough`, so it only forwards data to the readable part.
+- The `PassThrough` is read as an iterable using `passThrough.iterator()`.
+- This iterable is transformed to another iterable, by applying the encoding generators.
+	Those convert the chunk type from `Buffer` to `string | Uint8Array` depending on the encoding option.
+- This new iterable is transformed again to another one, this time by applying the user-supplied generator.
+- Finally, `Readable.from()` is used to convert this final iterable to a `Readable` stream.
+ */
+export const generatorsToDuplex = (generators, {objectMode}) => {
+	const highWaterMark = getDefaultHighWaterMark(objectMode);
+	const passThrough = new PassThrough({objectMode, highWaterMark, destroy: destroyPassThrough});
+	let iterable = passThrough.iterator();
+
+	for (const generator of generators) {
+		iterable = generator(iterable);
+	}
+
+	const readableStream = Readable.from(iterable, {objectMode, highWaterMark});
+	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
+	return duplexStream;
+};
+
+/*
+When an error is thrown in a generator, the PassThrough is aborted.
+
+This creates a race condition for which error is propagated, due to the Duplex throwing twice:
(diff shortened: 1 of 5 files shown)


lib/stdio/duplex.js
```diff
@@ -0,0 +1,42 @@
...
+	let iterable = passThrough.iterator();
+
+	for (const generator of generators) {
+		iterable = generator(iterable);
+	}
+
+	const readableStream = Readable.from(iterable, {objectMode, highWaterMark});
+	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
+	return duplexStream;
+};
```
ehmicky, line 25: IMHO that's actually a fairly useful utility on its own, that might deserve its own package? It also does not have any external dependencies and is pretty much self-contained.

It relies on Node.js streams, although web streams can be produced too using [`Duplex.toWeb()`](https://nodejs.org/api/stream.html#streamduplextowebstreamduplex).

That could be useful for anyone wanting to transform streams but is too scared of using them. Generators are pretty straightforward.

lib/stdio/generator.js
```diff
@@ -17,79 +18,12 @@ The `highWaterMark` is kept as the default value, since this is what `childProce
...
-	const objectMode = false;
-	const highWaterMark = getDefaultHighWaterMark(objectMode);
-	const passThrough = new PassThrough({objectMode, highWaterMark, destroy: destroyPassThrough});
-	const iterable = passThrough.iterator();
-	const encodedIterable = applyEncoding(iterable, encoding);
-	const mappedIterable = value(encodedIterable);
-	const readableStream = Readable.from(mappedIterable, {objectMode, highWaterMark});
-	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
+export const generatorToDuplexStream = ({value, encoding}) => {
+	const duplexStream = generatorsToDuplex([getEncodingStartGenerator(encoding), value], {objectMode: false});
```
ehmicky, line 23: Hopefully, this makes understanding this function easier. All we are doing is:
  - Apply the encoding generator
  - Apply the user-supplied generator
  - Convert this into a `Duplex` stream

lib/stdio/generator.js
```diff
@@ -17,79 +18,13 @@ The `highWaterMark` is kept as the default value, since this is what `childProce
...
-	const highWaterMark = getDefaultHighWaterMark(objectMode);
-	const passThrough = new PassThrough({objectMode, highWaterMark, destroy: destroyPassThrough});
-	const iterable = passThrough.iterator();
-	const encodedIterable = applyEncoding(iterable, encoding);
-	const mappedIterable = value(encodedIterable);
-	const readableStream = Readable.from(mappedIterable, {objectMode, highWaterMark});
-	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
+export const generatorToDuplexStream = ({value, encoding}) => {
+	const generators = [getEncodingStartGenerator(encoding), value];
+	const duplexStream = generatorsToDuplex(generators, {objectMode: false});
```
ehmicky, line 24: Hopefully, this makes this function easier to understand now. 

All we are doing is:
  - Apply the encoding generator, to ensure each chunk is either `string` or `Uint8Array`
  - Apply the user-supplied generator
  - Convert this into a `Duplex` stream

lib/stdio/duplex.js
sindresorhus: > IMHO that's actually a fairly useful utility on its own, that might deserve its own package? 🤔

Yeah, potentially. I think it would be good to let the code mature a bit here first though.

Also not clear to me how this utility is different from `stream.compose`. Might be worth documenting that.

PR #699, opened 2024-01-17
Line-wise transforms

### LINKED ISSUE #121: Ability to prefix child process stdio lines with a custom string

### LINKED ISSUE #695: Line-wise transforms
## Background

This feature was suggested by both #21 and #121.
Also mentioned by @sindresorhus in https://github.com/sindresorhus/execa/pull/693#discussion_r1453380836.

## Rationale

Users can now transform `stdin`/`stdout`/`stderr`/`stdio[*]` by passing generators to those options.

The vast majority of processes' input and output is text-based. When transforming text-based data, it is very convenient to map one line at a time. For example, this is required for #121.

Currently, each chunk being transformed might be broken in the middle of a line. For example, the following does not currently work because the word `secret` might be split into 2 different chunks.

```js
const transform = async function * (chunks) {
		if (!chunk.includes('secret')) {
			yield chunk;
		}
	}
};

const {stdout} = await execa('echo', ['This is a secret.'], {stdout: transform});
```

## Solution

Transforms should iterate over chunks line by line. We would use `\n` as an end delimiter (which works for `\r\n` too).

However, I think we should still allow users to opt out of this behavior because, although less common, some processes' input/output is in fact binary. Splitting binary input/output line by line would result in chunks of varying size, some potentially very large if the binary input/output does not contain any newline. This would take away the main benefits of streaming (incremental CPU load + low memory).

## API

The users would get this behavior by default:

```js
execa('...', {stdout: transform});
```

To allow users to opt-out, I would suggest the following format.

```js
execa('...', {stdout: { transform, binary: true }});
```

In other words, a transform can be either `AsyncGeneratorFunction` or `{ transform: AsyncGeneractorFunction, binary: boolean }`. 

This format is also extensible to other options.

## Alternatives

One other possible API would be:

```js
execa('...', {stdout: ['binary', transform]});
```

However, some users might want to use multiple transforms for `stdout`, with some of them being binary, others not. For example, one transform might convert a binary `stdout` to a text-based one. Then a second transform might operate line-wise. The above approach would not allow this.

```js
execa('...', {stdout: ['binary', binaryToTextTransform, secretFilterTransform]});
```

---

Yet another possible API would be:

```js
execa('...', {stdout: transform, binary: true});
```

However, this has the same problem. It also has an additional one: it would apply to all streams (`stdin`/`stdout`/`stderr`/`stdio[*]`). For example, some processes might produce a binary `stdout` but a text-based `stderr`.

```js
execa('...', {stdout: binaryTransform, stderr: textTransform, binary: true});
```

---

Implementing this should be rather simple, since splitting chunks line-wise can be done using the same kind of generator users can now use themselves. In other words, we just need to add an internal generator, like we already do for encoding the chunks.

https://github.com/sindresorhus/execa/blob/e8f6bbd65547a3b19d001fbe4c34d7ae140a1d33/lib/stdio/generator.js#L74-L85

ehmicky:
Fixes #695  
Fixes #121

This makes transforms operate line-wise by default, unless `{ transform, binary: true }` is used.

### DIFF
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -53,8 +53,8 @@ await execa('./binary.js', {stdout: {transform, binary: true}});
 ```
 
 This is more efficient and recommended if the data is either:
-  - Binary, i.e. it does not have lines.
-	- Text, but the transform works even if a line or word is split across multiple chunks.
+	- Binary: Which does not have lines.
+	- Text: But the transform works even if a line or word is split across multiple chunks.
 
 ## Combining
 
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -7,9 +7,10 @@ Transforms map or filter the input or output of a child process. They are define
 ```js
 import {execa} from 'execa';
 
-const transform = async function * (chunks) {
-	for await (const chunk of chunks) {
-		yield chunk.toUpperCase();
+const transform = async function * (lines) {
+	for await (const line of lines) {
+		const prefix = line.includes('error') ? 'ERROR' : 'INFO'
+		yield `${prefix}: ${line}`
 	}
 };
 
@@ -17,23 +18,29 @@ const {stdout} = await execa('echo', ['hello'], {stdout: transform});
 console.log(stdout); // HELLO
 ```
 
+Alternatively, a `{ transform }` plain object can also be passed.
+
+```js
+await execa('echo', ['hello'], {stdout: {transform}});
+```
+
 ## Encoding
 
-The `chunks` argument passed to the transform is an [`AsyncIterable<string>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols). If the [`encoding`](../readme.md#encoding) option is `buffer`, it is an `AsyncIterable<Uint8Array>` instead.
+The `lines` argument passed to the transform is an [`AsyncIterable<string>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols). If the [`encoding`](../readme.md#encoding) option is `buffer`, it is an `AsyncIterable<Uint8Array>` instead.
 
-The transform can `yield` either a `string` or an `Uint8Array`, regardless of the `chunks` argument's type.
+The transform can `yield` either a `string` or an `Uint8Array`, regardless of the `lines` argument's type.
 
 ## Filtering
 
-`yield` can be called 0, 1 or multiple times. Not calling `yield` enables filtering a specific chunk.
+`yield` can be called 0, 1 or multiple times. Not calling `yield` enables filtering a specific line.
 
 ```js
 import {execa} from 'execa';
 
-const transform = async function * (chunks) {
-	for await (const chunk of chunks) {
-		if (!chunk.includes('secret')) {
-			yield chunk;
+const transform = async function * (lines) {
+	for await (const line of lines) {
+		if (!line.includes('secret')) {
+			yield line;
 		}
 	}
 };
@@ -41,6 +48,19 @@ const transform = async function * (chunks) {
 const {stdout} = await execa('echo', ['This is a secret.'], {stdout: transform});
 console.log(stdout); // ''
 ```
+
+## Binary data
+
+The transform iterates over lines by default.\
+However, if the `binary` transform option is `true`, it iterates over arbitrary chunks of data instead.
+
+```js
+await execa('./binary.js', {stdout: {transform, binary: true}});
+```
+
+This is more efficient and recommended if the data is either:
+  - Binary, i.e. it does not have lines.
+	- Text, but the transform works even if a line or word is split across multiple chunks.
 
 ## Combining
 
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -18,12 +18,6 @@ const {stdout} = await execa('echo', ['hello'], {stdout: transform});
 console.log(stdout); // HELLO
 ```
 
-Alternatively, a `{ transform }` plain object can also be passed.
-
-```js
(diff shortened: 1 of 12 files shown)


lib/stdio/encoding.js
```diff
@@ -7,8 +7,16 @@ export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
...
 
-	const value = encodingEndGenerator.bind(undefined, encoding);
-	return [...stdioStreams, {...stdioStreams[0], type: 'generator', value, encoding: 'buffer'}];
+	const transform = encodingEndGenerator.bind(undefined, encoding);
+	return [
+		...stdioStreams,
+		{
+			...stdioStreams[0],
+			type: 'generator',
+			value: {transform, binary: true},
```
ehmicky, line 16: This is the transform that we internally use for the `encoding: 'hex | base64'` option. It does not need to operate line-wise, so we make opt-out since this is more efficient.

lib/stdio/lines.js
```diff
@@ -0,0 +1,57 @@
...
+	return chunk;
+};
+
+const linesStringGenerator = async function * (chunks) {
+	yield * linesGenerator(chunks, '', '\n', concatString);
+};
+
+const concatString = (firstChunk, secondChunk) => `${firstChunk}${secondChunk}`;
+
+// This imperative logic is much faster than using `String.split()` and uses very low memory.
```
ehmicky, line 27: Quick benchmark on my machine was showing this code to be ~6 times faster than `String.split()`.

test/stdio/lines.js
```diff
@@ -0,0 +1,94 @@
...
+			resultGenerator.bind(undefined, lines),
+		]),
+		encoding: isUint8Array ? 'buffer' : 'utf8',
+		stripFinalNewline: false,
+	});
+	t.is(uint8ArrayToString(stdio[index], isUint8Array), input.join(''));
+	t.deepEqual(lines, stringsToUint8Arrays(expectedLines, isUint8Array));
+};
+
+test('Split string stdout - n newlines, 1 chunk', testLines, 1, ['aaa\nbbb\nccc'], ['aaa\n', 'bbb\n', 'ccc'], false);
```
ehmicky, line 53: With those tests, we pass a sets of chunks as `input`, and compare with how it gets split into `expectedLines`.

Under the hood, a first generator emits the `input` chunks. A second generator records each line that it gets as argument. Those `lines` are then compared with `expectedLines`.

test/stdio/lines.js
```diff
@@ -0,0 +1,94 @@
...
+const bigLine = '.'.repeat(1e5);
+const manyChunks = Array.from({length: 1e5}).fill('.');
+
+const inputGenerator = async function * (input, chunks) {
+	// eslint-disable-next-line no-unused-vars
+	for await (const chunk of chunks) {
+		for (const inputItem of input) {
+			yield inputItem;
+			// eslint-disable-next-line no-await-in-loop
+			await scheduler.yield();
```
ehmicky, line 18: This line ensures that each `yield` emits its own chunk of data. Node.js would otherwise merge those into a single one, for performance reasons.

docs/transform.md
```diff
@@ -7,33 +7,40 @@ Transforms map or filter the input or output of a child process. They are define
...
+		const prefix = line.includes('error') ? 'ERROR' : 'INFO'
+		yield `${prefix}: ${line}`
 	}
 };
 
 const {stdout} = await execa('echo', ['hello'], {stdout: transform});
 console.log(stdout); // HELLO
 ```
 
+Alternatively, a `{ transform }` plain object can also be passed.
```
sindresorhus, line 21: I don't see the point in documenting this. It's only there for the `binary` option and that option is already documented further down.

docs/transform.md
```diff
@@ -42,6 +49,19 @@ const {stdout} = await execa('echo', ['This is a secret.'], {stdout: transform})
...
+The transform iterates over lines by default.\
+However, if the `binary` transform option is `true`, it iterates over arbitrary chunks of data instead.
+
+```js
+await execa('./binary.js', {stdout: {transform, binary: true}});
+```
+
+This is more efficient and recommended if the data is either:
+  - Binary, i.e. it does not have lines.
+	- Text, but the transform works even if a line or word is split across multiple chunks.
```
sindresorhus, line 63:
```suggestion
	- Binary: Which does not have lines.
	- Text: But the transform works even if a line or word is split across multiple chunks.
```

sindresorhus (changes requested):

sindresorhus:
> Fixes https://github.com/sindresorhus/execa/issues/695 and https://github.com/sindresorhus/execa/issues/121.

FYI, you need to write `fixes` in front of each number.

docs/transform.md
ehmicky: Yes, I was hesitating too whether to include this.
Fixed in 51881563dec1a20ad9d336d0c20c49908c887121

ehmicky:
Added to `Why` in aef33b7130c3ed08633604727de1eb18546fbb8d.

Thanks for the tip with "Fixes" in GitHub! :ok_hand:

PR #705, opened 2024-01-20
Improve `error.message`

ehmicky:
This PR slightly improves how `error.message` is built. It also improves the related tests.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -1,5 +1,6 @@
 import process from 'node:process';
 import {signalsByName} from 'human-signals';
+import stripFinalNewline from 'strip-final-newline';
 import {isUint8Array} from './stdio/utils.js';
 
 const getErrorPrefix = ({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled}) => {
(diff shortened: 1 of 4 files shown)


lib/error.js
```diff
@@ -55,8 +64,8 @@ export const makeError = ({
 	const isError = Object.prototype.toString.call(error) === '[object Error]';
 	const shortMessage = isError ? `${execaMessage}\n${error.message}` : execaMessage;
 	const message = [shortMessage, stdio[2], stdio[1], ...stdio.slice(3)]
+		.map(messagePart => stripFinalNewline(serializeMessagePart(messagePart)))
```
ehmicky, line 67: We are joining each part with newlines. If a part ends with a newline, it just results in a single newline, not two.

lib/error.js
```diff
@@ -55,8 +64,8 @@ export const makeError = ({
...
 		.filter(Boolean)
-		.map(messagePart => serializeMessagePart(messagePart))
 		.join('\n');
```
ehmicky, line 69: I wonder whether we should separate the error message, stdout and stderr with an empty line or not?

```
Error: this failed

stdout
...

stderr
...
```

vs (current behavior)

```
Error: this failed
stdout
stderr
```

The first separates each part in a clearer way. But the second one is more compact. :thinking: 

Which one do you prefer?

sindresorhus: The first one

ehmicky: Done :+1:

lib/error.js
```diff
@@ -54,10 +63,11 @@ export const makeError = ({
 	const execaMessage = `Command ${prefix}: ${command}`;
 	const isError = Object.prototype.toString.call(error) === '[object Error]';
 	const shortMessage = isError ? `${execaMessage}\n${error.message}` : execaMessage;
-	const message = [shortMessage, stdio[2], stdio[1], ...stdio.slice(3)]
+	const messageStdio = all === undefined ? [stdio[2], stdio[1]] : [all];
```
ehmicky, line 66: I have just added this line too: if `all: true` is used, the error message should show `result.all` instead, since it is intertwined, so it is more useful for debugging that splitting `stdout` and `stderr`.

Issue #706, opened 2024-01-20
Line-wise standard streams


ehmicky:
With #695, we just added line-wise transforms. However, this does not cover `childProcess.std*`.

```js
const childProcess = execa('...')
for await (const chunk of childProcess.stdout) {
  console.log(chunk) // Buffer, and not line-wise
}
```

What are your thoughts on adding the following?

```js
const childProcess = execa('...', {lines: true})
for await (const line of childProcess.stdout) {
  console.log(line) // String, and line-wise
}
```

This would work also for `childProcess.stdout.read()`, `childProcess.stdout.on('data')` and when redirecting from/to Node.js/web streams.

```js
const writableStream = new Writable({
  write(line, encoding, done) {
    console.log(line) // String, and line-wise
    done()
  }
})
await execa('...', {lines: true, stdout: writableStream})
```

This would be pretty fairly simple to implement since it can re-use most of the logic of 2 separate features:
  - Line-wise transforms (#695)
  - `objectMode` transform (#21)

Under the hood, when the `lines: true` option is used, we would just add an internal transform that is a noop, but is both line-wise and in `objectMode`.

sindresorhus:
Yes, this would be nice 👍

PR #709, opened 2024-01-21
Allow any type with transforms

### LINKED ISSUE #21: Allow transform streams convert output to an array of any objects, instead of a single string

ehmicky:
Fixes #21. That's our oldest open issue (almost 8 years old)!
This is also a requirement towards fixing #706.

This allows using any type with transforms by using `{transform, objectMode: true}`. The process' return `stdout`/`stderr` will be an array. Typically, this is an array of objects, but this can be any type, including strings (which is very useful for #706).

```js
const transform = async function * (lines) {
	for await (const line of lines) {
		yield JSON.parse(line)
	}
}

const {stdout} = await execa('./jsonlines-output.js', {stdout: {transform, objectMode: true}});
for (const data of stdout) {
	console.log(stdout) // {...}
}
```

This also works for `stdin`. Users can pass objects to `stdin` with `childProcess.stdin.write()`. They can also directly pass an iterable of objects to the `stdin` option.

```js
const transform = async function * (lines) {
	for await (const line of lines) {
		yield JSON.stringify(line)
	}
}

const input = [{event: 'example'}, {event: 'otherExample'}]
await execa('./jsonlines-input.js', {stdin: [input, {transform, objectMode: true}]});
```

While this is a big PR, most of it are code quality related (tests, types, documentation). The runtime part is actually not as big. 

When it comes to the tests, they mostly repeat existing tests, but using object mode instead. I tried to make sure every edge case was tested.

This PR only has runtime changes for users of transforms with `objectMode: true`. I.e. it has no behavior changes for current users.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,9 +18,15 @@ type BaseStdioOption =
 	| 'ignore'
 	| 'inherit';
 
-// @todo Use either `Iterable<string>` or `Iterable<Uint8Array>` based on whether `encoding: 'buffer'` is used.
+// @todo Use `string`, `Uint8Array` or `unknown` for both the argument and the return type, based on whether `encoding: 'buffer'` and `objectMode: true` are used.
 // See https://github.com/sindresorhus/execa/issues/694
-type StdioTransform = ((chunks: Iterable<string | Uint8Array>) => AsyncGenerator<string | Uint8Array, void, void>);
+type StdioTransform = ((chunks: Iterable<unknown>) => AsyncGenerator<unknown, void, void>);
+
+type StdioTransformFull = {
+	transform: StdioTransform;
+	binary?: boolean;
+	objectMode?: boolean;
+};
 
 type CommonStdioOption<IsSync extends boolean = boolean> =
 	| BaseStdioOption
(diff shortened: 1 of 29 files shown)


index.d.ts
```diff
@@ -18,9 +18,15 @@ type BaseStdioOption =
 	| 'ignore'
 	| 'inherit';
 
-// @todo Use either `Iterable<string>` or `Iterable<Uint8Array>` based on whether `encoding: 'buffer'` is used.
+// @todo Use `string`, `Uint8Array` or `unknown` for both the argument and the return type, based on whether `encoding: 'buffer'` and `objectMode: true` are used.
 // See https://github.com/sindresorhus/execa/issues/694
```
ehmicky, line 22: I will update that issue accordingly once this PR is merged.

sindresorhus:
Conflict

ehmicky:
Fixed :+1:

PR #710, opened 2024-01-22
Improve documentation of the `extendEnv` and `env` options

### LINKED ISSUE #390: Ambiguity in env/extendEnv documentation
The docs:

For `extendEnv`

> Set to false if you don't want to extend the environment variables when providing the env property.

For `env`:

> Environment key-value pairs. Extends automatically from process.env. Set extendEnv to false if you don't want this.

Having read this, I was under the impression that if you set `extendEnv` to `false`, `env` has no effect at all (which is exactly the behaviour I needed).

Here is the code:
```js
const env = extendEnv ? {...process.env, ...envOption} : envOption;
```

So the correction:

For `extendEnv`:

> If set to true, the existing `process.env` to be extended by the `env` property, and the result will be passed to the child process. If set to false, only the `env` object will be passed to the child process.

For `env`:

> Environment key-value pairs. By default, process.env is extended by this object, and the result is given to the child process. When extendEnv is set to false, only this object will be given to the child process..

ehmicky:
Fixes #390.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -391,7 +391,7 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 
 	/**
 	If `true`, the child process uses both the `env` option and the current process' environment variables ([`process.env`](https://nodejs.org/api/process.html#processenv)).
-	If `false`, it uses only the `env` option is used, not `process.env`.
+	If `false`, only the `env` option is used, not `process.env`.
 
 	@default true
 	*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -390,7 +390,8 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 	readonly stripFinalNewline?: boolean;
 
 	/**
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -335,7 +335,8 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 	readonly stripFinalNewline?: boolean;
 
 	/**
-	Set to `false` if you don't want to extend the environment variables when providing the `env` property.
+	If `true`, the child process uses both the `env` option and the current process' environment variables ([`process.env`](https://nodejs.org/api/process.html#processenv)).
+	If `false`, it uses only the `env` option is used, not `process.env`.
```
sindresorhus, line 339: typo
```suggestion
	If `false`, only the `env` option is used, not `process.env`.
```

sindresorhus (changes requested):

ehmicky:
Fixed, thanks :+1:

PR #712, opened 2024-01-22
Fix passing `undefined` options

ehmicky:
Right now, if options are passed with an `undefined` value, it overrides the default value. The default value should be used instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,25 +49,8 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
 
 	const {command: file, args, options: initialOptions} = crossSpawn._parse(filePath, rawArgs, rawOptions);
 
-	const options = {
-		maxBuffer: DEFAULT_MAX_BUFFER,
-		buffer: true,
-		stripFinalNewline: true,
-		extendEnv: true,
-		preferLocal: false,
-		localDir: initialOptions.cwd || process.cwd(),
-		execPath: process.execPath,
-		encoding: 'utf8',
-		reject: true,
-		cleanup: true,
-		all: false,
-		windowsHide: true,
-		verbose: verboseDefault,
-		killSignal: 'SIGTERM',
-		...initialOptions,
-		shell: normalizeFileUrl(initialOptions.shell),
-	};
-
+	const options = addDefaultOptions(initialOptions);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -49,25 +49,8 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
...
-		localDir: initialOptions.cwd || process.cwd(),
-		execPath: process.execPath,
-		encoding: 'utf8',
-		reject: true,
-		cleanup: true,
-		all: false,
-		windowsHide: true,
-		verbose: verboseDefault,
-		killSignal: 'SIGTERM',
-		...initialOptions,
```
ehmicky, line 67: Spreading `undefined` values overrides the default values above.

index.js
```diff
@@ -80,6 +63,42 @@ const handleArguments = (rawFile, rawArgs, rawOptions = {}) => {
...
+	localDir = cwd,
+	execPath = process.execPath,
+	encoding = 'utf8',
+	reject = true,
+	cleanup = true,
+	all = false,
+	windowsHide = true,
+	verbose = verboseDefault,
+	killSignal = 'SIGTERM',
+	...options
```
ehmicky, line 82: But this works.

test/test.js
```diff
@@ -76,20 +76,26 @@ test('skip throwing when using reject option in sync mode', t => {
 
 const testStripFinalNewline = async (t, index, stripFinalNewline, execaCommand) => {
 	const {stdio} = await execaCommand('noop-fd.js', [`${index}`, 'foobar\n'], {...fullStdio, stripFinalNewline});
-	t.is(stdio[index], `foobar${stripFinalNewline ? '' : '\n'}`);
+	t.is(stdio[index], `foobar${stripFinalNewline === false ? '\n' : ''}`);
 };
 
-test('stripFinalNewline: true with stdout', testStripFinalNewline, 1, undefined, execa);
+test('stripFinalNewline: undefined with stdout', testStripFinalNewline, 1, undefined, execa);
+test('stripFinalNewline: true with stdout', testStripFinalNewline, 1, true, execa);
```
ehmicky, line 83: Default value is `true`, so this tests that default values are correctly assigned.

PR #714, opened 2024-01-22
Improve graceful exit

### LINKED ISSUE #511: Allow the ‘forceKillAfterTimeout’ option to be passed to ‘.cancel()’
The [documentation](https://github.com/sindresorhus/execa/blob/7a8f5cd3e900416a158d0e1b47dceaf6983862ac/index.d.ts#L444) says we should prefer `.cancel()` over `.kill()`, so it would be nice for `.cancel()` to accept the `forceKillAfterTimeout` option that `.kill()` has.

ehmicky:
Fixes #511.

This extends graceful exits to the following termination methods:
  - `signal` option
  - `cleanup` option
  - `timeout` option
  - `maxBuffer` option
  - stream errors
  - `childProcess.on('error')` event

This moves the `forceKillAfterTimeout` option from the `.kill()` method to the top-level options instead.


index.d.ts
```diff
@@ -779,11 +788,6 @@ export type ExecaChildPromise<OptionsType extends Options = Options> = {
 		onRejected?: (reason: ExecaError<OptionsType>) => ResultType | PromiseLike<ResultType>
 	): Promise<ExecaReturnValue<OptionsType> | ResultType>;
 
-	/**
-	Same as the original [`child_process#kill()`](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal), except if `signal` is `SIGTERM` (the default value) and the child process is not terminated after 5 seconds, force it by sending `SIGKILL`. Note that this graceful termination does not work on Windows, because Windows [doesn't support signals](https://nodejs.org/api/process.html#process_signal_events) (`SIGKILL` and `SIGTERM` has the same effect of force-killing the process immediately.) If you want to achieve graceful termination on Windows, you have to use other means, such as [`taskkill`](https://github.com/sindresorhus/taskkill).
-	*/
-	kill(signal?: string, options?: KillOptions): void;
```
ehmicky, line 785: `process.kill()` now has the same signature as in core Node.js.

The only difference is the graceful exit behavior, which we document with the `forceKillAfterTimeout` option, so there's no need to document `process.kill()` here anymore.

readme.md
```diff
@@ -221,20 +221,6 @@ try {
...
-
-```js
-const subprocess = execa('node');
-
-setTimeout(() => {
-	subprocess.kill('SIGTERM', {
-		forceKillAfterTimeout: 2000
-	});
-}, 1000);
-```
```
ehmicky, line 236: The graceful exit behavior happens by default. It's probably not critical to add an example for it anymore.

test/fixtures/no-killable.js
```diff
@@ -6,6 +6,7 @@ process.on('SIGTERM', () => {
 });
 
 process.send('');
+console.log('.');
```
ehmicky, line 9: Needed for the `maxBuffer` test.

readme.md
```diff
@@ -783,6 +754,26 @@ Default: `SIGTERM`
 
 Signal value to be used when the spawned process will be killed.
 
+#### forceKillAfterTimeout
```
ehmicky, line 757: Since we're moving the option to the top-level, should this take this opportunity to rename it to `gracefulExit`? I am wondering whether `forceKillAfterTimeout` might lead to some confusion with the other `timeout` option which is quite different.

lib/kill.js
```diff
@@ -1,34 +1,36 @@
...
-	if (timeout === undefined) {
+const setKillTimeout = async ({kill, signal, forceKillAfterTimeout, killResult, controller}) => {
+	if (!shouldForceKill(signal, forceKillAfterTimeout, killResult)) {
 		return;
 	}
 
-	await pSetTimeout(timeout, undefined, {ref: false});
-	kill('SIGKILL');
+	try {
+		await setTimeout(forceKillAfterTimeout, undefined, {signal: controller.signal});
```
ehmicky, line 20: The controller is aborted when the child process exits (successfully or not). When this happens, `setTimeout()` throws, hence the empty `catch` block.

index.d.ts
```diff
@@ -468,6 +468,26 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 	*/
 	readonly killSignal?: string | number;
 
+	/**
+	If the child process is terminated but does not exit, forcefully exit it by sending [`SIGKILL`](https://en.wikipedia.org/wiki/Signal_(IPC)#SIGKILL).
+
+	The graceful period is 5 seconds by default. This feature can be disabled with `false`.
```
sindresorhus, line 474:
```suggestion
	The grace period is 5 seconds by default. This feature can be disabled with `false`.
```

readme.md
```diff
@@ -50,7 +50,7 @@ This package improves [`child_process`](https://nodejs.org/api/child_process.htm
 - Redirect [`stdin`](#stdin)/[`stdout`](#stdout-1)/[`stderr`](#stderr-1) from/to files, streams, iterables, strings, `Uint8Array` or [objects](docs/transform.md#object-mode).
 - [Transform](docs/transform.md) `stdin`/`stdout`/`stderr` with simple functions.
 - Iterate over [each text line](docs/transform.md#binary-data) output by the process.
-- [Graceful termination](#optionsforcekillaftertimeout).
+- [Graceful termination](#forcekillaftertimeout).
```
sindresorhus, line 53: We should use a better word than "graceful" as it can be misunderstood.

sindresorhus, line 53: Maybe:

```suggestion
- ["Fail-safe process termination](#forcekillaftertimeout).
```

sindresorhus (changes requested):

readme.md
ehmicky: Done in 731208291c2d153ba51fe4f1bfe0ee2ced8bbc4b

PR #724, opened 2024-01-24
Improve process termination

### LINKED ISSUE #719: `error.isTerminated` and `error.signal` are sometimes invalid
When the `maxBuffer` option is set and is hit, Execa calls `childProcess.kill()`.
When this happens, `error.isTerminated` is `false` and `error.signal` is `undefined`, which is incorrect.
`error.isTerminated` and `error.signal` describe whether the process has been terminated by a signal. They should be set regardless of whether that signal termination was done by the user or by Execa itself.

This bug also applies to the `cleanup` option, `signal` option, and `error` events on the child process or its streams.
Inconsistently, this bug does not apply to the `timeout` option.

ehmicky:
Fixes #719.

### DIFF
--- a/lib/kill.js
+++ b/lib/kill.js
@@ -47,7 +47,7 @@ export const normalizeForceKillAfterTimeout = forceKillAfterTimeout => {
 const killAfterTimeout = async ({spawned, timeout, killSignal, context, controller}) => {
 	await setTimeout(timeout, undefined, {signal: controller.signal});
 	spawned.kill(killSignal);
-	Object.assign(context, {timedOut: true, signal: killSignal});
+	context.timedOut = true;
 	throw new Error('Timed out');
 };
 
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -105,6 +105,29 @@ const cleanupStdioStreams = (customStreams, error) => {
 	}
 };
 
+// Like `once()` except it never rejects, especially not on `error` event.
+const pEvent = (eventEmitter, eventName) => new Promise(resolve => {
+	eventEmitter.once(eventName, (...payload) => {
+		resolve([eventName, ...payload]);
+	});
+});
+
+const throwOnProcessError = async processErrorPromise => {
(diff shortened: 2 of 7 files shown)


lib/kill.js
```diff
@@ -47,7 +47,7 @@ export const normalizeForceKillAfterTimeout = forceKillAfterTimeout => {
 const killAfterTimeout = async ({spawned, timeout, killSignal, context, controller}) => {
 	await setTimeout(timeout, undefined, {signal: controller.signal});
 	spawned.kill(killSignal);
-	Object.assign(context, {timedOut: true, signal: killSignal});
```
ehmicky, line 50: `context.signal` is not needed. When calling `spawned.kill(killSignal)`, we then wait for the process' `exit` event, which will return that signal.

lib/stream.js
```diff
@@ -105,6 +105,29 @@ const cleanupStdioStreams = (customStreams, error) => {
...
+};
+
+// First the `spawn` event is emitted, then `exit`.
+// If the `error` event is emitted:
+//  - before `spawn`: `exit` is never emitted.
+//  - after `spawn`: `exit` is always emitted.
+// We only want to listen to `exit` if it will be emitted, i.e. if `spawn` has been emitted.
+// Therefore, the arguments order of `Promise.race()` is significant.
+const waitForFailedProcess = async (processSpawnPromise, processErrorPromise, processExitPromise) => {
+	const [eventName] = await Promise.race([processSpawnPromise, processErrorPromise]);
```
ehmicky, line 127: Either `spawn` or `error` is emitted.
Node.js source code about this: https://github.com/nodejs/node/blob/0056147a87c3964f83cb3c4b31a7b7fbfcba28e8/lib/internal/child_process.js#L398-L424

That's also [documented](https://nodejs.org/api/child_process.html#event-spawn):

> If the child process does not spawn successfully, the 'spawn' event is not emitted and the 'error' event is emitted instead.

PR #727, opened 2024-01-24
Fix `timeout` option

### LINKED ISSUE #726: `timeout` option's improper behavior
The `timeout` option has been part of `child_process.spawnSync()` for a long time. However, it's only been introduced to `child_process.spawn()` with Node 14. 

`execaSync()` just passes the `timeout` option to `child_process.spawnSync()`.

Instead, `execa()` emulates it and calls `process.kill()` on timeout. However, it still passes the `timeout` option to `child_process.spawn()`. This creates a race condition where, on timeout, the child process is terminated sometimes by Execa, sometimes by `child_process.spawn()`'s internal logic. 

We should fix this by _not_ passing the `timeout` option to `child_process.spawn()`, but still pass it to `child_process.spawnSync()`.

Note: we could potentially consider removing our `timeout` option's logic since it is now implemented in core Node.js. Unfortunately, `child_process.spawn()`'s `timeout` logic just sends a `SIGTERM` without any hint that the process was terminated due to a timeout. 

https://github.com/nodejs/node/blob/26f63be8787a9289539e0f19b5d9197de5038cb0/lib/child_process.js#L763-L781

This is very unhelpful, as users might be left wondering why the child process ended. Execa adds `error.timedOut` and indicates the timeout in `error.message`, which we should kept. Therefore, we should keep our existing `timeout` logic.

ehmicky:
Fixes #726.

### DIFF
--- a/test/fixtures/no-killable.js
+++ b/test/fixtures/no-killable.js
@@ -9,6 +9,4 @@ process.on('SIGINT', noop);
 process.send('');
 console.log('.');
 
-setInterval(() => {
-	// Run forever
-}, 20_000);
+setTimeout(noop, 1e8);
--- a/index.js
+++ b/index.js
@@ -103,6 +103,9 @@ const addDefaultOptions = ({
 	forceKillAfterDelay,
 });
 
+// Prevent passing the `timeout` option directly to `child_process.spawn()`
+const handleAsyncOptions = ({timeout, ...options}) => ({...options, timeoutDuration: timeout});
+
 const handleOutput = (options, value) => {
 	if (value === undefined || value === null) {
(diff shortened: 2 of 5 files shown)


test/fixtures/no-killable.js
```diff
@@ -9,6 +9,4 @@ process.on('SIGINT', noop);
 process.send('');
 console.log('.');
 
-setInterval(() => {
-	// Run forever
-}, 20_000);
+setTimeout(noop, 1e8);
```
ehmicky, line 12: Use same style as other fixtures.

ehmicky:
Fixed.

PR #729, opened 2024-01-24
Handle stream errors even with `buffer: false`

### LINKED ISSUE #716: Process streams uncaught exceptions when `buffer: false` is used
When the `buffer: false` option is used, any error on the child process streams creates an uncaught exception. For example, this crashes the current process:

```js
const subprocess = execa('echo', {buffer: false, reject: false});
subprocess.stdout.destroy(new Error('test'));
await subprocess
```

ehmicky:
Fixes #716.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -11,6 +11,23 @@ export const makeAllStream = ({stdout, stderr}, {all}) => all && (stdout || stde
 	? mergeStreams([stdout, stderr].filter(Boolean))
 	: undefined;
 
+// `childProcess.stdout|stderr` do not end until they have been consumed by `childProcess.all`.
+// `childProcess.all` does not end until `childProcess.stdout|stderr` have ended.
+// That's a good thing, since it ensures those streams are fully read and destroyed, even on errors.
+// However, this creates a deadlock if `childProcess.all` is not being read.
+// Doing so prevents the process from exiting, even when it failed.
+// It also prevents those streams from being properly destroyed.
+// This can only happen when:
+//  - `all` is `true`
+//  - `buffer` is `false`
+//  - `childProcess.all` is not read by the user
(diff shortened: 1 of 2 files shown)


lib/stream.js
```diff
@@ -11,6 +11,17 @@ export const makeAllStream = ({stdout, stderr}, {all}) => all && (stdout || stde
 	? mergeStreams([stdout, stderr].filter(Boolean))
 	: undefined;
 
+// When `buffer` is `false`, the user is responsible for reading the process streams.
+// If they don't, `stream.readableFlowing` remains `null`.
+// `childProcess.stdout|stderr` do not end until they have been consumed by `childProcess.all`.
+// `childProcess.all` does not end until `childProcess.stdout|stderr` have ended.
+// Therefore, there is a deadlock if `childProcess.all.readableFlowing` is not reading.
```
ehmicky, line 18: This took me some time to figure out!

lib/stream.js
```diff
@@ -11,6 +11,23 @@ export const makeAllStream = ({stdout, stderr}, {all}) => all && (stdout || stde
...
+// `childProcess.all` does not end until `childProcess.stdout|stderr` have ended.
+// That's a good thing, since it ensures those streams are fully read and destroyed, even on errors.
+// However, this creates a deadlock if `childProcess.all` is not being read.
+// Doing so prevents the process from exiting, even when it failed.
+// It also prevents those streams from being properly destroyed.
+// This can only happen when:
+//  - `all` is `true`
+//  - `buffer` is `false`
+//  - `childProcess.all` is not read by the user
+// Therefore, we forcefully resume `childProcess.all` flow on errors.
```
ehmicky, line 24: This took me some time to figure out!

sindresorhus: Oh wow. I bet. 👏

PR #732, opened 2024-01-24
Update index.d.ts

codesmith-emmy:


sindresorhus:
This should be done in the same pull request as the other change.

PR #735, opened 2024-01-25
Refactor `encoding` option

ehmicky:
This refactors the `encoding` option. This does not change the behavior, just simplify the implementation.

### DIFF
--- a/lib/stdio/encoding.js
+++ b/lib/stdio/encoding.js
@@ -5,18 +5,19 @@ import {isUint8Array} from './utils.js';
 // Apply the `encoding` option using an implicit generator.
 // This encodes the final output of `stdout`/`stderr`.
 export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
-	if (stdioStreams[0].direction === 'input' || IGNORED_ENCODINGS.has(encoding) || isSync) {
-		return stdioStreams.map(stdioStream => ({...stdioStream, encoding}));
+	const newStdioStreams = stdioStreams.map(stdioStream => ({...stdioStream, encoding}));
+	if (newStdioStreams[0].direction === 'input' || IGNORED_ENCODINGS.has(encoding) || isSync) {
+		return newStdioStreams;
 	}
 
 	const transform = encodingEndGenerator.bind(undefined, encoding);
-	const objectMode = stdioStreams.findLast(({type}) => type === 'generator')?.value.readableObjectMode === true;
+	const objectMode = newStdioStreams.findLast(({type}) => type === 'generator')?.value.objectMode === true;
 	return [
-		...stdioStreams,
+		...newStdioStreams,
 		{
-			...stdioStreams[0],
+			...newStdioStreams[0],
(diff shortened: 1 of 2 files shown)


lib/stdio/encoding.js
```diff
@@ -5,21 +5,22 @@ import {isUint8Array} from './utils.js';
 // Apply the `encoding` option using an implicit generator.
 // This encodes the final output of `stdout`/`stderr`.
 export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
-	if (stdioStreams[0].direction === 'input' || IGNORED_ENCODINGS.has(encoding) || isSync) {
-		return stdioStreams.map(stdioStream => ({...stdioStream, encoding}));
```
ehmicky, line 9: `stdioStream` represents a value with the `stdin`/`stdout`/`stderr`/`stdio` option.

`stdioStream.encoding` is set the same as the `encoding` option. It is used to know whether to convert the chunks passed to the transform as strings or as `Uint8Array`.

lib/stdio/encoding.js
```diff
@@ -5,21 +5,22 @@ import {isUint8Array} from './utils.js';
...
-	}
-
-	const transform = encodingEndGenerator.bind(undefined, encoding);
-	const objectMode = stdioStreams.findLast(({type}) => type === 'generator')?.value.readableObjectMode === true;
```
ehmicky, line 13: The following line basically indicates: uses the same `objectMode` as the previous generator, or `false` if none. This PR simplifies this by using an internal `"previous"` value, which is handled below.

lib/stdio/generator.js
```diff
@@ -36,7 +36,7 @@ The same applies to the first output's generator's `writableObjectMode`.
 */
 const getOutputObjectModes = (objectMode, index, newGenerators) => {
 	const writableObjectMode = index !== 0 && newGenerators[index - 1].value.readableObjectMode;
-	const readableObjectMode = objectMode;
+	const readableObjectMode = objectMode === 'previous' ? writableObjectMode : objectMode;
```
ehmicky, line 39: For output streams, `writableObjectMode` is always the same as the previous generator's `readableObjectMode` (if any).

lib/stdio/handle.js
```diff
@@ -13,8 +13,8 @@ export const handleInput = (addProperties, options, isSync) => {
 	const stdioStreamsGroups = [[...stdinStreams, ...handleInputOptions(options)], ...otherStreamsGroups]
 		.map(stdioStreams => validateStreams(stdioStreams))
 		.map(stdioStreams => addStreamDirection(stdioStreams))
-		.map(stdioStreams => normalizeGenerators(stdioStreams))
 		.map(stdioStreams => handleStreamsEncoding(stdioStreams, options, isSync))
+		.map(stdioStreams => normalizeGenerators(stdioStreams))
```
ehmicky, line 17: Moving this line allows re-using the normalization logic from `normalizeGenerators()`.

PR #736, opened 2024-01-25
Fix using both `objectMode` and `encoding` options

ehmicky:
This fixes using both: 
  - Transforms in `objectMode`
  - The `encoding` option with values like `hex` or `base64`

When a transform in `objectMode` returns an object, `encoding: 'base64'` should be a noop.

When a transform in `objectMode` returns a string or an `Uint8Array`, `encoding: base64` can work. However, it must operate chunk-wise instead of line-wise, since consumers of streams in `objectMode` might assume for emitted chunks to be independent from each other. For example, we will make that assumption with #706: we will emit one line at a time in `objectMode`, and would assume to be encoded separately from the next one.

### DIFF
--- a/lib/stdio/encoding.js
+++ b/lib/stdio/encoding.js
@@ -10,8 +10,9 @@ export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
 		return newStdioStreams;
 	}
 
-	const transform = encodingEndGenerator.bind(undefined, encoding);
 	const objectMode = newStdioStreams.findLast(({type}) => type === 'generator')?.value.objectMode === true;
+	const encodingEndGenerator = objectMode ? encodingEndObjectGenerator : encodingEndStringGenerator;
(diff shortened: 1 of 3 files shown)


lib/stdio/encoding.js
```diff
@@ -29,7 +29,7 @@ const encodingEndGenerator = async function * (encoding, chunks) {
 	const stringDecoder = new StringDecoder(encoding);
 
 	for await (const chunk of chunks) {
-		yield stringDecoder.write(chunk);
+		yield isUint8Array(chunk) ? stringDecoder.write(chunk) : chunk;
```
ehmicky, line 32: This is the transform used for the `encoding: 'base64'` option.
Since previous transforms might use `objectMode: true` and return any type, chunks of this transform might be of any type. Encoding non-strings/`Uint8Array` to base64 becomes a noop then.

PR #737, opened 2024-01-25
Do not allow returning `null` from transforms

ehmicky:
The `null` values have a special value in Node.js streams. With a readable stream, it ends it. With a writable stream, it is not allowed and throws.

This PR adds some better validation to ensure users do not return `null` with transforms.


sindresorhus:
I think we should more clearly document this too.

ehmicky:
:+1: Done in https://github.com/sindresorhus/execa/pull/740

PR #738, opened 2024-01-25
Fix `encoding` option not working with `ignore` and `inherit`

ehmicky:
Using `encoding: 'hex'` or `encoding: 'base64'` currently turn any `{ stdout: 'ignore' }`, `{ stdout: 'inherit' }`, etc. into `{ stdout: 'pipe' }`. This PR fixes this bug.

### DIFF
--- a/lib/stdio/encoding.js
+++ b/lib/stdio/encoding.js
@@ -1,12 +1,13 @@
 import {StringDecoder} from 'node:string_decoder';
 import {Buffer} from 'node:buffer';
 import {isUint8Array} from './utils.js';
+import {willPipeStreams} from './pipe.js';
 
 // Apply the `encoding` option using an implicit generator.
 // This encodes the final output of `stdout`/`stderr`.
 export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
 	const newStdioStreams = stdioStreams.map(stdioStream => ({...stdioStream, encoding}));
-	if (newStdioStreams[0].direction === 'input' || IGNORED_ENCODINGS.has(encoding) || isSync) {
+	if (!shouldEncodeOutput(newStdioStreams, encoding, isSync)) {
 		return newStdioStreams;
 	}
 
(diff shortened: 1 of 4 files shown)


lib/stdio/pipe.js
```diff
@@ -0,0 +1,18 @@
...
+
+const PIPED_STDIO_VALUES = new Set(['pipe', 'overlapped', undefined, null]);
+
+const updateStdioItem = stdioStreams => {
+	if (stdioStreams.length > 1) {
+		return stdioStreams.some(({value}) => value === 'overlapped') ? 'overlapped' : 'pipe';
+	}
+
+	const [stdioStream] = stdioStreams;
+	return stdioStream.type !== 'native' && stdioStream.value !== 'overlapped' ? 'pipe' : stdioStream.value;
```
ehmicky, line 17: When using transforms, we set the `std*` option passed to `child_process.spawn()` as `pipe`. This allows piping transforms.

The `encoding: 'hex'` and `encoding: 'base64'` options use an internal transform. The bug is: this internal transform currently sets the `std*` option to `pipe`, due to the above, even when something like `stdout: 'ignore'` has been used.

For example, `{ stdout: 'ignore', encoding: 'base64' }` would become `{ stdout: 'pipe', encoding: 'base64' }`.

However, there is no need to add the `encoding` internal transform when the stream is ignored. This is the solution implemented by this PR.

lib/stdio/pipe.js
```diff
@@ -0,0 +1,18 @@
+// When the `std*: Iterable | WebStream | URL | filePath`, `input` or `inputFile` option is used, we pipe to `childProcess.std*`.
+// When the `std*: Array` option is used, we emulate some of the native values ('inherit', Node.js stream and file descriptor integer). To do so, we also need to pipe to `childProcess.std*`.
+// Therefore the `std*` options must be either `pipe` or `overlapped`. Other values do not set `childProcess.std*`.
+export const updateStdio = stdioStreamsGroups => stdioStreamsGroups.map(stdioStreams => updateStdioItem(stdioStreams));
+
+// Whether `childProcess.std*` will be set
+export const willPipeStreams = stdioStreams => PIPED_STDIO_VALUES.has(updateStdioItem(stdioStreams));
+
+const PIPED_STDIO_VALUES = new Set(['pipe', 'overlapped', undefined, null]);
```
ehmicky, line 9: Those are the `std*` values that set `childProcess.std*`.

lib/stdio/encoding.js
```diff
@@ -24,6 +25,11 @@ export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
 	];
 };
 
+const shouldEncodeOutput = (stdioStreams, encoding, isSync) => stdioStreams[0].direction === 'output'
+	&& !IGNORED_ENCODINGS.has(encoding)
+	&& !isSync
+	&& willPipeStreams(stdioStreams);
```
ehmicky, line 31: If this condition is `true`, the `encoding` internal transform is used.

PR #739, opened 2024-01-26
Fix error handling of transforms

ehmicky:
When using multiple transforms and one of them throws, the other transforms are currently not always properly destroyed. This makes the promise returned by Execa hang forever.

This PR fixes this.

### DIFF
--- a/lib/stdio/utils.js
+++ b/lib/stdio/utils.js
@@ -10,7 +10,7 @@ const textDecoder = new TextDecoder();
 export const binaryToString = uint8ArrayOrBuffer => textDecoder.decode(uint8ArrayOrBuffer);
 
 // Like `source.pipe(destination)`, if `source` ends, `destination` ends.
-// Like `stream.pipeline(source, destination)`, if `source` aborts/errors, `destination` aborts.
+// Like `Stream.pipeline(source, destination)`, if `source` aborts/errors, `destination` aborts.
 // Unlike `stream.pipeline(source, destination)`, `destination` behavior does not propagate to `source`.
 export const pipeline = async (source, destination) => {
 	source.pipe(destination);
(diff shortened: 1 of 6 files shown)


lib/stdio/utils.js
```diff
@@ -7,3 +8,16 @@ export const isBinary = value => isUint8Array(value) || Buffer.isBuffer(value);
 
 const textDecoder = new TextDecoder();
 export const binaryToString = uint8ArrayOrBuffer => textDecoder.decode(uint8ArrayOrBuffer);
+
+// Like `source.pipe(destination)`, if `source` ends, `destination` ends.
+// Like `stream.pipeline(source, destination)`, if `source` aborts/errors, `destination` aborts.
+// Unlike `stream.pipeline(source, destination)`, `destination` behavior does not propagate to `source`.
```
sindresorhus, line 14: It's unclear what exactly "behavior" means here.

lib/stdio/utils.js
```diff
@@ -7,3 +8,16 @@ export const isBinary = value => isUint8Array(value) || Buffer.isBuffer(value);
...
+export const pipeline = async (source, destination) => {
```
sindresorhus, line 15: I would call it something else so it's not confused at callsite with `Stream.pipeline`

lib/stdio/utils.js
```diff
@@ -7,3 +8,16 @@ export const isBinary = value => isUint8Array(value) || Buffer.isBuffer(value);
 
 const textDecoder = new TextDecoder();
 export const binaryToString = uint8ArrayOrBuffer => textDecoder.decode(uint8ArrayOrBuffer);
+
+// Like `source.pipe(destination)`, if `source` ends, `destination` ends.
+// Like `stream.pipeline(source, destination)`, if `source` aborts/errors, `destination` aborts.
```
sindresorhus, line 13:
```suggestion
// Like `Stream.pipeline(source, destination)`, if `source` aborts/errors, `destination` aborts.
```

Otherwise it reads as if it's `Stream.prototype.pipeline`.

lib/stdio/utils.js
ehmicky: Fixed in e8f3cab

lib/stdio/utils.js
ehmicky: Fixed in 1533da2

PR #748, opened 2024-01-28
Simplify transforms syntax

ehmicky:
This PR simplifies the transforms' syntax.
The assumption is that most transforms will simply map or filter lines, so the syntax should be optimized for this use case.

#### Before

```js
const transform = async function * (lines) {
  for await (const line of lines) {
    yield line.toUpperCase();
  }
};
```

#### After

```js
const transform = function * (line) {
  yield line.toUpperCase();
};
```

#### Before

```js
const transform = async function * (lines) {
  for await (const line of lines) {
    if (!line.includes('secret')) {
      yield line;
    }
  }
};
```

#### After

```js
const transform = function * (line) {
  if (!line.includes('secret')) {
    yield line.toUpperCase();
  }
};
```

#### Before

```js
const transform = async function * (lines) {
  let count = 0; 

  for await (const line of lines) {
    yield `[${count++}] ${line}`;
  }
};
```

#### After

```js
let count = 0;

const transform = function * (line) {
  yield `[${count++}] ${line}`;
};
```

### Final lines

Not sure whether this one results in a simpler syntax, but this is more of an edge case.

#### Before

```js
const transform = async function * (lines) {
  let count = 0; 

  for await (const line of lines) {
    count += 1;
    yield line;
  }

  yield `Number of lines: ${count}`; 
};

await execa('...', {stdout: transform});
```

#### After

```js
let count = 0;

const transform = function * (line) {
  count += 1
  yield line;
};

const final = function * () {
  yield `Number of lines: ${count}`; 
}

await execa('...', {stdout: {transform, final}});
```

### Additional pros

Beyond the simplified syntax, this brings the following pros.

#### Implementation simplicity

The implementation can be moved to using a single, simpler `Transform` stream. As opposed to what we currently use: a `Duplex.from()` + `PassThrough` + `Readable.iterator()` + `Readable.from()` setup. Using a more straightforward `Transform` will be more efficient and result in fewer bugs.

I can implement this in a follow-up PR. 

#### Synchronous transforms

Allows the transform to be synchronous, which results in better performance in that case by avoiding repeatedly calling `await`. 

The transforms we use internally for the `lines` option, the `encoding` option, the transform line-wise logic and the transform encoding logic are all synchronous. Therefore, that's important not to run them asynchronously unnecessarily.

### DIFF
--- a/lib/stdio/duplex.js
+++ b/lib/stdio/duplex.js
@@ -20,7 +20,7 @@ export const generatorsToDuplex = (generators, {writableObjectMode, readableObje
 	let iterable = passThrough.iterator();
 
 	for (const generator of generators) {
-		iterable = generator(iterable);
+		iterable = applyGenerator(generator, iterable);
 	}
 
 	const readableStream = Readable.from(iterable, {
@@ -29,6 +29,16 @@ export const generatorsToDuplex = (generators, {writableObjectMode, readableObje
 	});
 	const duplexStream = Duplex.from({writable: passThrough, readable: readableStream});
 	return duplexStream;
+};
+
+const applyGenerator = async function * ({transform, final}, chunks) {
+	for await (const chunk of chunks) {
+		yield * transform(chunk);
+	}
+
+	if (final !== undefined) {
+		yield * final();
+	}
 };
 
 /*
--- a/lib/stdio/encoding.js
+++ b/lib/stdio/encoding.js
@@ -12,14 +12,21 @@ export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
 	}
 
 	const objectMode = newStdioStreams.findLast(({type}) => type === 'generator')?.value.objectMode === true;
-	const encodingEndGenerator = objectMode ? encodingEndObjectGenerator : encodingEndStringGenerator;
-	const transform = encodingEndGenerator.bind(undefined, encoding);
+	const stringDecoder = new StringDecoder(encoding);
+	const generator = objectMode
+		? {
+			transform: encodingEndObjectGenerator.bind(undefined, stringDecoder),
+		}
+		: {
+			transform: encodingEndStringGenerator.bind(undefined, stringDecoder),
+			final: encodingEndStringFinal.bind(undefined, stringDecoder),
+		};
 	return [
 		...newStdioStreams,
 		{
 			...newStdioStreams[0],
 			type: 'generator',
-			value: {transform, binary: true, objectMode},
+			value: {...generator, binary: true, objectMode},
 			encoding: 'buffer',
 		},
 	];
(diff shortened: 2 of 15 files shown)


lib/stdio/duplex.js
```diff
@@ -31,6 +31,16 @@ export const generatorsToDuplex = (generators, {writableObjectMode, readableObje
...
 
+const applyGenerator = async function * ({transform, final}, chunks) {
+	for await (const chunk of chunks) {
+		yield * transform(chunk);
+	}
+
+	if (final !== undefined) {
+		yield * final();
+	}
+};
```
ehmicky, line 42: This new function is the main change of this PR. This is what allows simplifying the syntax.

lib/stdio/encoding.js
```diff
@@ -12,14 +12,21 @@ export const handleStreamsEncoding = (stdioStreams, {encoding}, isSync) => {
...
-	const transform = encodingEndGenerator.bind(undefined, encoding);
+	const stringDecoder = new StringDecoder(encoding);
+	const generator = objectMode
+		? {
+			transform: encodingEndObjectGenerator.bind(undefined, stringDecoder),
+		}
+		: {
+			transform: encodingEndStringGenerator.bind(undefined, stringDecoder),
+			final: encodingEndStringFinal.bind(undefined, stringDecoder),
+		};
```
ehmicky, line 23: We use a few transforms internally. This PR rewrites them to using the new syntax.

lib/stdio/type.js
```diff
@@ -3,14 +3,10 @@ import {isUint8Array} from './utils.js';
...
 // The `stdin`/`stdout`/`stderr` option can be of many types. This detects it.
 export const getStdioOptionType = (stdioOption, optionName) => {
-	if (isAsyncGenerator(stdioOption)) {
+	if (isGenerator(stdioOption)) {
 		return 'generator';
 	}
 
-	if (isSyncGenerator(stdioOption)) {
-		throw new TypeError(`The \`${optionName}\` option must use an asynchronous generator, not a synchronous one.`);
-	}
```
ehmicky, line 12: The previous syntax required the generators to be `async` due to implementation details. This new syntax does not.

test/helpers/generator.js
```diff
@@ -1,55 +1,61 @@
...
+});
+
+export const convertTransformToFinal = (transform, final) => {
+	if (!final) {
+		return transform;
+	}
+
+	const generatorOptions = typeof transform === 'function' ? {transform} : transform;
+	return ({...generatorOptions, transform: noYieldTransform, final: generatorOptions.transform});
+};
```
ehmicky, line 61: Test helper to allow re-using the same test for both the `transform` and the `final` methods.

test/stdio/generator.js
```diff
@@ -601,72 +620,55 @@ test('Generators take "maxBuffer" into account', async t => {
...
-	const state = {};
-	const childProcess = execa('noop-fail.js', ['1'], {stdout: errorHandlerGenerator.bind(undefined, state)});
-	const error = new Error('test');
-	childProcess.stdout.emit('error', error);
-	const thrownError = await t.throwsAsync(childProcess);
-	t.is(error, thrownError);
-	await setImmediate();
-	t.is(state.error.code, 'ABORT_ERR');
+	await t.throwsAsync(childProcess, {message: GENERATOR_ERROR_REGEXP});
 });
```
ehmicky, line 674: This test does not make sense anymore with the new syntax.

sindresorhus:
This is a very nice simplification 👍

PR #752, opened 2024-01-28
Only allow child processes as argument to `.pipeStdout()`

### LINKED ISSUE #744: Simplify `.pipe*` argument type
## Proposal

The `.pipe*(...)` methods can get three possible argument types: a child process, a file path or a stream.
The last two are now redundant with the `std*` options.

For example:

```js
await execa(...).pipeStdout('file.txt');
await execa(...).pipeStdout(stream);
```

Can now use:

```js
await execa(..., {stdout: {file: 'file.txt'}});
await execa(..., {stdout: stream});
```

We should remove support for passing a file path or stream to the `.pipe*()` methods, and only allow passing a child process.

## Pros

The `std*` options are better for file paths and streams and should be preferred.
  - They automatically handle listening for the underlying stream's errors and completion.
  - They can be combined with other destinations, by passing an array of values.

Also, this removes a source of confusion with the `.pipeStdout()` method (and other `.pipe(...)` methods). 
  - Right now, `source.pipeStdout(destination)` returns `destination` if it is a child process, but returns `source` if it is a file path or stream. 
  - This is confusing. For example: 
     - `source.pipeStdout(childProcess).pipeStdout(otherChildProcess)` sets `source -> childProcess -> otherChildProcess`, 
     - while `source.pipeStdout(stream).pipeStdout(otherStream)` sets `source -> stream` + `source -> otherStream`.

The above also means that the pipe methods currently do not listen to the stream's argument error and completion, which can result in uncaught exceptions crashing the process.

Finally, this makes our API more straightforward from a design standpoint:
  - The `stdin`/`stdout`/`stderr`/`stdio` options are the main way to redirect standard streams
  - The `.pipe*()` methods are only intended to pipe multiple processes

--- 

What do you think?

ehmicky:
Fixes #744.


readme.md
```diff
@@ -319,28 +319,31 @@ This is `undefined` if either:
...
+`execaChildProcess`: [`execa()` return value](#pipe-multiple-processes)
 
-If the `target` is another [`execa()` return value](#execacommandcommand-options), it is returned. Otherwise, the original `execa()` return value is returned. This allows chaining `pipeStdout()` then `await`ing the [final result](#childprocessresult).
+[Pipe](https://nodejs.org/api/stream.html#readablepipedestination-options) the child process' `stdout` to another Execa child process' `stdin`.
 
-The [`stdout` option](#stdout-1) must be kept as `pipe`, its default value.
+Returns `execaChildProcess`, which allows chaining `pipeStdout()` then `await`ing the [final result](#childprocessresult).
 
-#### pipeStderr(target)
+[`childProcess.stdout`](#stdout) must not be `undefined`.
```
ehmicky, line 330: This line is more accurate than the previous one.
For example, `{stdout: {file: 'file.txt'}}` works because it uses `pipe` under the hood.

readme.md
```diff
@@ -319,28 +319,31 @@ This is `undefined` if either:
...
-The [`stderr` option](#stderr-1) must be kept as `pipe`, its default value.
+`execaChildProcess`: [`execa()` return value](#pipe-multiple-processes)
 
-#### pipeAll(target)
+Like [`pipeStdout()`](#pipestdoutexecachildprocess) but piping the child process's `stderr` instead.
 
-Combines both [`pipeStdout()`](#pipestdouttarget) and [`pipeStderr()`](#pipestderrtarget).
+[`childProcess.stderr`](#stderr) must not be `undefined`.
 
-Either the [`stdout` option](#stdout-1) or the [`stderr` option](#stderr-1) must be kept as `pipe`, their default value. Also, the [`all` option](#all-2) must be set to `true`.
```
ehmicky, line 343: The first sentence is redundant, since we already mention that this method combines both `pipeStdout()` and `pipeStderr()` which already document that requirement.

PR #754, opened 2024-01-29
Add a new test related to generators

ehmicky:
This adds a new test related to generators.

### DIFF
--- a/test/stdio/generator.js
+++ b/test/stdio/generator.js
@@ -526,6 +526,41 @@ const testMultipleYields = async (t, final) => {
 test('Generator can yield "transform" multiple times at different moments', testMultipleYields, false);
 test('Generator can yield "final" multiple times at different moments', testMultipleYields, true);
 
+const partsPerChunk = 4;
+const chunksPerCall = 10;
+const callCount = 5;
+const fullString = '\n'.repeat(getDefaultHighWaterMark(false) / partsPerChunk);
+
+const yieldFullStrings = function * () {
+	yield * Array.from({length: partsPerChunk * chunksPerCall}).fill(fullString);
+};
+
+const manyYieldGenerator = async function * () {
+	for (let index = 0; index < callCount; index += 1) {
+		yield * yieldFullStrings();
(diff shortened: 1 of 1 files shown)


test/stdio/generator.js
```diff
@@ -526,6 +526,41 @@ const testMultipleYields = async (t, final) => {
...
+	yield * Array.from({length: partsPerChunk * chunksPerCall}).fill(fullString);
+};
+
+const manyYieldGenerator = async function * () {
+	for (let index = 0; index < callCount; index += 1) {
+		yield * yieldFullStrings();
+		// eslint-disable-next-line no-await-in-loop
+		await setImmediate();
+	}
+};
```
ehmicky, line 544: What this is doing: use a transform that calls `yield` multiple times.
It essentially calls `yield '\n\n\n...\n'` 40 times, then repeat this 5 times, i.e. does it 200 times.
Each `yield` contains 4096 bytes.

test/stdio/generator.js
```diff
@@ -526,6 +526,41 @@ const testMultipleYields = async (t, final) => {
...
+		stripFinalNewline: false,
+		buffer: false,
+	});
+	const [chunks] = await Promise.all([
+		getStreamAsArray(childProcess.stdout),
+		childProcess,
+	]);
+	const expectedLength = chunksPerCall * callCount;
+	const expectedChunk = Buffer.alloc(getDefaultHighWaterMark(false)).fill('\n');
+	t.deepEqual(chunks, Array.from({length: expectedLength}).fill(expectedChunk));
```
ehmicky, line 558: Since the highWaterMark is 16KB, we expect:
  - Every 4 `yield` will result in a single `data` event on `childProcess.stdout`. I.e. this ensures that data can be read as soon as it reaches the highWaterMark, as opposed to being buffered beyond it.
  - Each data being read will be exactly 16KB (`expectedChunk`)
  - Data will be emitted 200 / 4 = 50 times (`expectedLength`)

ehmicky:
Closing in favor of #755, which implements this test too.

PR #755, opened 2024-01-30
Simplify transforms implementation

ehmicky:
This PR refactors transforms using a single `Transform` stream instead of the current implementation which combines `PassThrough`, `Readable.iterator()`, `Readable.from()` and `Duplex.from()`. This should make it easier to understand that part of the code. This should also make the feature more stable by relying less on the internal mysteries of Node.js streams.

This does not change behavior except for a small detail that I am highlighting below.


test/stdio/generator.js
```diff
@@ -310,16 +311,17 @@ const testHighWaterMark = async (t, passThrough, binary, objectMode) => {
 			...(objectMode ? [outputObjectGenerator] : []),
 			writerGenerator,
 			...(passThrough ? [noopGenerator(false, binary)] : []),
-			{transform: getLengthGenerator, binary: true},
+			{transform: getLengthGenerator.bind(undefined, t), binary: true, objectMode: true},
 		],
 	});
-	t.is(stdout, `${getDefaultHighWaterMark()}`.repeat(repeatHighWaterMark));
+	t.is(stdout.length, repeatCount);
+	t.true(stdout.every(chunk => chunk.toString() === '\n'));
```
ehmicky, line 318: Small difference of behavior. When a transform calls `yield`: 
  - Multiple times
  - And synchronously
  - And with a value smaller than the chunk it gets as argument
  - And `objectMode` is not `true`

Then each `yield` will trigger a single read. Previously, those `yield`s would be buffered into a single read, up to the `highWaterMark`.

The new behavior will result in less memory consumption in that case. It might or might not result in faster performance, depending on the case. That being said, the above conditions should be fairly rare. 

This change of behavior is due to some implementation details of how `Transform` works, compared to how `Readable.iterator()` + `Readable.from()` work. Namely, `Readable.iterator()` [uses the `readable` event](https://github.com/nodejs/node/blob/d4eb03f27fb199625f2aa763c817442fb1ac02b0/lib/internal/streams/readable.js#L1369) while `Transform` [does not](https://github.com/nodejs/node/blob/d4eb03f27fb199625f2aa763c817442fb1ac02b0/lib/internal/streams/transform.js#L198). The `readable` event is emitted on the next tick, which allows synchronous writes to be buffered in a given tick. On the other hand, the `data` event is emitted synchronously as the write is being performed.

PR #758, opened 2024-01-30
Improve default value of the `stdin` option with `$`

### LINKED ISSUE #742: Improve default value of the `stdin`  option with `$`
## Background

Some commands need their `stdin` to be interactive. For example, script prompts.

Other commands need it to be programmatic. For example when:
  - piping processes, `childProcess.pipeStdout(otherChildProcess)`
  - using `childProcess.stdin.write()`
  - using `stream.pipe(childProcess.stdin)`
  - passing the `input` or `inputFile` option

And then many commands allow both, such as most Unix utilities like `cat`, `grep` or `rg`.

## Previous default value

Initially, we've started with `$` having the same default value for the `stdin` option as `execa()`, i.e. `"pipe"`.

With #548, #549 and #550, we switched the default value to `"inherit"`.

## New default value

However, with the recent additions to the `stdin` option, we can now have the best of both worlds! We could switch the default value for the `stdin` option with `$` to `['pipe', 'inherit']` instead. 

This means users would be allowed to use both interactive and programmatic `stdin`. 

One of the upsides is to be able to use:

```js
$`command ...`.pipe($`otherCommand ...`)
```

Instead of the following, which is currently making piping processes more verbose and error-prone:

```js
$`command ...`.pipe($({ stdin: 'pipe' })`otherCommand ...`)
```

It also means users would be able to use `childProcess.stdin.write()` and `stream.pipe(childProcess.stdin)` without setting any `stdin` option.

What do you think?

ehmicky:
Fixes #742.

Currently in draft: Windows tests are failing in CI. The underlying issue requires fixes by other PRs, so I'm pausing this one.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1100,6 +1100,8 @@ type Execa$<OptionsType extends CommonOptions = {}> = {
 
 	Cannot use the following options: `all`, `cleanup`, `buffer`, `detached`, `serialization`, `signal` and `lines`. Also, the `stdin`, `stdout`, `stderr`, `stdio` and `input` options cannot be an array, an iterable or a web stream. Node.js streams must have a file descriptor unless the `input` option is used.
 
+	Unlike $\`command\`, the `stdin` option defaults to `"inherit"`, not `["pipe", "inherit"]`.
+
 	Returns or throws a `childProcessResult`. The `childProcess` is not returned: its methods and properties are not available. This includes [`.kill()`](https://nodejs.org/api/child_process.html#subprocesskillsignal), [`.pid`](https://nodejs.org/api/child_process.html#subprocesspid), `.pipe()` and the [`.stdin`/`.stdout`/`.stderr`](https://nodejs.org/api/child_process.html#subprocessstdout) streams.
 
 	@returns A `childProcessResult` object
--- a/test/command.js
+++ b/test/command.js
@@ -330,6 +330,11 @@ test('$ stdin defaults to "inherit"', async t => {
 	t.is(stdout, 'foo');
 });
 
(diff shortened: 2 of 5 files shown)


index.d.ts
```diff
@@ -1108,6 +1108,8 @@ type Execa$<OptionsType extends CommonOptions = {}> = {
 
 	Cannot use the following options: `all`, `cleanup`, `buffer`, `detached`, `serialization`, `signal` and `lines`. Also, the `stdin`, `stdout`, `stderr`, `stdio` and `input` options cannot be an array, an iterable or a web stream. Node.js streams must have a file descriptor unless the `input` option is used.
 
+	Unlike $\`command\`, the `stdin` option defaults to `"inherit"`, not `["pipe", "inherit"]`.
```
ehmicky, line 1111: `child_process.spawnSync()` does not return any child process. So we cannot do the manual piping logic we use to support `["pipe", "inherit"]`.

test/command.js
```diff
@@ -330,6 +330,11 @@ test('$ stdin defaults to "inherit"', async t => {
 	t.is(stdout, 'foo');
 });
 
+test('$ stdin also defaults to "pipe"', async t => {
+	const {stdout} = await $`noop.js foo`.pipeStdout($`stdin.js`);
```
ehmicky, line 334: Note: `.pipeStdout()` will need to be renamed to `.pipe()` once https://github.com/sindresorhus/execa/pull/757 is merged.

ehmicky:
Closing due to https://github.com/sindresorhus/execa/issues/742#issuecomment-1932653207

PR #768, opened 2024-02-01
Fix calling `.kill()` many times

### LINKED ISSUE #767: Calling `.kill()` multiple times in a row triggers a `maxListeners` warning
When calling `childProcess.kill()` many times in a row, a `maxListeners` warning is emitted, providing the `forceKillAfterDelay` is being used.

```js
const childProcess = execa('not-killable.js');
for (let index = 0; index < 11; index += 1) {
  childProcess.kill();
}
await childProcess;
```

ehmicky:
Fixes #767.

This PR also adds a few more tests related to `.kill()`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 import {Buffer} from 'node:buffer';
+import {setMaxListeners} from 'node:events';
 import path from 'node:path';
 import childProcess from 'node:child_process';
 import process from 'node:process';
@@ -151,6 +152,7 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	}
 
 	const controller = new AbortController();
+	setMaxListeners(Number.POSITIVE_INFINITY, controller.signal);
 
 	pipeOutputAsync(spawned, stdioStreamsGroups);
 
--- a/test/kill.js
+++ b/test/kill.js
@@ -1,5 +1,5 @@
 import process from 'node:process';
-import {once} from 'node:events';
+import {once, defaultMaxListeners} from 'node:events';
 import {constants} from 'node:os';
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -151,6 +152,7 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	}
 
 	const controller = new AbortController();
+	setMaxListeners(Number.POSITIVE_INFINITY, controller.signal);
```
ehmicky, line 155: One `AbortController` is created per `execa()` call and we're the only one using it, so it's pretty safe to undo the `maxListeners` warning altogether.

In principle, we could be stricter and compute the exact amount of times `signal.on('abort')` is being used, but that's fairly tedious and would not be give much benefits.

test/kill.js
```diff
@@ -71,18 +93,9 @@ if (process.platform !== 'win32') {
...
 	test('`forceKillAfterDelay` should kill after a timeout with the killSignal number', testForceKill, 50, constants.signals.SIGINT, {killSignal: constants.signals.SIGINT});
 
-	const testInvalidForceKill = async (t, forceKillAfterDelay) => {
-		t.throws(() => {
-			execa('empty.js', {forceKillAfterDelay});
-		}, {instanceOf: TypeError, message: /non-negative integer/});
-	};
-
-	test('`forceKillAfterDelay` should not be NaN', testInvalidForceKill, Number.NaN);
-	test('`forceKillAfterDelay` should not be negative', testInvalidForceKill, -1);
```
ehmicky, line 81: Those tests can be run on Windows, so they're moved outside of the "no Windows" condition.

PR #772, opened 2024-02-02
Small refactoring of `buffer: false` logic

ehmicky:
This is a small refactoring of the `buffer: false` logic.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -55,7 +55,10 @@ const getStreamPromise = async ({stream, encoding, buffer, maxBuffer}) => {
 	}
 
 	if (!buffer) {
-		await Promise.all([finished(stream), resumeStream(stream)]);
+		await Promise.all([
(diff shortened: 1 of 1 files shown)


lib/stream.js
```diff
@@ -55,7 +55,10 @@ const getStreamPromise = async ({stream, encoding, buffer, maxBuffer}) => {
 	}
 
 	if (!buffer) {
-		await Promise.all([finished(stream), resumeStream(stream)]);
+		await Promise.all([
+			finished(stream, {cleanup: true, readable: true, writable: false}),
```
ehmicky, line 59: `stream` is either `childProcess.stdout`, `childProcess.stderr` or `childProcess.all`.
Those are all technically both writable and readable, and we only need to await their readable side.
In practice, I do not see a situation where the readable side would end before the writable one, but let's be explicit here just in case.

PR #775, opened 2024-02-02
Improve stream cleanup and termination logic

### LINKED ISSUE #760: Handle aborted readable stream passed to the `stdin` option
Passing an aborted Node.js `Readable` stream to the `stdin` option, when combined with another value, currently makes the process hang forever.

```js
const stream = new Readable();
stream.destroy();
await execa('...', {stdin: [stream, 'pipe']}); // This never resolves
```

### LINKED ISSUE #761: Do not hang processes when passing a file path to `stdin`
When passing a file path to `stdin` and the process exits before that file was read, the Execa promise hangs forever.

```js
await execa('...', {stdin: {file: 'path.txt'}, timeout: 1}); // This never resolves
```

### LINKED ISSUE #762: Cancel `ReadableStream` when process errors
When passing a `ReadableStream` to the `stdin` option, and the process errors, that stream should be canceled.
Currently, this is not the case.

This also impacts passing an iterable, string and `Uint8Array` to the `stdin` option, since those create a stream under the hood, that is currently not canceled.

ehmicky:
Fixes #760.
Fixes #761.
Fixes #762.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -83,27 +83,12 @@ const resumeStream = async stream => {
 
 const applyEncoding = (contents, encoding) => encoding === 'buffer' ? new Uint8Array(contents) : contents;
 
-// Retrieve streams created by the `std*` options
-const getCustomStreams = stdioStreamsGroups => stdioStreamsGroups.flat().filter(({type}) => type !== 'native');
-
-// Some `stdout`/`stderr` options create a stream, e.g. when passing a file path.
-// The `.pipe()` method automatically ends that stream when `childProcess.stdout|stderr` ends.
-// This makes sure we want for the completion of those streams, in order to catch any error.
-// Since we want to end those streams, they cannot be infinite, except for `process.stdout|stderr`.
-// However, for the `stdin`/`input`/`inputFile` options, we only wait for errors, not completion.
-// This is because source streams completion should end destination streams, but not the other way around.
-// This allows for source streams to pipe to multiple destinations.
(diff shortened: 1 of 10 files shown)


lib/stream.js
```diff
@@ -80,27 +80,12 @@ const resumeStream = async stream => {
...
-	.map(({value}) => throwOnStreamError(value));
-
-const shouldWaitForCustomStream = (value, type, direction) => (type === 'fileUrl' || type === 'filePath')
-	|| (direction === 'output' && !isStandardStream(value));
+// Some `stdin`/`stdout`/`stderr` options create a stream, e.g. when passing a file path.
+// The `.pipe()` method automatically ends that stream when `childProcess` ends.
+// This makes sure we wait for the completion of those streams, in order to catch any error.
+const waitForCustomStreamsEnd = stdioStreamsGroups => stdioStreamsGroups.flat()
+	.filter(({type, value}) => type !== 'native' && !isStandardStream(value))
+	.map(({value}) => finished(value, {cleanup: true}));
```
ehmicky, line 88: We now properly clean up all streams, whether input or output. In particular, if the child process errors, the input streams (such as a transform or a `ReadableStream` passed to the `stdin` option) are aborted.

So we can safely wait for their completion.

ehmicky:
Fixed merge conflict.

sindresorhus:
Another merge conflict

ehmicky:
Fixed merged conflict. :+1:

PR #776, opened 2024-02-02
Refactor spawning logic

ehmicky:
This is a small refactoring of the logic related to spawning. It just moves some lines around, without changing the behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@ import {makeError} from './lib/error.js';
 import {handleInputAsync, pipeOutputAsync, cleanupStdioStreams} from './lib/stdio/async.js';
 import {handleInputSync, pipeOutputSync} from './lib/stdio/sync.js';
 import {normalizeStdioNode} from './lib/stdio/normalize.js';
-import {spawnedKill, validateTimeout, normalizeForceKillAfterDelay} from './lib/kill.js';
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -156,6 +156,7 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	setMaxListeners(Number.POSITIVE_INFINITY, controller.signal);
 
 	pipeOutputAsync(spawned, stdioStreamsGroups);
+	cleanupOnExit(spawned, options, controller);
```
ehmicky, line 159: Move this earlier since it is more related to `spawned.kill()` (the next line) that the logic in `lib/stream.js`. Also, it can be done synchronously.

lib/stream.js
```diff
@@ -173,14 +171,12 @@ export const getSpawnedResult = async ({
...
-		return await Promise.all([
+		return Promise.all([
 			error,
 			waitForFailedProcess(processSpawnPromise, processErrorPromise, processExitPromise),
 			Promise.all(stdioPromises.map(stdioPromise => getBufferedData(stdioPromise, encoding))),
 			getBufferedData(allPromise, encoding),
 			Promise.allSettled(customStreamsEndPromises),
 		]);
-	} finally {
-		controller.abort();
```
ehmicky, line 184: `finally` is not needed there. This function's `catch` block is guaranteed not to throw.
Moving the `controller` outside so that `lib/stream.js` only focuses on streams and process completion.

PR #780, opened 2024-02-03
Improve transform speed

ehmicky:
This improves the speed of transforms when the `transform` function is synchronous, which I am assuming should be the vast majority of the cases.

The performance improvement is quite big. On my machine, using a child process with some big output (1MB), the transform runs 4x faster.

### DIFF
--- a/lib/stdio/transform.js
+++ b/lib/stdio/transform.js
@@ -3,21 +3,30 @@ import {callbackify} from 'node:util';
 
 // Transform an array of generator functions into a `Transform` stream.
 // `Duplex.from(generator)` cannot be used because it does not allow setting the `objectMode` and `highWaterMark`.
-export const generatorsToTransform = (generators, {writableObjectMode, readableObjectMode}) => new Transform({
-	writableObjectMode,
-	writableHighWaterMark: getDefaultHighWaterMark(writableObjectMode),
-	readableObjectMode,
-	readableHighWaterMark: getDefaultHighWaterMark(readableObjectMode),
-	transform(chunk, encoding, done) {
-		pushChunks(transformChunk.bind(undefined, chunk, generators, 0), this, done);
-	},
-	flush(done) {
-		pushChunks(finalChunks.bind(undefined, generators), this, done);
-	},
-});
+export const generatorsToTransform = (generators, {transformAsync, finalAsync, writableObjectMode, readableObjectMode}) => {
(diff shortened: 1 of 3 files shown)


lib/stdio/transform.js
```diff
@@ -51,3 +60,45 @@ const generatorFinalChunks = async function * (final, index, generators) {
 		yield * transformChunk(finalChunk, generators, index + 1);
 	}
 };
+
+// Duplicate the code above but as synchronous functions.
+// This is a performance optimization when the `transform`/`flush` function is synchronous, which is the common case.
```
ehmicky, line 65: Duplicating the code is unfortunate. All of the following lines are just copies of the lines above, but without `async`/`await`. Since we're iterating over possibly millions of chunks, unnecessarily waiting for millions of ticks ends up being very slow for synchronous functions.

I tried to think of a smart way to get the same benefit without duplicating the code, but could not find one. For example, I've tried the following instead:

```js
const transformChunk = async function * (chunk, generators, index) {
	if (index === generators.length) {
		yield chunk;
		return;
	}

	const {transform, transformAsync} = generators[index];
  if (transformAsync) {
	  for await (const transformedChunk of transform(chunk)) {
		  yield * transformChunk(transformedChunk, generators, index + 1);
	  }
  } else {
	  for (const transformedChunk of transform(chunk)) {
		  yield * transformChunk(transformedChunk, generators, index + 1);
	  }
  }
};
```

But this does not give much performance benefits. This is because `transformChunk` still needs to be `async`, therefore the `yield *` statement iterates asynchronously, even when the `for` loop is synchronous.

So the only way to get the performance benefit is to make that function not `async`, which ends up requiring duplicating all the code.

The performance benefit is quite big, so I thought it was worth considering.

It turns out Node.js core team might actually have come to the same conclusion when implementation `Readable.from()`, which also involves bridging iterators and streams. They go even one step further and end up [duplicating their code](https://github.com/nodejs/node/blob/6f85b13c8657a5c01b2c66b5185df9ea2631ed20/lib/internal/streams/from.js#L90) in 3 instances.

>  // There are a lot of duplication here, it's done on purpose for performance
>  // reasons - avoid await when not needed.

I wish JavaScript had a version of the `await` keyword that does not wait for one tick when the argument is not a promise, as a performance optimization (at the expense of knowing whether a tick has passed or not).

PR #783, opened 2024-02-03
Cancel generators when the process errors

### LINKED ISSUE #774: Cancel generators when the process exits
When passing a transform to the `stdin` option, the generator function starts running. 
If the process errors, the generator should stop the next time `yield` is called. 
Currently, this is not the case. The generator keeps running until completion.

This bug could make the child process hang forever, for example when using a transform that generates values infinitely (like the Unix command yes).

ehmicky:
Fixes #774.


ehmicky:
Ready for review. :+1:

PR #787, opened 2024-02-04
Fix error handling of `stdin`

### LINKED ISSUE #786: Aborting `childProcess.stdin` leaves the child process hanging
Aborting `childProcess.stdin` does not make the child process end, which can leave it hanging.

```js
const childProcess = execa('forever.js');
childProcess.stdin.destroy();
await childProcess; // Never resolves
```

This applies also to `childProcess.stdio[3]` when using more than one file descriptor, and that file descriptor is used for reading (not writing).

ehmicky:
Fixes #786.

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,6 +155,7 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	const controller = new AbortController();
 	setMaxListeners(Number.POSITIVE_INFINITY, controller.signal);
 
+	const originalStreams = [...spawned.stdio];
 	pipeOutputAsync(spawned, stdioStreamsGroups, controller);
 	cleanupOnExit(spawned, options, controller);
 
@@ -162,12 +163,12 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	spawned.all = makeAllStream(spawned, options);
 	spawned.pipe = pipeToProcess.bind(undefined, {spawned, stdioStreamsGroups, options});
 
-	const promise = handlePromise({spawned, options, stdioStreamsGroups, command, escapedCommand, controller});
+	const promise = handlePromise({spawned, options, stdioStreamsGroups, originalStreams, command, escapedCommand, controller});
 	mergePromise(spawned, promise);
 	return spawned;
 }
 
-const handlePromise = async ({spawned, options, stdioStreamsGroups, command, escapedCommand, controller}) => {
+const handlePromise = async ({spawned, options, stdioStreamsGroups, originalStreams, command, escapedCommand, controller}) => {
 	const context = {timedOut: false};
 
 	const [
@@ -175,7 +176,7 @@ const handlePromise = async ({spawned, options, stdioStreamsGroups, command, esc
 		[, exitCode, signal],
 		stdioResults,
 		allResult,
-	] = await getSpawnedResult({spawned, options, context, stdioStreamsGroups, controller});
+	] = await getSpawnedResult({spawned, options, context, stdioStreamsGroups, originalStreams, controller});
 	controller.abort();
 
 	const stdio = stdioResults.map(stdioResult => handleOutput(options, stdioResult));
--- a/lib/stdio/async.js
+++ b/lib/stdio/async.js
@@ -42,7 +42,7 @@ export const pipeOutputAsync = (spawned, stdioStreamsGroups, controller) => {
 
 	for (const stdioStreams of stdioStreamsGroups) {
 		for (const generatorStream of stdioStreams.filter(({type}) => type === 'generator')) {
-			pipeGenerator(spawned, generatorStream, controller);
+			pipeGenerator(spawned, generatorStream);
 		}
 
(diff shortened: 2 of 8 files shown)


lib/stdio/pipeline.js
```diff
@@ -45,15 +44,6 @@ const handleDestinationComplete = async (source, destination, {signal}) => {
...
 	} catch {} finally {
-		await destroyStream(source);
+		source.destroy();
 	}
 };
-
-// Propagating errors across different streams in the same pipeline can create race conditions.
-// For example, a `Duplex` stream might propagate an error on its writable side and another on its readable side.
-// This leads to different errors being thrown at the top-level based on the result of that race condition.
-// We solve this by waiting for one macrotask with `setImmediate()`.
```
ehmicky, line 55: That race condition was actually due to the problem solved by this PR.

lib/stream.js
```diff
@@ -83,20 +95,30 @@ const resumeStream = async stream => {
...
+// But ignores the `childProcess.stdin.destroy()` automatically done on child process exit by Node.js.
+// `controller.signal` is used to ensure this never resolves, only rejects.
+const throwOnInputStdioError = ({spawned, stdioStreamsGroups, processExitPromise, controller}) => spawned.stdio
+	.filter((stream, index) => stream !== null && stdioStreamsGroups[index][0].direction === 'input')
+	.map(stream => throwOnInputStreamError(stream, processExitPromise, controller));
+
+const throwOnInputStreamError = async (stream, processExitPromise, {signal}) => {
+	await Promise.all([
+		Promise.race([
+			processExitPromise,
```
ehmicky, line 108: This is due to:

https://github.com/nodejs/node/blob/eca4b1a403959d5b83fb5a9adaae8f4c3580eea2/lib/internal/child_process.js#L278

Calling `childProcess.stdin.destroy()` while a child process is running should be handled as abnormal, since that's aborting its input.
However, Node.js calls `childProcess.stdin.destroy()` when the child process ends, and that is normal. 
Therefore, we use `Promise.race()` so that `finished()` does nothing once the child process has ended.

lib/stream.js
```diff
@@ -83,20 +95,30 @@ const resumeStream = async stream => {
...
+			finished(stream, {cleanup: true, readable: false, writable: true}),
+		]),
+		once(signal, 'abort'),
```
ehmicky, line 111: This line is just meant to ensure the function never resolves, since its only purpose is to potentially throw.

lib/stream.js
sindresorhus: Maybe add this as a short code comment.

lib/stream.js
sindresorhus: Maybe add this as a short code comment.

lib/stream.js
ehmicky: Fixed :+1:

lib/stream.js
ehmicky: Fixed :+1:

index.js
```diff
@@ -155,27 +155,28 @@ export function execa(rawFile, rawArgs, rawOptions) {
 	const controller = new AbortController();
 	setMaxListeners(Number.POSITIVE_INFINITY, controller.signal);
 
+	const originalStreams = [...spawned.stdio];
```
ehmicky, line 158: Transforms directly modify `spawned.stdin|stdout|stderr`. We need to keep it track of the original values.

lib/stdio/async.js
```diff
@@ -88,10 +88,9 @@ const setStandardStreamMaxListeners = (stream, {signal}) => {
 };
 
 // `source.pipe(destination)` adds at most 1 listener for each event.
-// We also listen for the stream's completion using `finished()`, which adds 1 more listener.
 // If `stdin` option is an array, the values might be combined with `merge-streams`.
 // That library also listens for `source` end, which adds 1 more listener.
-const maxListenersIncrement = 3;
```
ehmicky, line 94: Value changed since we are not calling `finished(stream)` on `process.std*` anymore.

lib/stream.js
```diff
@@ -83,19 +102,18 @@ const resumeStream = async stream => {
...
+const waitForCustomStreamsEnd = (stdioStreamsGroups, waitForStream) => stdioStreamsGroups.flat()
 	.filter(({type, value}) => type !== 'native' && !isStandardStream(value))
-	.map(({value}) => finished(value, {cleanup: true}));
-
-const throwIfStreamError = stream => stream === null ? [] : [throwOnStreamError(stream)];
-
-const throwOnStreamError = async stream => {
-	const [error] = await once(stream, 'error');
-	throw error;
-};
```
ehmicky, line 98: Previously we only listened to the `error` event for `childProcess.stdin`. 
Now we wait for its completion, which automatically happens when the child process exits.

ehmicky:
Fixed. :+1: 
I reworked the PR too a little.

PR #790, opened 2024-02-06
Fix child process not handling multiple `error` events

### LINKED ISSUE #789: The child process not handle multiple `error` events
The parent process crashes with an uncaught exception when the child process receives multiple `error` events.

```js
const abortController = new AbortController();
const childProcess = execa('...', {signal: abortController.signal});
await once(childProcess, 'spawn');
childProcess.emit('error', new Error('test')); // This is handled
await setImmediate();
abortController.abort(); // This creates an uncaught exception
await childProcess;
```

ehmicky:
Fixes #789.

### DIFF
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,4 +1,4 @@
-import {addAbortListener} from 'node:events';
+import {once} from 'node:events';
 import {finished} from 'node:stream/promises';
 import {setImmediate} from 'node:timers/promises';
 import getStream, {getStreamAsArrayBuffer, getStreamAsArray} from 'get-stream';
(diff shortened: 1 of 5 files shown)


lib/stream.js
```diff
@@ -115,49 +115,49 @@ const waitForCustomStreamsEnd = (stdioStreamsGroups, waitForStream) => stdioStre
...
+	return exitPayload.status === 'rejected'
+		? waitForProcessExit(spawned)
+		: exitPayload.value;
+};
+
+const waitForProcessExit = async spawned => {
+	try {
+		return await once(spawned, 'exit');
+	} catch {
+		return waitForProcessExit(spawned);
```
ehmicky, line 159: Once `spawn` has been emitted, `exit` is guaranteed to be emitted, even if `error` is emitted. 
I.e. no possible infinite recursion.

test/stdio/async.js
```diff
@@ -65,6 +65,7 @@ const testMaxListeners = async (t, isMultiple, maxListenersCount) => {
 			Array.from({length: processesCount}, () => execa('empty.js', getComplexStdio(isMultiple))),
 		);
 		await setImmediate();
+		await setImmediate();
```
ehmicky, line 68: Brittle test that depends on the number of ticks needed by Execa to spawn a child process. :-/

PR #791, opened 2024-02-06
Split files

ehmicky:
The same file is currently handling the logic for `result.command|escapedCommand`, `execaCommand()` and `$`.

Those are 3 separate features, so this PR splits those in their own files. 
This only split those files, without changing them.


sindresorhus:
Feel free to commit things like this directly.

PR #794, opened 2024-02-07
Add `ipc` option

### LINKED ISSUE #782: `ipc` option
When using Node.js IPC feature (which is pretty nice), I sometimes find having to add `'ipc'` at the end of the `stdio` array slightly cumbersome. It is also potentially error-prone, since one needs to remember for it to be the 4th item (or later), and not to repeat it (as some users might be tempted to). Also, it forces users to use the `stdio` option instead of the simpler `stdin`/`stdout`/`stderr` options Execa provides.

Moreover, while `ipc` does use a file descriptor, it does so in a different way than other `stdio` values.
  - The wire format is undocumented, so that file descriptor is not really something users directly use. It is more of an implementation detail.
  - Only one file descriptor can use it.
  - With Execa, it cannot be combined with other values for a given file descriptor. 

I am wondering whether Node.js should have exposed this feature using an `{ ipc: true }` option instead? Should we add that option, which would automatically add `ipc` to the underlying `stdio` option? We would then undocument passing `ipc` to the `stdio` option, but still support it for compatibility.

What are your thoughts about this?

ehmicky:
Fixes #782.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -309,7 +309,6 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 	- `'pipe'`: Sets [`childProcess.stdin`](https://nodejs.org/api/child_process.html#subprocessstdin) stream.
 	- `'overlapped'`: Like `'pipe'` but asynchronous on Windows.
 	- `'ignore'`: Do not use `stdin`.
-	- `'ipc'`: Sets an [IPC channel](https://nodejs.org/api/child_process.html#subprocesssendmessage-sendhandle-options-callback). You can also use `execaNode()` instead.
 	- `'inherit'`: Re-use the current process' `stdin`.
 	- an integer: Re-use a specific file descriptor from the current process.
 	- a Node.js `Readable` stream.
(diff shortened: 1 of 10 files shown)


index.d.ts
```diff
@@ -309,7 +309,6 @@ type CommonOptions<IsSync extends boolean = boolean> = {
 	- `'pipe'`: Sets [`childProcess.stdin`](https://nodejs.org/api/child_process.html#subprocessstdin) stream.
 	- `'overlapped'`: Like `'pipe'` but asynchronous on Windows.
 	- `'ignore'`: Do not use `stdin`.
-	- `'ipc'`: Sets an [IPC channel](https://nodejs.org/api/child_process.html#subprocesssendmessage-sendhandle-options-callback). You can also use `execaNode()` instead.
```
ehmicky, line 312: Undocument passing `ipc` to the `stdio` option, but still support it.

index.js
```diff
@@ -286,6 +288,12 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 	};
 }
 
+const validateSyncOptions = ({ipc}) => {
+	if (ipc) {
+		throw new TypeError('The "ipc: true" option cannot be used with synchronous methods.');
```
ehmicky, line 293: `child_process.spawnSync()` does not allow IPC. It would not make sense anyway, since `childProcess.send()` would not be available.

PR #796, opened 2024-02-07
Fix return value being re-used between multiple calls 

### LINKED ISSUE #795: Return value is re-used between multiple calls
When two `execa()` calls fail, the second return value re-uses the first one, providing the underlying error instance is the same. This means the second call incorrectly re-uses the first call's return value.

Execa calls being rejected by the same error instance can happen in a number of cases, e.g. when emitting an `error` event on the child process, or (more likely) when their `stdin`/`stdout`/`stderr` option share the same stream or transform.

```js
const error = new Error('test');

const firstChildProcess = execa('echo', ['one'], {reject: false});
firstChildProcess.emit('error', error);
const firstError = await firstChildProcess;
console.log(firstError); // Command failed: echo one

const secondChildProcess = execa('echo', ['two'], {reject: false});
secondChildProcess.emit('error', error);
const secondError = await secondChildProcess;
console.log(secondError); // Command failed: echo one (instead of "echo two")
```

This is becoming more of a problem right now due to the ongoing improvement we are doing of the `childProcess.pipe(otherChildProcess)` method, since that method has a higher risk of both child processes sharing the same error instance. For example, any error on their standard stream might result in that situation.

ehmicky:
Fixes #795.

### TLDR

Cloning an error is tricky, and this solves an edge case for Execa, so this PR does the bare minimum.

### Details

By cloning the error, we are losing the `stack`, `name`, `cause`, `errors`, and any user-set properties. Unfortunately, setting those is not straight-forward. Some time ago, I have created the [`set-error-message`](https://github.com/ehmicky/set-error-message), [`wrap-error-message`](https://github.com/ehmicky/wrap-error-message), [`set-error-class`](https://github.com/ehmicky/set-error-class), [`set-error-stack`](https://github.com/ehmicky/set-error-stack) and [`set-error-props`](https://github.com/ehmicky/set-error-props) libraries for this specific problem. 

For example, on V8, `error.stack` is memoized after being accessed first. So when updating `error.message`, `error.stack` might remain unchanged, even though its first lines include the error message. One can set `error.stack` manually, but it can get tricky since Node.js source maps prepends it, which makes string replacement difficult.

Also, one must be careful after setting the non-enumerability of `error.*` correctly since this results in errors being printed weirdly otherwise.

Setting `error.name` has the same issue as setting `error.message` since `error.name` is included (usually) in the beginning of `error.stack`. Also, it is actually incorrect to set `error.name`: `ErrorClass.prototype.name` should be used instead, otherwise the error gets printed weirdly as well. In other words, it should be an inherited property, not an own property. However, this requires changing an error's prototype, which has its own pitfalls.

Then, the error itself might have a wide range of abnormal shapes and problems. I have created a library [`normalize-exception`](https://github.com/ehmicky/normalize-exception) which fixes it. It has a long list of [potential issues](https://github.com/ehmicky/normalize-exception?tab=readme-ov-file#features).

Finally, the cloning logic itself has its own caveats. I also have created a library for it ([`error-serializer`](https://github.com/ehmicky/error-serializer)). It enables some good cloning behavior, handling many edge cases. It includes the libraries above, so cloning is just:

```js
import { serialize, parse } from 'error-serializer'

const error = new TypeError('example')
const newError = parse(serialize(error)) // Error instance: 'TypeError: example ...'
```

### Implementation in this PR

However, to avoid introducing a new dependency, or re-implementing it (which is a rabbit hole), I have kept the logic in this PR very simple. We only keep the `error.message` and do not handle any potential issue with the exception being handled.

On the plus side:
  - In the vast majority of cases, exceptions thrown inside Execa are done by the `child_process` core module, which creates individual error instances.
  - When processes are run individually or serially (which is the main usage), they tend not to share exceptions being thrown.    
  - Even if this cloning does happen, it means another Execa call previously failed with the full original error instance. So users will still have access to it, just not on each failing Execa call.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -61,23 +61,24 @@ export const makeError = ({
 	signal = signal === null ? undefined : signal;
 	const signalDescription = signal === undefined ? undefined : signalsByName[signal].description;
 
-	const errorCode = error && error.code;
-
+	const errorCode = error?.code;
 	const prefix = getErrorPrefix({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled});
 	const execaMessage = `Command ${prefix}: ${command}`;
-	const isError = Object.prototype.toString.call(error) === '[object Error]';
-	const shortMessage = isError ? `${execaMessage}\n${error.message}` : execaMessage;
+	const originalMessage = previousErrors.has(error) ? error.originalMessage : String(error?.message ?? error);
+	const shortMessage = error === undefined ? execaMessage : `${execaMessage}\n${originalMessage}`;
 	const messageStdio = all === undefined ? [stdio[2], stdio[1]] : [all];
 	const message = [shortMessage, ...messageStdio, ...stdio.slice(3)]
 		.map(messagePart => stripFinalNewline(serializeMessagePart(messagePart)))
 		.filter(Boolean)
 		.join('\n\n');
 
-	if (isError) {
-		error.originalMessage = error.message;
-		error.message = message;
-	} else {
+	if (Object.prototype.toString.call(error) !== '[object Error]' || previousErrors.has(error)) {
 		error = new Error(message);
+		error.originalMessage = originalMessage;
+	} else {
+		previousErrors.add(error);
+		error.message = message;
+		error.originalMessage = originalMessage;
 	}
 
 	error.shortMessage = shortMessage;
@@ -106,3 +107,5 @@ export const makeError = ({
 
 	return error;
 };
+
+const previousErrors = new WeakSet();
--- a/test/error.js
+++ b/test/error.js
@@ -4,6 +4,7 @@ import {execa, execaSync} from '../index.js';
 import {FIXTURES_DIR, setFixtureDir} from './helpers/fixtures-dir.js';
 import {fullStdio, getStdio} from './helpers/stdio.js';
 import {noopGenerator, outputObjectGenerator} from './helpers/generator.js';
+import {foobarString} from './helpers/input.js';
 
 const isWindows = process.platform === 'win32';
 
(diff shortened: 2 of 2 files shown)


lib/error.js
```diff
@@ -61,23 +61,24 @@ export const makeError = ({
...
 	const signalDescription = signal === undefined ? undefined : signalsByName[signal].description;
 
-	const errorCode = error && error.code;
-
+	const errorCode = error?.code;
 	const prefix = getErrorPrefix({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled});
 	const execaMessage = `Command ${prefix}: ${command}`;
-	const isError = Object.prototype.toString.call(error) === '[object Error]';
-	const shortMessage = isError ? `${execaMessage}\n${error.message}` : execaMessage;
+	const originalMessage = previousErrors.has(error) ? error.originalMessage : String(error?.message ?? error);
```
ehmicky, line 67: `String(Symbol())` does not throw, but `${Symbol()}` does. That's a very very tiny edge case, but does not cost much to do.

test/error.js
```diff
@@ -292,3 +293,66 @@ test('error.cwd is undefined on failure if not passed as options', async t => {
...
+	const childProcess = execa('empty.js');
+	childProcess.emit('error', error);
+	const {message, originalMessage} = await t.throwsAsync(childProcess);
+	t.true(message.includes(String(error)));
+	t.is(originalMessage, String(error));
+};
+
+test('error instance can be null', testUnusualError, null);
+test('error instance can be false', testUnusualError, false);
+test('error instance can be a string', testUnusualError, 'test');
```
ehmicky, line 307: Libraries throwing a string unfortunately happens more than we'd hope. :/

test/error.js
```diff
@@ -292,3 +293,66 @@ test('error.cwd is undefined on failure if not passed as options', async t => {
...
+test('error instance can be a function', testUnusualError, () => {});
+test('error instance can be an array', testUnusualError, ['test', 'test']);
+
+test('error instance can be undefined', async t => {
+	const childProcess = execa('empty.js');
+	childProcess.emit('error');
+	await t.throwsAsync(childProcess, {message: 'Command failed: empty.js'});
+});
+
+test('error instance can be a plain object', async t => {
```
ehmicky, line 320: I have seen this too. For example, `js-yaml` [used to do that](https://github.com/nodeca/js-yaml/commit/acb734f8daf9ce8094b3f48d0b8d92b832a42153) (although those were not plain objects but non-error object instances).

Issue #798, opened 2024-02-07
`error.cwd` is set but not `result.cwd`


ehmicky:
The `result.cwd` property is typed and documented, but it is missing (unlike the `error.cwd` property).

PR #801, opened 2024-02-07
Re-order return value properties

ehmicky:
This re-orders the properties of the Execa return value (on success or error). When the result is printed, that order is preserved, so this PR makes them follow a logical order.

Also, this adds a test to make sure we are not missing properties, which avoids bugs like https://github.com/sindresorhus/execa/issues/798.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -84,26 +84,28 @@ export const makeError = ({
 	error.shortMessage = shortMessage;
 	error.command = command;
 	error.escapedCommand = escapedCommand;
-	error.exitCode = exitCode;
-	error.signal = signal;
-	error.signalDescription = signalDescription;
-	error.stdio = stdio;
-	error.stdout = stdio[1];
-	error.stderr = stdio[2];
 	error.cwd = cwd;
-
-	if (all !== undefined) {
-		error.all = all;
-	}
-
-	if ('bufferedData' in error) {
-		delete error.bufferedData;
-	}
 
 	error.failed = true;
 	error.timedOut = Boolean(timedOut);
 	error.isCanceled = isCanceled;
 	error.isTerminated = signal !== undefined;
+	error.exitCode = exitCode;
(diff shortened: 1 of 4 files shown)


lib/error.js
```diff
@@ -84,27 +84,29 @@ export const makeError = ({
 	error.shortMessage = shortMessage;
```
ehmicky, line 84: `error.*message` is the most important, so should come first.

lib/error.js
```diff
@@ -84,27 +84,29 @@ export const makeError = ({
...
 	error.command = command;
 	error.escapedCommand = escapedCommand;
+	error.cwd = cwd;
```
ehmicky, line 87: The command's request properties (`command` and `cwd`) come before the result ones (`exitCode`, etc.).

lib/error.js
```diff
@@ -84,27 +84,29 @@ export const makeError = ({
...
+
+	error.failed = true;
+	error.timedOut = Boolean(timedOut);
+	error.isCanceled = isCanceled;
+	error.isTerminated = signal !== undefined;
 	error.exitCode = exitCode;
 	error.signal = signal;
 	error.signalDescription = signalDescription;
```
ehmicky, line 95: Boolean result properties like `timedOut` come before more descriptive ones like `exitCode` and `signal`.

readme.md
```diff
@@ -362,13 +362,11 @@ Same as [`command`](#command-1) but escaped.
...
-#### exitCode
-
-Type: `number | undefined`
+#### cwd
 
-The numeric exit code of the process that was run.
+Type: `string`
 
-This is `undefined` when the process could not be spawned or was terminated by a [signal](#signal-1).
+The `cwd` of the command if provided in the [command options](#cwd-1). Otherwise it is `process.cwd()`.
```
ehmicky, line 369: Re-orders the documentation too.

PR #803, opened 2024-02-08
Improve `cwd` option

### LINKED ISSUE #798: `error.cwd` is set but not `result.cwd`

### LINKED ISSUE #799: Better error message when `cwd` is invalid
`process.cwd()` throws when the current directory does not exist.

```
Error: ENOENT: no such file or directory, uv_cwd
    at process.wrappedCwd [as cwd] (node:internal/bootstrap/switches/does_own_process_state:138:28) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'uv_cwd'
}
```

`child_process.spawn()` also throws when the `cwd` option's directory does not exist.

```
Error: spawnSync echo ENOENT
    at __node_internal_captureLargerStackTrace (node:internal/errors:563:5)
    at __node_internal_errnoException (node:internal/errors:690:12)
    ... // 20 lines of stack trace
    at TTY.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawnSync echo',
  path: 'echo',
  spawnargs: []
},
```

There are also other reasons which can make `cwd` fail such as missing file permissions, pointing to a non-directory, or using a file name too long (libuv uses [`getcwd`](https://www.man7.org/linux/man-pages/man3/getcwd.3.html#ERRORS) on Unix and [`GetCurrentDirectory`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcurrentdirectory) on Windows).

When the above problems happen, we should provide with better error messages. The current error messages are confusing as they do not always mention `cwd`, making it look like the issue is the command not existing. I have actually gotten this myself in the past, and it can lead to long debugging sessions before realizing what the problem is.

ehmicky:
Fixes #798.
Fixes #799.

### DIFF
--- a/lib/cwd.js
+++ b/lib/cwd.js
@@ -1,0 +1,49 @@
+import {statSync} from 'node:fs';
+import {resolve} from 'node:path';
+import {fileURLToPath} from 'node:url';
+import process from 'node:process';
+
+export const getDefaultCwd = () => {
+	try {
+		return process.cwd();
+	} catch (error) {
+		error.message = `The current directory does not exist.\n${error.message}`;
+		throw error;
+	}
+};
+
+export const normalizeCwd = cwd => {
+	const cwdString = safeNormalizeFileUrl(cwd, 'The "cwd" option');
+	return resolve(cwdString);
+};
+
+export const safeNormalizeFileUrl = (file, name) => {
+	const fileString = normalizeFileUrl(file);
+
+	if (typeof fileString !== 'string') {
+		throw new TypeError(`${name} must be a string or a file URL: ${fileString}.`);
+	}
+
+	return fileString;
+};
+
+export const normalizeFileUrl = file => file instanceof URL ? fileURLToPath(file) : file;
+
+export const fixCwdError = (originalMessage, cwd) => {
+	if (cwd === getDefaultCwd()) {
+		return originalMessage;
+	}
+
+	let cwdStat;
+	try {
+		cwdStat = statSync(cwd);
+	} catch (error) {
+		return `The "cwd" option is invalid: ${cwd}.\n${error.message}\n${originalMessage}`;
+	}
+
+	if (!cwdStat.isDirectory()) {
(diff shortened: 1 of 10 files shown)


lib/cwd.js
```diff
@@ -0,0 +1,49 @@
...
+export const normalizeFileUrl = file => file instanceof URL ? fileURLToPath(file) : file;
+
+export const fixCwdError = (originalMessage, cwd) => {
+	if (cwd === getDefaultCwd()) {
+		return originalMessage;
+	}
+
+	let cwdStat;
+	try {
+		cwdStat = statSync(cwd);
```
ehmicky, line 39: This function is used both by `execa()` and `execaSync()`.
In principle, we could use two separate methods, one async and one sync. This would require having two versions of `makeError()`, one async and one sync.
However, this function happens only when both:
  - the child process failed 
  - the `cwd` option was set with a different value than `process.cwd()`

So I thought it might be ok to keep it simple for the time being.

lib/cwd.js
```diff
@@ -0,0 +1,49 @@
...
+	if (typeof fileString !== 'string') {
+		throw new TypeError(`${name} must be a string or a file URL: ${fileString}.`);
+	}
+
+	return fileString;
+};
+
+export const normalizeFileUrl = file => file instanceof URL ? fileURLToPath(file) : file;
+
+export const fixCwdError = (originalMessage, cwd) => {
```
ehmicky, line 32: We cannot perform this `cwd` validation check before process spawning because it involves making a `stat` I/O call. Right now, `execa()` spawns processes right away, and introducing a sync I/O call before spawning for every `execa()` call is not worth it, just to improve the `cwd` validation error message.

That being said, since `process.cwd()` is synchronous, when the `cwd` option default value is used, it is validated before process spawning.

sindresorhus:
Conflict

ehmicky:
Fixed. :+1:

PR #804, opened 2024-02-08
Add `node` option

### LINKED ISSUE #781: Re-using the features from `execaNode()`
`execaNode()` has some nice features:
  - prepend `'node'` as the binary argument
  - re-use the current Node.js binary, including its version
  - re-use the current Node.js CLI flags, without getting any issues when using `--inspect`
  - forces `shell: false`

However, those features are not available for users who either:
  - use `$` or `execaCommand()`
  - use synchronous methods like `execaSync()`
  - do not wish to add an `ipc` channel

We could fix this by adding a `{ node: true }` option (any other option name works too), which would do the same thing as what `execaNode()` is doing, except for adding an `ipc` channel.

What do you think?

ehmicky:
Fixes #781.


index.js
```diff
@@ -288,6 +282,8 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 	};
 }
 
+const normalizeSyncOptions = (options = {}) => options.node && !options.ipc ? {...options, ipc: false} : options;
```
ehmicky, line 285: `node: true` can be used with synchronous methods, but `ipc: true` cannot.
Therefore, `ipc: true` is not added when using `node: true` with synchronous methods.

lib/cwd.js
```diff
@@ -0,0 +1,13 @@
...
+	const fileString = normalizeFileUrl(file);
+
+	if (typeof fileString !== 'string') {
+		throw new TypeError(`${name} must be a string or a file URL: ${fileString}.`);
+	}
+
+	return fileString;
+};
+
+export const normalizeFileUrl = file => file instanceof URL ? fileURLToPath(file) : file;
```
ehmicky, line 13: The other ongoing PR #803 has this new file too.

test/test.js
```diff
@@ -58,14 +57,6 @@ if (process.platform === 'win32') {
 	});
 }
 
-const testNullOptions = async (t, execaMethod) => {
-	const {stdout} = await execaMethod('noop.js', [foobarString], null);
-	t.is(stdout, foobarString);
-};
-
-test('Can pass null to options', testNullOptions, execa);
-test('Can pass null to options - sync', testNullOptions, execaSync);
```
ehmicky, line 67: This test was documenting that `execa('file', [...args], null)` currently does not throw.
This PR breaks this behavior. We could still support passing `null`, but it feels like this is not something we should support, and it is undocumented.

sindresorhus: Yeah, it should not be supported.

ehmicky:
Conflict fixed. :+1:

Issue #753, opened 2024-01-28
Fix error propagation of `.pipe()`


ehmicky:
When using `firstProcess.pipe(secondProcess)`, if `firstProcess.stdout` aborts or errors, `secondProcess.stdin` should be aborted. Otherwise, the following hangs forever:

```js
const firstProcess = execa('node', {reject: false})
const secondProcess = execa('node')
await Promise.all([
	firstProcess.pipe(secondProcess),
	firstProcess.stdout.destroy(),
])
```

The first process exits, but the second does not, since its `stdin` is still open, waiting for data.

Issue #745, opened 2024-01-26
Improve promise returned by the `.pipe()` methods


ehmicky:
## Problem

The following currently results in an unhandled promise rejection:

```js
await execa('node', ['--invalidFlag', 'script.js'])
  .pipe(execa('cat'));
```

That's because `.pipe()` returns `execa('cat')`, so only the second child process is `await`ed, not the first one.

## Solution

`.pipe()` should `await` both child processes instead of only the second one. 

The return value can still be the second child process, which is the current behavior.

The returned promise would still resolve with the second child process' result, which is also the current behavior. 
Additionally, it would now set `result.pipedFrom` or `error.pipedFrom` with the first child process' result/error.

## Details

```js
await execa('echo', ['foobar'])
  .pipe('tr', ['o', 'O'])
```

Would return:

```js
{
  stdout: 'fOObar',
  exitCode: 0,
  failed: false,
  ...
  pipedFrom: {
    stdout: 'foobar',
    exitCode: 0,
    failed: false,
    ...
    // `pipedFrom` is recursive, when piping multiple processes
  },
}
```

When the second process fails, `error.pipedFrom` would contain the first process' result and output, which could be very useful for debugging.

When the first process fails, its `error` would be propagated as is, since this is what actually failed.

--- 

What do you think?

sindresorhus:
:+1: Looks like a good idea to me. I cannot think of any downsides.

ehmicky:
I have been working on this feature, and it appears that the `.pipe()` should not return the second child process. Instead, it should only resolve/reject with the second child process' result. 

Taking the following example: `const returnValue = source.pipe(destination)`, the reasons are the following:
  1. It might be confusing to users whether `returnValue` is `source` or `destination`, without looking up the documentation. `destination` makes more sense. But some users might expect `source` since "fluent" interfaces usually return the same value along the chain.
  2. It might be confusing to users whether `returnValue` methods apply to both `source` and `destination`. For example, one might expect that `returnValue.kill()` would terminate both `source` and `destination`.
  3. While `returnValue` and `destination` would resolve to the same value, they would need to use different promises. That's because `returnValue`'s promise should await both `source` and `destination`. This results in an odd situation, where `returnValue` is the same child process as `destination`, and resolve to the same value as `destination`, but uses a different underlying promise. Not only is this confusing, but it is also tricky to implement. Since child processes cannot be cloned, the only way I could think of was to use a `Proxy`. This works, but comes with its own quirks.
[...]

PR #807, opened 2024-02-09
Improve process failure logic

ehmicky:
This PR refactors some logic related to process failure.

Processes fail due to either:
  1. a process `error` event, timeout, a stream failure, or a transform error
  2. a non-0 exit code or a signal termination

Right now, those lead to different code paths. This PR ensures that all types of process failures end up in the same `catch` block.

This does not change behavior, except for some small bug fixes highlighted in the comments below. The main purpose is to make it easier to add logic which handles any type of process failure. This is something I am currently needing as part of implementing #753 and #745.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ import {makeError} from './lib/error.js';
 import {handleNodeOption} from './lib/node.js';
 import {handleInputAsync, pipeOutputAsync, cleanupStdioStreams} from './lib/stdio/async.js';
 import {handleInputSync, pipeOutputSync} from './lib/stdio/sync.js';
-import {spawnedKill, validateTimeout, normalizeForceKillAfterDelay, cleanupOnExit} from './lib/kill.js';
+import {spawnedKill, validateTimeout, normalizeForceKillAfterDelay, cleanupOnExit, isFailedExit} from './lib/kill.js';
 import {pipeToProcess} from './lib/pipe.js';
 import {getSpawnedResult, makeAllStream} from './lib/stream.js';
 import {mergePromise} from './lib/promise.js';
@@ -169,7 +169,7 @@ const handlePromise = async ({spawned, options, stdioStreamsGroups, originalStre
 	const context = {timedOut: false};
 
 	const [
-		error,
+		errorInfo,
 		[exitCode, signal],
 		stdioResults,
 		allResult,
@@ -179,10 +179,10 @@ const handlePromise = async ({spawned, options, stdioStreamsGroups, originalStre
 	const stdio = stdioResults.map(stdioResult => handleOutput(options, stdioResult));
 	const all = handleOutput(options, allResult);
 
-	if (error || exitCode !== 0 || signal !== null) {
+	if ('error' in errorInfo) {
 		const isCanceled = options.signal?.aborted === true;
 		const returnedError = makeError({
-			error,
+			error: errorInfo.error,
 			exitCode,
 			signal,
 			stdio,
@@ -250,7 +250,7 @@ export function execaSync(rawFile, rawArgs, rawOptions) {
 	const output = result.output || Array.from({length: 3});
 	const stdio = output.map(stdioOutput => handleOutput(options, stdioOutput));
 
-	if (result.error || result.status !== 0 || result.signal !== null) {
+	if (result.error !== undefined || isFailedExit(result.status, result.signal)) {
 		const error = makeError({
 			stdio,
 			error: result.error,
--- a/lib/error.js
+++ b/lib/error.js
@@ -3,28 +3,48 @@ import stripFinalNewline from 'strip-final-newline';
 import {isBinary, binaryToString} from './stdio/utils.js';
 import {fixCwdError} from './cwd.js';
 
-const getErrorPrefix = ({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled}) => {
+export const getErrorPrefix = ({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled}) => {
 	if (timedOut) {
-		return `timed out after ${timeout} milliseconds`;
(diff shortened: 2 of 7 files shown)


lib/kill.js
```diff
@@ -48,10 +48,22 @@ export const normalizeForceKillAfterDelay = forceKillAfterDelay => {
 	return forceKillAfterDelay;
 };
 
+export const waitForSuccessfulExit = async exitPromise => {
+	const [exitCode, signal] = await exitPromise;
```
ehmicky, line 52: `exitPromise` never rejects, and it always resolves.

lib/kill.js
```diff
@@ -65,6 +77,9 @@ export const validateTimeout = ({timeout}) => {
 	}
 };
 
+// Used only to interrupt control flow, but not used in return value
+export const TEMPORARY_ERROR = new Error('Temporary');
```
ehmicky, line 81: For timeouts, non-0 exit codes and signals, we do not want to add anything to the error message. We are also not interested in the stack trace, since the error would be created inside Execa code base. 

Instead, we rely on the main `makeError()` logic to set the right error message, and create some stack trace.

lib/stream.js
```diff
@@ -193,7 +193,7 @@ export const getSpawnedResult = async ({
 	} catch (error) {
 		spawned.kill();
```
ehmicky, line 194: This is a noop when `exitPromise` has resolved.

When either the `exit` or `error` event is emitted, `childProcess._handle` is set to `null`.

https://github.com/nodejs/node/blob/8a41d9b636be86350cd32847c3f89d327c4f6ff7/lib/internal/child_process.js#L282
https://github.com/nodejs/node/blob/8a41d9b636be86350cd32847c3f89d327c4f6ff7/lib/internal/child_process.js#L420

`.kill()` is a noop when `childProcess._handle` is `null`.

https://github.com/nodejs/node/blob/8a41d9b636be86350cd32847c3f89d327c4f6ff7/lib/internal/child_process.js#L498

lib/stream.js
```diff
@@ -181,7 +181,7 @@ export const getSpawnedResult = async ({
 		return await Promise.race([
 			Promise.all([
 				undefined,
-				exitPromise,
+				waitForSuccessfulExit(exitPromise),
```
ehmicky, line 184: Previously, on a non-0 exit code, `Promise.all()` would wait for the streams to complete, then return.
Now, it throws right away, and the `catch` block handles waiting for the streams to complete instead.

The `catch` block ends up doing the exact same thing. 
However, now, that `catch` block can be used to capture any type of error, not just stream errors and process `error` events.

lib/kill.js
```diff
@@ -65,6 +81,11 @@ export const validateTimeout = ({timeout}) => {
 	}
 };
 
+// Indicates that the error is used only to interrupt control flow, but not in the return value
+const setDiscardedError = error => Object.assign(error, {[discardSymbol]: true});
+export const isDiscardedError = error => error?.[discardSymbol] === true;
+const discardSymbol = Symbol('discard');
```
ehmicky, line 87: For timeouts, non-0 exit codes and signals, we do not want to add anything to the error message. We are also not interested in the stack trace, since the error would be created inside the Execa code base. 

Instead, we rely on the main `makeError()` logic to set the right error message, and create some stack trace.

lib/kill.js
```diff
@@ -48,10 +49,25 @@ export const normalizeForceKillAfterDelay = forceKillAfterDelay => {
 	return forceKillAfterDelay;
 };
 
+export const waitForSuccessfulExit = async exitPromise => {
+	const [exitCode, signal] = await exitPromise;
+
+	if (!isProcessErrorExit(exitCode, signal) && isFailedExit(exitCode, signal)) {
+		const message = getErrorPrefix({signal, exitCode});
+		throw setDiscardedError(new Error(message));
```
ehmicky, line 57: When `firstProcess.pipe(secondProcess)` is used and one of the process fails, we terminate the other. The second process's error message indicates it failed due to the first one. 

The second process will then use the error message being set here, to show why the first process failed.

Otherwise (when `.pipe()` is not involved), that error message is unnecessary and is discarded since `makeError()` builds it later on.

lib/kill.js
```diff
@@ -65,6 +81,9 @@ export const validateTimeout = ({timeout}) => {
 	}
 };
 
+// Indicates that the error is used only to interrupt control flow, but not in the return value
+const setDiscardedError = error => Object.assign(error, {discarded: true});
```
ehmicky, line 85: For timeouts, non-0 exit codes and signal terminations, we discard the original error message and stack trace. Instead, we let the main makeError() logic set those later on.

index.js
```diff
@@ -179,10 +179,10 @@ const handlePromise = async ({spawned, options, stdioStreamsGroups, originalStre
 	const stdio = stdioResults.map(stdioResult => handleOutput(options, stdioResult));
 	const all = handleOutput(options, allResult);
 
-	if (error || exitCode !== 0 || signal !== null) {
+	if ('error' in errorInfo) {
```
ehmicky, line 182: Allows handling `childProcess.emit('error')`, which produces an error which is `undefined`.

lib/error.js
```diff
@@ -46,27 +66,21 @@ const serializeMessageItem = messageItem => {
...
-	exitCode,
+	error: rawError,
+	signal: rawSignal,
+	exitCode: rawExitCode,
 	command,
 	escapedCommand,
 	timedOut,
 	isCanceled,
-	options: {timeoutDuration: timeout, cwd},
+	options: {timeoutDuration, timeout = timeoutDuration, cwd},
```
ehmicky, line 76: Fixes a bug where the error message of timeouts with `execaSync()` would show `Command failed with undefined milliseconds`.

ehmicky:
This PR is ready for review. :+1:

PR #809, opened 2024-02-12
Do not prevent process normal exit

### LINKED ISSUE #808: Do not prevent process normal exit
When `childProcess.stdin|stdout|stderr|stdio` is aborted or errored, we currently send `SIGTERM` to `childProcess`. This behavior is creating two problems:
1. It prevents the process from exiting gracefully, as it is expected to.
2. That graceful exit might happen either before or after the OS processed `SIGTERM`. This creates a race condition, where the process might either have `error.exitCode: 0` or `signal: 'SIGTERM'`, based on timing conditions.

To expand on the first point: if the child process reads from a file descriptor (typically: `stdin`), aborting or erroring `childProcess.std*` will stop writing new data, i.e. the child process will stop reading. It then automatically continues its normal execution.

If the child process writes to a file descriptor (typically: `stdout`/`stderr`), aborting or erroring `childProcess.std*` will make the next write fail (usually with `EPIPE`), causing the child process to automatically error and exit.

If the child process neither reads nor writes a file descriptor, the state of that file descriptor becomes irrelevant to the child process execution. Therefore, aborting or erroring `childProcess.std*` should not terminate the child process.

The same reasoning happens for the `error` event on the child process: it is Node.js-specific and should not impact the child process execution.

The only case where `SIGTERM` (or more precisely: the `killSignal`, which defaults to `SIGTERM`) should be sent is when one of the `signal`, `timeout` and `maxBuffer` options triggers.

All of the above (not sending `SIGTERM` except for `signal`/`timeout`/`maxBuffer`) is actually how the core Node.js module `child_process` behaves. I just checked the source code in details. Execa started deviating from that behavior as part of the ongoing (and unreleased yet) changes I've been doing. So this is a good time to fix that erroneous behavior.

ehmicky:
Fixes #808.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
 	},
 	"ava": {
 		"workerThreads": false,
-		"concurrency": 1,
 		"timeout": "60s"
 	},
 	"xo": {
--- a/test/kill.js
+++ b/test/kill.js
@@ -119,24 +119,6 @@ if (isWindows) {
 		t.is(signal, 'SIGKILL');
 	});
 
-	test('`forceKillAfterDelay` works with "error" events on childProcess', async t => {
-		const subprocess = spawnNoKillableSimple();
-		await once(subprocess, 'spawn');
-		subprocess.emit('error', new Error('test'));
-		const {isTerminated, signal} = await t.throwsAsync(subprocess);
-		t.true(isTerminated);
-		t.is(signal, 'SIGKILL');
-	});
-
-	test('`forceKillAfterDelay` works with "error" events on childProcess.stdout', async t => {
-		const subprocess = spawnNoKillableSimple();
-		await once(subprocess, 'spawn');
-		subprocess.stdout.destroy(new Error('test'));
-		const {isTerminated, signal} = await t.throwsAsync(subprocess);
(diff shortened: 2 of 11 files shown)


test/kill.js
```diff
@@ -119,24 +119,6 @@ if (isWindows) {
...
-	});
-
-	test('`forceKillAfterDelay` works with "error" events on childProcess.stdout', async t => {
-		const subprocess = spawnNoKillableSimple();
-		await once(subprocess, 'spawn');
-		subprocess.stdout.destroy(new Error('test'));
-		const {isTerminated, signal} = await t.throwsAsync(subprocess);
-		t.true(isTerminated);
-		t.is(signal, 'SIGKILL');
-	});
```
ehmicky, line 138: Those two tests do not make sense anymore.

package.json
```diff
@@ -80,7 +80,6 @@
 	},
 	"ava": {
 		"workerThreads": false,
-		"concurrency": 1,
```
ehmicky, line 83: It appears this PR removing the race condition made this workaround unnecessary.

PR #811, opened 2024-02-12
Add `.kill(error)`

### LINKED ISSUE #810: Add `childProcess.kill(error)`
Terminating a child process with `childProcess.kill()` does not allow setting an error message or stack trace. This can result in obscure errors which only indicate `Terminated with SIGTERM` and are hard to debug.

It would be nice to allow users to do `childProcess.kill(error)`. It would behave exactly like `childProcess.kill()`, except the error rejected by Execa would be `error`. `error.signal` and `error.isTerminated` would still be set, and `error.message` would still indicate that the child process was terminated by a signal.

It would only be allowed with no signal name, i.e. using the `killSignal` (which defaults to `SIGTERM`), since this is explicitly meant for child process termination, not for any signal. Also, it makes the API simpler.

What do you think?

ehmicky:
Fixes #810.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -325,6 +325,19 @@ A `streamName` can be passed to pipe `"stderr"`, `"all"` (both `stdout` and `std
 
 Returns `execaChildProcess`, which allows chaining `.pipe()` then `await`ing the [final result](#childprocessresult).
 
+#### kill(signalOrError?)
+
+`signalOrError`: `string | number | Error`\
+_Returns_: `boolean`
+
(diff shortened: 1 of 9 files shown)


readme.md
```diff
@@ -325,6 +325,19 @@ A `streamName` can be passed to pipe `"stderr"`, `"all"` (both `stdout` and `std
 
 Returns `execaChildProcess`, which allows chaining `.pipe()` then `await`ing the [final result](#childprocessresult).
 
+#### kill(signalOrError?)
+
+`signalOrError`: `string | number | Error`\
+_Returns_: `boolean`
```
ehmicky, line 331: I am not sure which convention to use to document the return value, please let me know if this works.

sindresorhus: Yeah, that's fine.

PR #812, opened 2024-02-12
Merge `execPath` and `nodePath` options

### LINKED ISSUE #792: Merge `execPath` and `nodePath` options
The `execPath` and `nodePath` options have similarities:
  - The value is a path to the Node.js binary
  - The value can be a string or a file URL
  - The default value is `process.execPath`

However, they also have the following differences:
  - `nodePath` is the Node.js binary used to spawn the child process. This is what Execa directly executes. `execPath` is the Node.js binary used by the child process to spawn other child processes. This is set using the `PATH` environment variable.
  - `execPath` is relative to the `cwd` option (which defaults to `process.cwd()`). `nodePath` is always relative to `process.cwd()`.
  - `nodePath` is used with `execaNode()` (or the new `{node: true}` option). `execPath` is used when the `preferLocal` option is `true`.
  - Their value can be independent from each other. For example, when using `execaNode(..., { nodePath: '.../v16/node' })`, the child process will use Node 16, but its own child processes will use the parent process' Node version.

This might create some confusion. I don't think there is a strong reason for having two separate options. We probably should merge them into a single `nodePath` option.
  - It is the Node.js binary used both by the child processes, and its own child processes.
  - It always takes the `cwd` option into account.
  - It sets the `PATH` environment variable (if `preferLocal` is `true`).
  - With `execaNode()` (or the new `{node: true}` option), it is used to spawn the child process.

What do you think?

ehmicky:
Fixes #792.


lib/escape.js
```diff
@@ -1,21 +1,12 @@
...
+	? `"${arg.replaceAll('"', '\\"')}"`
+	: arg;
 
-	return `"${arg.replaceAll('"', '\\"')}"`;
-};
-
-export const joinCommand = (file, rawArgs) => normalizeArgs(file, rawArgs).join(' ');
-
-export const getEscapedCommand = (file, rawArgs) => normalizeArgs(file, rawArgs).map(arg => escapeArg(arg)).join(' ');
+const NO_ESCAPE_REGEXP = /^[\w.-]+$/;
```
ehmicky, line 12: This refactoring (and the one in `index.js`) is not strictly related to this PR. I can split it into another PR if you want, or leave it in this one if this is ok with you.

test/node.js
```diff
@@ -11,8 +11,12 @@ import {foobarString} from './helpers/input.js';
...
 process.chdir(FIXTURES_DIR);
 
-const runWithNodeOption = (file, args = [], options = {}) => execa(file, args, {...options, node: true});
-const runWithNodeOptionSync = (file, args = [], options = {}) => execaSync(file, args, {...options, node: true});
+const runWithNodeOption = (file, args, options) => Array.isArray(args)
+	? execa(file, args, {...options, node: true})
+	: execa(file, {...options, node: true});
+const runWithNodeOptionSync = (file, args, options) => Array.isArray(args)
+	? execaSync(file, args, {...options, node: true})
+	: execaSync(file, {...options, node: true});
```
ehmicky, line 19: This ensures we are testing both with and without `args`. 
There was a bug with not using `args`, and it did not get caught by the tests due to not trying both code paths.

package.json
```diff
@@ -80,6 +80,7 @@
 	},
 	"ava": {
 		"workerThreads": false,
+		"concurrency": 1,
```
ehmicky, line 83: Unfortunately, it seems like this needs to be brought back. 
Otherwise, Windows tests still seem to be randomly failing in CI: https://github.com/sindresorhus/execa/actions/runs/7878070501/job/21495511609

sindresorhus:
I think this should throw a friendly error about using `nodePath` if the user passes in `execPath`.

ehmicky:
Done in e98c12c. :+1:

PR #814, opened 2024-02-13
Fix `error.stack` with `stream.destroy(error)`

### LINKED ISSUE #813: `error.stack` is not updated when `childProcess.destroy(error)` is called (due to Node.js bug)
When `childProcess.destroy(error)` is called, the error rejected by Execa has a correct `message`, but not a correct `stack`.

The following works well:

```js
const childProcess = execa('echo', {reject: false})
childProcess.stdout.destroy()
const {message, stack} = await childProcess
console.log('MESSAGE')
console.log(message)
console.log('\nSTACK')
console.log(stack)
```

```
MESSAGE
Command failed with ERR_STREAM_PREMATURE_CLOSE: echo
Premature close

STACK
Error [ERR_STREAM_PREMATURE_CLOSE]: Command failed with ERR_STREAM_PREMATURE_CLOSE: echo
Premature close
    at Socket.onclose (node:internal/streams/end-of-stream:154:30)
    at Socket.emit (node:events:527:35)
    at Pipe.<anonymous> (node:net:337:12)
```

But the following stack lacks `Command failed with ...`:

```js
const childProcess = execa('echo', {reject: false})
childProcess.stdout.destroy(new Error('example'))
const {message, stack} = await childProcess
console.log('MESSAGE')
console.log(message)
console.log('\nSTACK')
console.log(stack)
```

```
MESSAGE
Command failed with exit code 0: echo
example

STACK
Error: example
    at file:///home/ether/code/execa/a.js:4:29
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12)
```

This is due to [a bug in Node.js](https://github.com/nodejs/node/issues/51715), which is related to a quirky behavior with V8 that might or might not be a bug.

ehmicky:
Fixes #813.

This PR works around [a bug in Node.js](https://github.com/nodejs/node/issues/51715) which makes `stream.destroy(error)` create Execa errors without a proper error message.

This PR also improves how errors are cloned. When `execa()` returns twice the same error instance, it clones it to avoid directly mutating the first return value. This PR enhances that error cloning so that `error.name`, `error.stack`, `error.cause` and `error.errors` are kept. Previously, they were not kept.

Finally, this PR does some minor refactoring of the error logic, mostly moving lines of code around.

### DIFF
--- a/lib/clone.js
+++ b/lib/clone.js
@@ -50,6 +50,8 @@ const COPIED_ERROR_PROPERTIES = [
 
 // Sets `error.message`.
 // Fixes `error.stack` not being updated when it has been already accessed, since it is memoized by V8.
+// For example, this happens when calling `stream.destroy(error)`.
+// See https://github.com/nodejs/node/issues/51715
 const setErrorMessage = (error, newMessage) => {
 	const {name, message, stack} = error;
 	error.message = newMessage;
--- a/lib/clone.js
+++ b/lib/clone.js
@@ -1,0 +1,68 @@
+export const getFinalError = (initialError, message) => {
+	const error = createFinalError(initialError, message);
+	previousErrors.add(error);
+	return error;
+};
+
+const createFinalError = (error, message) => {
+	if (!isErrorInstance(error)) {
+		return new Error(message);
+	}
+
+	return previousErrors.has(error)
+		? cloneError(error, message)
+		: setErrorMessage(error, message);
+};
+
+export const isErrorInstance = value => Object.prototype.toString.call(value) === '[object Error]';
+
+const cloneError = (oldError, newMessage) => {
+	const {name, message, stack} = oldError;
+	const error = new Error(newMessage);
+	error.stack = fixStack(name, stack, message, newMessage);
+	Object.defineProperty(error, 'name', {value: name, enumerable: false, configurable: true, writable: true});
+	copyErrorProperties(error, oldError);
+	return error;
+};
+
+const copyErrorProperties = (newError, previousError) => {
+	for (const propertyName of COPIED_ERROR_PROPERTIES) {
+		const descriptor = Object.getOwnPropertyDescriptor(previousError, propertyName);
+		if (descriptor !== undefined) {
+			Object.defineProperty(newError, propertyName, descriptor);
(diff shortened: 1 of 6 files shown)


lib/clone.js
```diff
@@ -0,0 +1,66 @@
...
+	return previousErrors.has(error)
+		? cloneError(error, message)
+		: setErrorMessage(error, message);
+};
+
+const cloneError = (oldError, newMessage) => {
+	const {name, message, stack} = oldError;
+	const error = new Error(newMessage);
+	error.stack = fixStack(name, stack, message, newMessage);
+	Object.defineProperty(error, 'name', {value: name, enumerable: false, configurable: true, writable: true});
```
ehmicky, line 21: Unlike the other properties below, `error.name` is usually set on the prototype.

lib/clone.js
```diff
@@ -0,0 +1,66 @@
...
+const setErrorMessage = (error, newMessage) => {
+	const {name, message, stack} = error;
+	error.message = newMessage;
+	error.stack = fixStack(name, stack, message, newMessage);
+	return error;
+};
+
+const fixStack = (name, stack, message, newMessage) => stack.includes(newMessage)
+	? stack
+	: stack.replace(`${name}: ${message}`, `${name}: ${newMessage}`);
```
ehmicky, line 60: This works with `--enable-source-maps`, which usually prepends a preview of the source code at the beginning of `error.stack`.

package.json
```diff
@@ -80,6 +80,7 @@
 	},
 	"ava": {
 		"workerThreads": false,
+		"concurrency": 1,
```
ehmicky, line 83: See https://github.com/sindresorhus/execa/pull/812#discussion_r1486794527

sindresorhus:
Can you add a TODO comment at the part of the code that works around the Node.js issue and link to the Node.js issue there?

sindresorhus:
> So we might not need a TODO?

👍

PR #819, opened 2024-02-17
Simplify logic related to error handling

ehmicky:
This PR simplifies some logic related to error handling.
It does not change the behavior.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -30,7 +30,7 @@ const createMessages = ({
 	return {originalMessage, shortMessage, message};
 };
 
-export const getErrorPrefix = ({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled}) => {
+const getErrorPrefix = ({timedOut, timeout, errorCode, signal, signalDescription, exitCode, isCanceled}) => {
 	if (timedOut) {
 		return `Command timed out after ${timeout} milliseconds`;
 	}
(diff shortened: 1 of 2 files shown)


lib/error.js
```diff
@@ -79,9 +79,12 @@ const serializeMessageItem = messageItem => {
 	return '';
 };
 
+// Indicates that the error is used only to interrupt control flow, but not in the return value
+export class DiscardedError extends Error {}
```
ehmicky, line 83: For example, when a process ends with `SIGTERM`, we throw an error to interrupt the flow. But we haven't constructed the error and its message yet. So we use this dummy error class.

lib/kill.js
```diff
@@ -72,8 +72,7 @@ export const waitForSuccessfulExit = async exitPromise => {
 	const [exitCode, signal] = await exitPromise;
 
 	if (!isProcessErrorExit(exitCode, signal) && isFailedExit(exitCode, signal)) {
-		const message = getErrorPrefix(normalizeExitPayload(exitCode, signal));
-		throw setDiscardedError(new Error(message));
```
ehmicky, line 76: Those `DiscardedError`s used to set some basic error message, because I was thinking this would be used on the upcoming PR related to improving the `.pipe()` method. However, it turns out this is not needed, so this can be removed. It is not used right now.

PR #826, opened 2024-02-17
Fix stream error handling

### LINKED ISSUE #822: Do not error a stream when another stream only aborted
Execa pipes multiple streams internally. For example `{stdin: webStream}` pipes `webStream` to `childProcess.stdin`. Also, transforms pipe with each other and with `childProcess.std*`.

We pipe the source and the destination stream with `source.pipe(destination)`. When either aborts or errors, we destroy the other stream, for cleanup and error handling purposes.

Node.js streams distinguish between aborting (`stream.destroy()`) and erroring (`stream.destroy(error)`). The propagation logic described above currently does not make that distinction. So when a stream is aborted, the other streams are errored instead of being aborted as well.

### LINKED ISSUE #823: Ignore `EPIPE` on writable streams
When piping two processes, the source process usually ends first, closes the destination process' `stdin`, then the destination ends.

However, there are some situations when the destination process decides when to end instead. This usually happens when either: 
  - The source process pipes some infinite stream of data to the destination process. For example, `yes | head -n10`
  - The source process pipes a finite stream of data, but the destination process only needs a subset of it. For example, `cat file.txt | head -n10`.

In a Unix shell: 
  - First, the destination process ends. 
  - Then, on the next write system call the source process does, it receives a `SIGPIPE` signal. 
  - By default, this terminates the source process. 
  - But it can handle that signal, in which case the kernel sends back an `EPIPE` error. The source process can then proceed to do some graceful exit.

Node.js does not provide with out-of-box process piping (I am working on it for Execa! :) ) but it does provide with ways to pipe a source process `stdout` to a destination process `stdin`. What happens in the above situation is: 
  - First, the destination process closes its `stdin`.
  - Then, the next time the source process `stdout` writes to the destination process `stdin`, it receives an `EPIPE` error.
  - This makes the destination process `stdin` error. 
  - This does not make the source process `stdout` error, but it makes it stop reading due to the `unpipe` event.

`EPIPE` seems to be the only way to handle the above situation. In particular, there seems to be no way to know whether the next write call to the destination process `stdin` will result in an `EPIPE` or not. `EPIPE` fails before any event is emitted or any property is set on both the destination `childProcess` and its `childProcess.stdin` stream. For example, `childProcess.stdin.writable` is `true` before that call. 

In other words, it seems that, by design, Node.js considers that `EPIPE` error (which is returned by the underlying `Writable._write()` method) to be the "event" we should listen to in order to know piping is done (in a situation where the destination process decides when to end piping).

This results in the following: `EPIPE` errors should be ignored on `childProcess.stdin`. Moreover, it should be ignored on any writable stream Execa uses, such as any passed to the `stdout` option since those could be `{stdout: childProcess.stdin}`. The stream would still end on `EPIPE`, but this would not be reported as an error.

### LINKED ISSUE #824: Ignore aborts on readable streams
This issue is a consequence of #823. For concision, I will continue this issue where #823 left.

In a nutshell, this related to Execa supporting piping processes, in the specific case where the destination process decides when to end the piping. For example `cat file.txt | head -n10`.

While #823 dealt with the issue with `EPIPE`, there is a correlated issue but on the other end of the piping. When `EPIPE` is emitted on the destination `stdin`, the destination `stdout` stops piping. However, it is not automatically destroyed, i.e. it must be manually aborted using `childProcess.stdout.destroy()`. Doing so ensures proper cleanup.

Generally speaking, aborting a readable stream is not necessarily the result of a problem or a bug. Calling `readable.destroy()` is the only way to end an infinite `readable` stream, or to notify that the reader does not need any more data. This is as opposed to `readable.destroy(error)` which should always be treated as an error.

This results in the following: aborts should not be treated as errors on `childProcess.stdout`. The same applies to any readable stream Execa uses, such as any passed to the `stdin` option since those could be `{stdin: childProcess.stdout}`.

### LINKED ISSUE #825: Handle errors from streams passed to the `std*` option
Users can pass Node.js streams to the `stdin`/`stdout`/`stderr` option.

If that stream fail, the promise returned by Execa should fail. This currently happens for any stream created under-the-hood when passing special values to the `std*` option. For example, when passing `{stdout: {file: 'file.txt'}}` and a file error happens, Execa will reject with it.

However, this error handling currently only happens when the Node.js stream is passed as an array like `{stdout: [stream, 'pipe']}` but not when it is passed on its own like `{stdout: stream}`. We should fix this.

ehmicky:
Fixes #822 
Fixes #823 
Fixes #824 
Fixes #825 

Those are all preparatory issues I detected while working on improving Execa `.pipe()` method (to pipe processes).

### DIFF
--- a/lib/stdio/pipeline.js
+++ b/lib/stdio/pipeline.js
@@ -1,5 +1,6 @@
 import {finished} from 'node:stream/promises';
 import mergeStreams from '@sindresorhus/merge-streams';
+import {isStreamAbort} from '../wait.js';
 import {isStandardStream} from './utils.js';
 
 // Like `Stream.pipeline(source, destination)`, but does not destroy standard streams.
@@ -51,5 +52,9 @@ const destroyStream = (stream, finishedStreams, error) => {
 		return;
 	}
 
-	stream.destroy(error);
+	if (isStreamAbort(error)) {
+		stream.destroy();
+	} else {
+		stream.destroy(error);
+	}
 };
--- a/lib/stream.js
+++ b/lib/stream.js
@@ -1,11 +1,12 @@
 import {once} from 'node:events';
-import {finished} from 'node:stream/promises';
 import {setImmediate} from 'node:timers/promises';
 import getStream, {getStreamAsArrayBuffer, getStreamAsArray, MaxBufferError} from 'get-stream';
+import {isStream} from 'is-stream';
 import mergeStreams from '@sindresorhus/merge-streams';
 import {waitForSuccessfulExit, throwOnTimeout, errorSignal} from './kill.js';
 import {isStandardStream} from './stdio/utils.js';
 import {generatorToDuplexStream} from './stdio/generator.js';
+import {waitForStream, handleStreamError, isInputFileDescriptor} from './wait.js';
 
 // `all` interleaves `stdout` and `stderr`
 export const makeAllStream = ({stdout, stderr}, {all}) => all && (stdout || stderr)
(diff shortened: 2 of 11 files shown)


lib/stdio/pipeline.js
```diff
@@ -12,56 +13,48 @@ export const pipeStreams = (sources, destination) => {
...
-const handleDestinationComplete = async (source, destination, finishedStreams) => {
-	if (isStandardStream(source) || isStandardStream(destination)) {
-		return;
-	}
-
-	try {
-		await onFinishedStream(destination, finishedStreams);
-		destroyStream(source, finishedStreams);
-	} catch (error) {
-		destroyStream(source, finishedStreams, error);
```
ehmicky, line 49: The changes in this file are copied from ongoing PR #821. And also the other changes in that PR, such as the new tests.

lib/stream.js
```diff
@@ -117,26 +121,20 @@ const applyEncoding = (contents, encoding) => encoding === 'buffer' ? new Uint8A
...
-	await Promise.race([
-		...(stream === originalStdin ? [exitPromise] : []),
-		finished(stream, {cleanup: true}),
-	]);
-};
+const waitForCustomStreamsEnd = (stdioStreamsGroups, streamInfo) => stdioStreamsGroups.flatMap((stdioStreams, index) => stdioStreams
+	.filter(({value}) => isStream(value) && !isStandardStream(value))
+	.map(({type, value}) => waitForStream(value, index, streamInfo, {
+		isSameDirection: type === 'generator',
+		stopOnExit: type === 'native',
```
ehmicky, line 136: That fixes https://github.com/sindresorhus/execa/issues/825.

`type: 'native'` related to the Node.js streams described in #825. Before, they were not passed to `waitForStream()`, which handles any errors from streams. Now, it is passed, but stops listening for errors once the process exits. That's because Node.js does not destroy streams passed as options to `child_process.spawn()` once the process has ended.

sindresorhus:
Conflict

ehmicky:
Fixed :+1:

PR #833, opened 2024-02-21
Improve types of `childProcess.stdin`

ehmicky:
This improves the types for `childProcess.stdin` so that, when the `stdin` option is anything but `pipe` (the default value) or `overlapped` (for example `ignore` or `inherit`), `childProcess.stdin` is typed as `null`. This will show at type-time when users are trying to write to `childProcess.stdin` even not available.


sindresorhus:
> childProcess.stdin is typed as null

Would be nice if these could be `undefined`.

PR #836, opened 2024-02-21
Add `.kill(signal, error)`

### LINKED ISSUE #816: Add `.kill(signal, error)`
Users can send `SIGTERM`.

```js
childProcess.kill()
```

Or any signal as argument.

```js
childProcess.kill('SIGKILL')
```

Or using the `killSignal` option.

```js
const childProcess = execa(..., {killSignal: 'SIGKILL'})
childProcess.kill()
```

With #810, we just added specifying an error message/stack.

```js
childProcess.kill(new Error('example'))
```

I initially assumed this would only be needed with the default termination signal (`killSignal` option).
However, I am currently working on a new feature which would require sending `SIGPIPE` + an error message/stack, regardless of the child process's `killSignal`.
Also, I can see that some users might want to both set `SIGKILL` and an error message/stack.

Therefore, we probably should allow the following.

```js
childProcess.kill('SIGKILL', new Error('example'))
```

ehmicky:
Fixes #816.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -851,8 +851,8 @@ export type ExecaChildPromise<OptionsType extends Options = Options> = {
 
 	[More info.](https://nodejs.org/api/child_process.html#subprocesskillsignal)
 	*/
-	kill(signal: Parameters<ChildProcess['kill']>[0], error: Error): ReturnType<ChildProcess['kill']>;
-	kill(error: Error): ReturnType<ChildProcess['kill']>;
+	kill(signal: Parameters<ChildProcess['kill']>[0], error?: Error): ReturnType<ChildProcess['kill']>;
+	kill(error?: Error): ReturnType<ChildProcess['kill']>;
 };
 
 export type ExecaChildProcess<OptionsType extends Options = Options> = ChildProcess &
--- a/index.d.ts
+++ b/index.d.ts
@@ -851,7 +851,8 @@ export type ExecaChildPromise<OptionsType extends Options = Options> = {
 
 	[More info.](https://nodejs.org/api/child_process.html#subprocesskillsignal)
 	*/
-	kill(signalOrError: Parameters<ChildProcess['kill']>[0] | Error): ReturnType<ChildProcess['kill']>;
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -325,9 +325,10 @@ A `streamName` can be passed to pipe `"stderr"`, `"all"` (both `stdout` and `std
 
 Returns `execaChildProcess`, which allows chaining `.pipe()` then `await`ing the [final result](#childprocessresult).
 
-#### kill(signalOrError?)
+#### kill(signal?, error?)
```
sindresorhus, line 328: The current signature does not make it clear you can pass `error` as the first argument.

```suggestion
#### kill(signal, error?)
#### kill(error?)
```

index.d.ts
```diff
@@ -833,7 +833,8 @@ export type ExecaChildPromise<OptionsType extends Options = Options> = {
 
 	[More info.](https://nodejs.org/api/child_process.html#subprocesskillsignal)
 	*/
-	kill(signalOrError: Parameters<ChildProcess['kill']>[0] | Error): ReturnType<ChildProcess['kill']>;
+	kill(signal: Parameters<ChildProcess['kill']>[0], error: Error): ReturnType<ChildProcess['kill']>;
+	kill(error: Error): ReturnType<ChildProcess['kill']>;
```
sindresorhus, line 837:
```suggestion
	kill(signal: Parameters<ChildProcess['kill']>[0], error?: Error): ReturnType<ChildProcess['kill']>;
	kill(error?: Error): ReturnType<ChildProcess['kill']>;
```

?

sindresorhus (changes requested):

index.d.ts
ehmicky: Right! The type tests happened to pass because this type is intersected with `& ChildProcess`, which has the definition of `.kill()` with 0 arguments, but that was by accident.

readme.md
ehmicky: That's better.

ehmicky:
Fixed. :+1:

PR #840, opened 2024-02-22
`.pipe()` shortcut for `$`

### LINKED ISSUE #797: `.pipe()` shortcut with `$`
Using `.pipe()` with `$` is currently quite verbose. It is also error-prone since users need to remember to set the `stdin` option of the second process to `'pipe'` since `$` changes the default value to `'inherit'`.

```js
const {stdout} = await $({preferLocal: true})`npm run docs`
  .pipe($({stdin: 'pipe'})($`grep ${message}`))
  .pipe($({stdin: 'pipe'})($`sort`))
  .pipe($({stdin: 'pipe', lines: true})($`head -n ${count}`));
```

We could introduce a `.pipe` shortcut specifically for `$`. The following would be the same as above.

```js
const {stdout} = await $({preferLocal: true})`npm run docs`
  .pipe`grep ${message}`
  .pipe`sort`
  .pipe({lines: true})`head -n ${count}`;
```

What do you think?

ehmicky:
Fixes #797.

This adds a shortcut syntax so that users can turn this:

```js
const {stdout} = await $({preferLocal: true})`npm run docs`
  .pipe($({stdin: 'pipe'})($`grep ${message}`))
  .pipe($({stdin: 'pipe'})($`sort`))
  .pipe($({stdin: 'pipe', lines: true})($`head -n ${count}`), {signal});
```

Into this instead:

```js
const {stdout} = await $({preferLocal: true})`npm run docs`
  .pipe`grep ${message}`
  .pipe`sort`
  .pipe({lines: true, signal})`head -n ${count}`;
```

### DIFF
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -6,7 +6,7 @@ With Execa, you can write scripts with Node.js instead of a shell language. It i
 import {$} from 'execa';
 
 const {stdout: name} = await $`cat package.json`
-  .pipe($({stdin: 'pipe'})`grep name`);
+  .pipe`grep name`;
 console.log(name);
 
 const branch = await $`git branch --show-current`;
@@ -599,8 +599,8 @@ await $`npm run build | sort | head -n2`;
 ```js
 // Execa
 await $`npm run build`
-	.pipe($({stdin: 'pipe'})`sort`)
-	.pipe($({stdin: 'pipe'})`head -n2`);
+	.pipe`sort`
+	.pipe`head -n2`;
 ```
 
 ### Piping stdout and stderr to another command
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
 		"cross-spawn": "^7.0.3",
 		"get-stream": "^8.0.1",
 		"human-signals": "^6.0.0",
+		"is-plain-obj": "^4.1.0",
```
ehmicky, line 52: Also added by https://github.com/sindresorhus/execa/pull/838

stekovinbranturry:
not published yet?

PR #852, opened 2024-02-26
Rename `signal` option to `unpipeSignal`

### LINKED ISSUE #851: Rename `signal` option to `unpipeSignal` option
There are currently two `signal` options:
  - One passed to `execa()`, which calls `childProcess.kill()` on abort
  - One passed to `childProcess.pipe()`, which calls `childProcess.stdout.unpipe()` on abort

Having those share the same name is confusing.

```js
await execa(...)
  .pipe(execa(..., {signal: thisSignal}), {signal: thatSignal})
```

It also leads to namespace conflict since `$.pipe(options)` allows setting both normal options and pipe options.

```js
await $`...`
  .pipe({signal: thisSignal, signal: thatSignal})`...`
```

The option name `signal` is borrowed from `child_process.spawn()`. It is itself borrowed from how Web APIs call it: `AbortSignal`. In our case, this is unfortunate because OS signals (like `SIGTERM`) are a different concept that shares that word. This means we also return `error.signal` and have `options.killSignal` which use that term, to add to the confusion.

An immediate fix we can do is: we probably should rename the second `signal` option to `unpipeSignal` instead.

ehmicky:
Fixes #851.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -831,7 +831,7 @@ type PipeOptions = {
 
 	The `.pipe()` method will be rejected with a cancellation error.
 	*/
-	readonly signal?: AbortSignal;
+	readonly unpipeSignal?: AbortSignal;
 };
 
 type PipableProcess = {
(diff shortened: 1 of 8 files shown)


sindresorhus:
Conflict

ehmicky:
Fixed. :+1:

PR #854, opened 2024-02-26
Fix errors thrown by `.pipe()`

### LINKED ISSUE #853: Fix errors thrown by `.pipe()`
When a series of `.pipe()` is used:

```js
await execa('example')
  .pipe(execa('example'))
  .pipe(execa('example'), {from: 'all'})
```

The error message is not always correct. In the example above, the error should indicate that `from: all` requires the previous process to have the `all: true` option set.

To be able to do this, the last `.pipe()` needs to know what the second `execa()` options are. Since those options are not available on the `childProcess` instance passed to `.pipe()`, it cannot know it. This could be solved by using a global `WeakMap` storing child process options.

ehmicky:
Fixes #853.

### DIFF
--- a/lib/pipe/throw.js
+++ b/lib/pipe/throw.js
@@ -1,0 +1,54 @@
+import {makeEarlyError} from '../return/error.js';
+import {abortSourceStream, endDestinationStream} from './streaming.js';
+
+export const handlePipeArgumentsError = ({
+	sourcePromise,
+	sourceStream,
+	sourceError,
+	destination,
+	destinationStream,
+	destinationError,
+	stdioStreamsGroups,
+	sourceOptions,
+}) => {
+	const error = getPipeArgumentsError({sourceStream, sourceError, destinationStream, destinationError});
+	if (error === undefined) {
+		return;
+	}
+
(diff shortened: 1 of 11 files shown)


lib/pipe/throw.js
```diff
@@ -0,0 +1,54 @@
...
+		return sourceError;
+	}
+};
+
+// `.pipe()` awaits the child process promises.
+// When invalid arguments are passed to `.pipe()`, we throw an error, which prevents awaiting them.
+// We need to ensure this does not create unhandled rejections.
+const preventUnhandledRejection = (sourcePromise, destination) => {
+	Promise.allSettled([sourcePromise, destination]);
+};
```
ehmicky, line 44: This file was split from `lib/pipe/validate.js` below. 
The addition to the `preventUnhandledRejection()` function is the only change.

test/pipe/template.js
```diff
@@ -0,0 +1,106 @@
+import {spawn} from 'node:child_process';
```
ehmicky, line 1: This whole file was split from `test/script.js` but left unchanged.

PR #861, opened 2024-02-27
Refactor file descriptor-related logic

ehmicky:
This PR refactors some logic related to file descriptor. It does not change the behavior.

### DIFF
--- a/lib/stream/resolve.js
+++ b/lib/stream/resolve.js
@@ -66,12 +66,12 @@ const waitForOriginalStreams = (originalStreams, spawned, streamInfo) =>
 // Some `stdin`/`stdout`/`stderr` options create a stream, e.g. when passing a file path.
 // The `.pipe()` method automatically ends that stream when `childProcess` ends.
 // This makes sure we wait for the completion of those streams, in order to catch any error.
-const waitForCustomStreamsEnd = (stdioStreamsGroups, streamInfo) => stdioStreamsGroups.flatMap((stdioStreams, fdNumber) => stdioStreams
+const waitForCustomStreamsEnd = (stdioStreamsGroups, streamInfo) => stdioStreamsGroups.flat()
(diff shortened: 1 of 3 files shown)


lib/stream/resolve.js
```diff
@@ -66,12 +66,12 @@ const waitForOriginalStreams = (originalStreams, spawned, streamInfo) =>
...
 // The `.pipe()` method automatically ends that stream when `childProcess` ends.
 // This makes sure we wait for the completion of those streams, in order to catch any error.
-const waitForCustomStreamsEnd = (stdioStreamsGroups, streamInfo) => stdioStreamsGroups.flatMap((stdioStreams, index) => stdioStreams
+const waitForCustomStreamsEnd = (stdioStreamsGroups, streamInfo) => stdioStreamsGroups.flat()
 	.filter(({value}) => isStream(value, {checkOpen: false}) && !isStandardStream(value))
-	.map(({type, value}) => waitForStream(value, index, streamInfo, {
+	.map(({type, value, index}) => waitForStream(value, index, streamInfo, {
 		isSameDirection: type === 'generator',
 		stopOnExit: type === 'native',
-	})));
```
ehmicky, line 74: `stdioStreamsGroups` is an array representing the `stdio` option (or the `stdin`/`stdout`/`stderr` options if those are used instead). For example, `stdioStreamsGroups[1]` are the values passed to the `stdout` option. 

Since an array of values can be passed to the `std*` options, `stdioStreamsGroups` is an array of arrays. Each `stdioStreamsGroups[*][*]` item is a `stdioStream`, which is an object representing a single value passed to the `std*` option. That object has an `.index` property representing the file descriptor number.

That `.index` property should used to get the file descriptor number. Right now, it always equals the first index in `stdioStreamsGroups` array of arrays, but it might not in the future.

In other words, this refactors using `index` from `stdioStreamsGroups[*][*].index` instead of using `stdioStreamsGroups[index][*]`. The behavior is the same, this is refactoring-only.

PR #870, opened 2024-02-29
Improve documentation of global script options

ehmicky:
This improves the documentation related to sharing options with `$`.

Unfortunately, users cannot just do:

```js
import {$} from 'execa';
$ = $(options);
```

Because imports are `const`.

Doing the following:

```js
import {$} from 'execa';
const $$ = $(options);
```

Requires renaming all `$` to `$$` in the file.

The following does not require any renaming:

```js
import {$ as $$} from 'execa';
const $ = $$(options);
```

I've tried thinking of a better pattern, but could not come up with one. :/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -110,17 +110,17 @@ await $({stdio: 'inherit'})`echo unicorns`;
 //=> 'unicorns'
 ```
 
-#### Shared options
+#### Global/shared options
 
 ```js
-import {$} from 'execa';
+import {$ as $_} from 'execa';
 
-const $$ = $({stdio: 'inherit'});
+const $ = $_({stdio: 'inherit'});
 
-await $$`echo unicorns`;
+await $`echo unicorns`;
 //=> 'unicorns'
 
-await $$`echo rainbows`;
+await $`echo rainbows`;
 //=> 'rainbows'
 ```
 
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -361,8 +361,11 @@ $.verbose = true;
 
 ```js
 // Execa
-const $$ = $({verbose: true});
-await $$`echo example`;
+import {$ as $_} from 'execa';
+
+const $ = $_({verbose: true});
+
+await $`echo example`;
 ```
 
 Or:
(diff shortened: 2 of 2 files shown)


sindresorhus:
It may be less confusing to do this, since we already use `$$` for a different purpose.


```js
import {$ as $_} from 'execa';
const $ = $_(options);
```

ehmicky:
`$_` sounds good to me since it makes it clearer that is no different than `$`.

For the other places where we are using `$$`, are those purposes different? :thinking: We document creating `$$` as a new `$` with bound options. We could draw a distinction between whether this new `$` will be used in the whole script ("global" options) or only a few commands ("shared" options), but in the end, it is quite similar and uses the same method.

tommy-mitchell:
I like the distinction between a global `$` and a local `$$`. I use them separately here: https://github.com/tommy-mitchell/listr-cli/blob/main/src/tasks.ts

ehmicky:
Thanks for your input @tommy-mitchell and @aaronccasanova, let's go with what you suggest:
  - global options: `import {$ as $_}` + `const $ = $_(options)`
  - shared options: `const $$ = $(options)`

In the end, those are just variable names, and users could use different ones, but it's nice to have a convention.

I have updated the PR accordingly.

readme.md
```diff
@@ -110,17 +110,17 @@ await $({stdio: 'inherit'})`echo unicorns`;
...
+const $ = $_({stdio: 'inherit'});
 
-await $$`echo unicorns`;
+await $`echo unicorns`;
 //=> 'unicorns'
 
-await $$`echo rainbows`;
+await $`echo rainbows`;
 //=> 'rainbows'
 ```
```
ehmicky, line 125: This file also includes [an example](https://github.com/sindresorhus/execa/?tab=readme-ov-file#options) of `const $$ = $(options)` for shared options. It shows the following text, which did not need to be changed:

```
This can be used to either:
  - Set options for a specific command: `$(options)`command` `
  - Share options for multiple commands: `const $$ = $(options); $$`command`; $$`otherCommand`;`
```

PR #873, opened 2024-03-01
Improve process arguments validation

### LINKED ISSUE #872: Improve validation of process arguments
We currently validate that the file passed as argument is a string, the arguments is an array and the options is a plain object.

We should also validate that the arguments are not objects. Allowing non-string primitive types might be ok since that's pretty common to pass those as arguments (especially numbers), even though they are converted to strings under-the-hood. But passing objects is almost always not intentional.

ehmicky:
Fixes #872.


ehmicky:
Fixed conflict.

PR #880, opened 2024-03-03
Rename `signal` option to `cancelSignal`

### LINKED ISSUE #856: Avoid confusion with signals
We currently expose the following options and error properties:
  - `signal` option
  - `unpipeSignal` option
  - `killSignal` option
  - `error.signal`
  - `error.signalDescription`

The first two are `AbortSignal` instances. The last three are OS signals.
This might be confusing to some users. Especially the `signal` option and `error.signal` which use the exact same term even though they are quite different.

What are your thoughts on renaming some of those to avoid any confusion?
For example, we could do the following two things:
  - renaming `signal` option to `cancelSignal` (like `error.isCanceled`)
  - renaming `error.signal` to `error.signalName`

ehmicky:
Fixes #856.


sindresorhus:
Lets throw an error if a `signal` option is passed. Users are notoriously bad at reading release notes.

PR #884, opened 2024-03-03
Print output with `verbose` option

### LINKED ISSUE #751: Improved `verbose` option
## Proposal

When `verbose: true` is used, we should also print the command's `stdout`/`stderr`.

## Details

The `verbose` option prints commands before executing them. The main use case is debugging.

Most of the times, when users want to debug a command, they want to see the output too. 

We can do this in a way that does not change the command's behavior thanks to the recent additions to the `std*` options.

## Alternatives

Without this, users would just `console.log(result.stdout, result.stderr)` instead. However, this has some shortcomings:
  - Requires using `result.all` instead to get interleaved output
  - Does not work for commands that never complete, which are a prime target for debugging. In other words, this prints output all at once at the end of command, instead of progressively.
  - When 2 commands run in parallel, in sequence with each other, this does not show their interleaved output

Alternatively, users might use `stdout: 'inherit'`. However, this has the following problems:
  - If the command's output does not end with a newline, the result will be a little messed up (with concatenated lines that should be separate)
  - When multiple commands run in parallel, it is hard to know which is which. Also, lines might not be printed in full, leading to half of a command's output line being mixed with another command's output line.
  - Does not work with `stdout`/`stderr` returning objects, due to transforms in `objectMode`
  - Users doing `stdio: 'inherit'` instead would change the command's behavior by making `stdin` interactive

In other words, getting this right is actually trickier than it looks.

In production, developers tend to use complex commands with weird setups, which can take hours to debug. When it comes to making command execution easier, improving debugging is a major way we can help out our users.

ehmicky:
Fixes #751.

### DIFF
--- a/lib/verbose/info.js
+++ b/lib/verbose/info.js
@@ -1,0 +1,19 @@
+import {debuglog} from 'node:util';
+
+export const getVerboseInfo = ({verbose = verboseDefault}) => {
+	if (verbose === 'none') {
+		return {verbose};
+	}
+
+	const verboseId = VERBOSE_ID++;
+	return {verbose, verboseId};
+};
+
+const verboseDefault = debuglog('execa').enabled ? 'full' : 'none';
+
+// Prepending the `pid` is useful when multiple commands print their output at the same time.
+// However, we cannot use the real PID since this is not available with `child_process.spawnSync()`.
+// Also, we cannot use the real PID if we want to print it before `child_process.spawn()` is run.
+// As a pro, it is shorter than a normal PID and never re-uses the same id.
+// As a con, it cannot be used to send signals.
+let VERBOSE_ID = 0n;
--- a/lib/verbose/output.js
+++ b/lib/verbose/output.js
@@ -1,0 +1,68 @@
+import {inspect} from 'node:util';
+import stripFinalNewline from 'strip-final-newline';
+import {escapeLines} from '../arguments/escape.js';
+import {PIPED_STDIO_VALUES} from '../stdio/forward.js';
+import {verboseLog} from './log.js';
+
+export const handleStreamsVerbose = ({stdioStreams, options, isSync, stdioState, verboseInfo}) => {
+	if (!shouldLogOutput(stdioStreams, options, isSync, verboseInfo)) {
+		return stdioStreams;
+	}
+
+	const [{fdNumber}] = stdioStreams;
+	return [...stdioStreams, {
+		...stdioStreams[0],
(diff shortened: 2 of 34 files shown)


lib/verbose/info.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+const verboseDefault = debuglog('execa').enabled ? 'full' : 'none';
+
+// Prepending the `pid` is useful when multiple commands print their output at the same time.
+// However, we cannot use the real PID since this is not available with `child_process.spawnSync()`.
+// Also, we cannot use the real PID if we want to print it before `child_process.spawn()` is run.
+// As a pro, it is shorter than a normal PID and never re-uses the same id.
+// As a con, it cannot be used to send signals.
+let VERBOSE_ID = 0n;
```
ehmicky, line 19: This is the number printed next to the timestamp.

lib/verbose/output.js
```diff
@@ -0,0 +1,68 @@
...
+
+// Only print text output, not binary
+// eslint-disable-next-line unicorn/text-encoding-identifier-case
+const ALLOWED_ENCODINGS = new Set(['utf8', 'utf-8']);
+
+// Printing input streams would be confusing.
+// Files and streams can produce big outputs, which we don't want to print.
+// We could print `stdio[3+]` but it often is redirected to files and streams, with the same issue.
+// So we only print stdout and stderr.
+const fdUsesVerbose = ([{fdNumber}]) => fdNumber === 1 || fdNumber === 2;
```
ehmicky, line 40: The comments in this file describe the reasons behind some of the restrictions for the `verbose: 'full'` option described in `readme.md`.

lib/verbose/output.js
```diff
@@ -0,0 +1,68 @@
...
+	}];
+};
+
+// `ignore` opts-out of `verbose` for a specific stream.
+// `ipc` cannot use piping.
+// `inherit` would result in double printing.
+// They can also lead to double printing when passing file descriptor integers or `process.std*`.
+// This only leaves with `pipe` and `overlapped`.
+const shouldLogOutput = (stdioStreams, {encoding}, isSync, {verbose}) => verbose === 'full'
+	&& !isSync
```
ehmicky, line 26: `execaSync()` cannot currently use this because it cannot use transforms and this feature uses transforms.

However, I am planning on adding support for both to `execaSync()`.

ehmicky:
Fixed merge conflict.

PR #897, opened 2024-03-08
Rename `spawned` and `child process` to `subprocess`

### LINKED ISSUE #877: Use consistent variable name for child processes
Our source code and documentation examples use different variable names for child processes: `spawned`, `subprocess` and `childProcess`. We should be consistent and use a single name. 

I would personally choose `childProcess` to reflect the `child_process` core module, but any other works.

ehmicky:
Fixes #877.

Using consistent wording actually makes both the code and the documentation simpler to grasp. Lots of replacements in this PR!


test/stream/subprocess.js
```diff
@@ -0,0 +1,118 @@
+import {platform} from 'node:process';
```
ehmicky, line 1: File was just renamed, but `git` seems not to detect it.

PR #903, opened 2024-03-12
Allow piping process to any file descriptor

ehmicky:
Users can currently pipe any file descriptor from a process to another process. It defaults to `stdout` and uses the `from` option.

```js
await execa(...).pipe(..., {from: 'stdout'})
await execa(...).pipe(..., {from: 'stderr'})
await execa(...).pipe(..., {from: 'all'})
await execa(...).pipe(..., {from: 3})
```

This PR adds the same feature, but for the file descriptor of the destination process instead. It defaults to `stdin` and uses the `to` option.

```js
await execa(...).pipe(..., {to: 'stdin'})
await execa(...).pipe(..., {to: 3})
```


test/exit/kill.js
```diff
@@ -179,7 +179,7 @@ test('Cannot call .kill(null)', testInvalidKillArgument, null);
 test('Cannot call .kill(0n)', testInvalidKillArgument, 0n);
 test('Cannot call .kill(true)', testInvalidKillArgument, true);
 test('Cannot call .kill(errorObject)', testInvalidKillArgument, {name: '', message: '', stack: ''});
-test('Cannot call .kill([error])', testInvalidKillArgument, [new Error('test')]);
```
ehmicky, line 182: GitHub action shows a weird `[error]` highlighted in red when a CI task prints the word `[error]` apparently. :)

ehmicky:
Conflict solved.

PR #907, opened 2024-03-13
Fix Markdown typos in documentation

ehmicky:
This fixes some small documentation typos.


docs/transform.md
```diff
@@ -49,8 +49,8 @@ await execa('./binary.js', {stdout: {transform, binary: true}});
 ```
 
 This is more efficient and recommended if the data is either:
-	- Binary: Which does not have lines.
-	- Text: But the transform works even if a line or word is split across multiple chunks.
```
ehmicky, line 53: Bullet points are not showing up when using tabs.

PR #911, opened 2024-03-17
Add `ExecaError` and `ExecaSyncError` classes

### LINKED ISSUE #909: Improve error typings
Hi,

Currently I have to do:
```ts
        try {
          const cmdResult = await $({
            timeout: 2000,
          })`sleep 10`;
        } catch (error) {
          if (error instanceof Error && !!(error as any).shortMessage) {
            console.log((error as any).shortMessage);
          } else {
            throw error;
          }
        }
```

I think it would be easier if people could do something like `error instanceof ExecaCommonError` so they can handle the case properly.

It would require some change so errors are instanciated but it should not break any linter after this switch.

In the meantime I saw currently it's a type (https://github.com/sindresorhus/execa/blob/3ecd447a8d93e3eae69836f960fe9ed7b0d16bb6/index.d.ts#L772) and I wanted to use it to cast the type of my error, but this one is not exported.

Thank you,

ehmicky:
Fixes #909.

This makes Execa errors instances of the `ExecaError` and `ExecaSyncError` classes.

The reason behind having a separate error class for `ExecaSyncError` is because some of their properties are different, e.g. `error.pipedFrom` and `error.all` do not exist with sync calls. This also allows type guards with TypeScript when using `instanceof`.

```ts
import {execa, ExecaSyncError} from 'execa';

try {
  execaSync(...);
} catch (error) {
  if (error instanceof ExecaSyncError) {
    console.log(error.pipedFrom) // This fails because this property is not available with sync calls
  }
}
``` 

This PR also changes the strategy for exposing inner errors. When a stream or a transform errors, it is useful to keep the inner error, i.e. the original error instance. Previously, we were cloning that inner error, to prevent multiple `execa()` from sharing the same error instance. We would then copy known error properties from the inner error to the top-level error. That logic was brittle since cloning errors [is hard](https://github.com/ehmicky/error-serializer).

Please note that most subprocess faillures do not have an inner error. This includes: non-0 exit code and signal termination.

This PR relies instead on [`error.cause`](https://emulator.toolsinbox.com/?_=%2Fen-US%2Fdocs%2FWeb%2FJavaScript%2FReference%2FGlobal_Objects%2FError%2Fcause%23qEloBH%2BIKdhRub4dTIFq0wzXG3T8ZIztRTpEFNo%3D). This removes any need to copy error properties or clone error instances. This is much cleaner and avoids having to directly mutate the underlying error instances. 

For example, when using `subprocess.kill(userError)`, Execa will throw an instance of `ExecaError` and `error.cause` will be `userError`.

If `error.cause.code` exists, it is copied to `error.code`. This is because checking that property is very common, so this reduces the potential for breaking changes. It is also convenient.


index.d.ts
```diff
@@ -788,10 +767,207 @@ type ExecaCommonError = {
...
+
+	/**
+	Node.js-specific [error code](https://nodejs.org/api/errors.html#errorcode), when available.
+	*/
+	code?: string;
 } & Error;
 
-export type ExecaError<OptionsType extends Options = Options> = ExecaCommonResult<false, OptionsType> & ExecaCommonError;
-export type ExecaSyncError<OptionsType extends SyncOptions = SyncOptions> = ExecaCommonResult<true, OptionsType> & ExecaCommonError;
+declare class CommonErrorClass<IsSync extends boolean, OptionsType extends Options = Options> extends Error {
```
ehmicky, line 800: This copies all fields and their descriptions from `CommonError` and `CommonResult`.

This is because of TypeScript shortcomings:
  - This must be a `class` so that `instanceof` and `InstanceType` work
  - Mixin and multiple inheritance are not possible in an ambient file
  - Classes cannot use `&` or merge types from another type, except through `extends` or composition, which cannot be used here.

Avoiding this duplication would be great. If you have any idea on how to prevent this, please let me know.

readme.md
```diff
@@ -567,6 +547,58 @@ The numeric exit code of the subprocess that was run.
...
+
+This is `undefined` unless the subprocess exited due to an `error` event or a timeout.
+
+#### cause
+
+Type: `unknown | undefined`
+
+Underlying error, if there is one. For example, this is set by [`.kill(error)`](#killerror).
+
+This is usually an `Error` instance.
```
ehmicky, line 600: We could normalize `error.cause` to an `Error` instance. But then, users might miss some information. For example, if it is an object but not an `Error` instance, it would require some more complex logic to serialize its properties.

I have [a library](https://github.com/ehmicky/normalize-exception) that does this, and getting this correctly is actually tricky.

But since it is not guaranteed to be an `Error` instance, we document it so users do `error?.message` instead of something like `error.message`.

index.js
```diff
@@ -3,3 +3,4 @@ export {execaSync} from './lib/sync.js';
 export {execaCommand, execaCommandSync} from './lib/command.js';
 export {execaNode} from './lib/arguments/node.js';
 export {$} from './lib/script.js';
+export {ExecaError, ExecaSyncError} from './lib/return/cause.js';
```
ehmicky, line 6: We could also call those `SubprocessError` and `SubprocessSyncError`. Not sure which one you prefer? Or any other name?

readme.md
```diff
@@ -587,12 +567,62 @@ If a signal terminated the subprocess, this property is defined and included in
 
 #### pipedFrom
 
-Type: [`SubprocessResult[]`](#subprocessresult)
+Type: [`Array<SubprocessResult | ExecaError>`](#subprocessresult)
```
ehmicky, line 570: `SubprocessResult` is a subclass of `ExecaError`, but it's useful to distinguish those to make it clear `pipedFrom` might contain both successful results and error instances.

readme.md
sindresorhus: I think it's fine to not guarantee it being an error. Even though it's suboptimal, it's how it works everywhere in JS.

index.js
sindresorhus: I think the current names are the best. An error could potentially not be related to subprocesses too, like validation errors.

index.d.ts
sindresorhus: I've hit this issue in the past too and not sure there is a better solution. It would be useful with a short code comment that points out why we are doing it like this, because if you are not aware of the limitation, the code looks unnecessarily boilerplaty. Maybe just add a URL to the above comment.

index.d.ts
```diff
@@ -788,10 +783,191 @@ type ExecaCommonError = {
 	This is `undefined` unless the subprocess exited due to an `error` event or a timeout.
```
sindresorhus, line 783: I would invert the formulation:

```suggestion
	This exists only if the subprocess exited due to an `error` event or a timeout.
```

Also because technically the property is not `undefined`, but rather non-existent. TypeScript can differentiate between those for users with the `--exactOptionalPropertyTypes` flag.

index.d.ts
```diff
@@ -788,10 +783,191 @@ type ExecaCommonError = {
...
+- there's not enough memory or there are already too many subprocesses
+
+This has the same shape as successful results, with a few additional properties.
+*/
+export class ExecaError<OptionsType extends Options = Options> extends CommonErrorClass<false, OptionsType> {
+	readonly name: 'ExecaError';
+}
+
+/**
+Exception thrown when the subprocess [fails](#failed), either:
```
sindresorhus, line 959: The link needs to be absolute.

sindresorhus (changes requested):

index.d.ts
ehmicky: I've found a solution. Having a common class for both success and failed results, from which they both inherit.
Common result's type type error properties as `string | undefined`
Success result's type narrows error properties to `never`.
Failed result's type narrows  error properties to `string` (no `undefined`).

I've updated the PR with this.
No more duplication. :tada:

index.js
ehmicky: Validation errors actually do not have `ExecaError` type, they are just `Error` instances. 

More precisely, we have 3 types of errors based on when they arise:
  1. Arguments validation, for example if the user does `execa(false)` or `execa(..., {cwd: false})`. Those throw synchronously an `Error` instance. 
  2. Exceptions by `child_process.spawn()`, i.e. the subprocess could not be spawned. Those are called "early errors" in the codebase. They reject asynchronously an `ExecaError` instance.
  3. Subprocess failures. They also reject asynchronously an `ExecaError` instance.

Ideally, the type `1.` would also reject an `ExecaError` instance, for consistency. I've tried to do implement this a few weeks ago and it was very difficult and complicated the codebase quite a lot, so I gave up on it.

But at the moment, `ExecaError` is always subprocess-related. That being said, that name still works for me. :)

index.d.ts
ehmicky: Fixed, keeping it consistent with the rest of the file, which only includes external links, as opposed to links inside the `readme.md`.

ehmicky:
Fixed. :+1:

Issue #591, opened 2023-12-14
Add `childProcess.readable` and `childProcess.writable`


ehmicky:
Background: #143

## New feature

We could add the following properties:
  - `childProcess.readable`: like `childProcess.stdin`, but as a `ReadableStream`. This is `undefined` if `childProcess.stdin` is `undefined`, i.e. if the `stdin` option is not `pipe` (its default value).
  - `childProcess.writable`: same but for `childProcess.stdout`, as a `WritableStream`. If `childProcess.stdout` is `undefined` but `childProcess.stderr` is not, the latter is used instead. If the `all` option is `true`, `childProcess.all` is used instead. I.e. users can choose `stdout`, `stderr` or both. Any error thrown by the child process is propagated to `childProcess.writable`.
  
This can be easily implemented using [`Stream.Readable.toWeb()`](https://nodejs.org/api/stream.html#streamreadabletowebstreamreadable-options) and [`Stream.Writable.toWeb()`](https://nodejs.org/api/stream.html#streamwritabletowebstreamwritable).

## TransformStream

Having a `readable` and `writable` properties would allow `childProcess` to be used as a `TransformStream` (even if it's not an instance of it) in `Stream.pipeline()`, `Stream.compose()` and `ReadableStream.pipeThrough()`.

This is also useful for users who want to manipulate `stdin`/`stdout`/`stderr` directly but prefer web streams.

## Example

```js
import { pipeline } from 'stream/promises'

await pipeline(
  execa(...),
  execa(...),
  myTransformingStream,
  createReadStream('filePath'),
)
```

This also works well with `$`.

```js
await pipeline(
  $`ls mydir`,
  $`sort`,
  $`head -n 20`,
)
```

sindresorhus:
👍

ehmicky:
I have pushed a `readable-writable` branch which implements this.
However, this seems rather unstable and needs more work. Notably, the tests are randomly failing.

The main issue is about how to propagate errors between the web streams (`readable`/`writable`), the Node.js streams (`process.stdin|stdout|stderr`) and the child process. It is not very easy because the web streams API does not allow consumers to error a stream, only producers.

So, in that branch, `childProcess` can be used as a `TransformStream`. However, when a child process errors, the stream does not emit an `error` event. 

This is quite a hard issue to solve. :thinking:

ehmicky:
[...]

PR #912, opened 2024-03-17
Add `.readable()`, `.writable()` and `.duplex()` methods

### LINKED ISSUE #143: Consider implementing a duplex mode

ehmicky:
Fixes #143.

The [following comment](https://github.com/sindresorhus/execa/issues/143#issuecomment-2002306393) explains the background behind the following points:
  - three different methods instead of only `.duplex()`
  - `.readable()` not named `.readableStream()`
  - Node.js streams instead of web streams
  - not using `.readable` and `.writable` properties as suggested in #591

This was some pretty hard work there!
I had to re-implement the whole PR from scratch 3 or 4 times. Streams are difficult! :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 import {type ChildProcess} from 'node:child_process';
-import {type Readable, type Writable} from 'node:stream';
+import {type Readable, type Writable, type Duplex} from 'node:stream';
 
 type IfAsync<IsSync extends boolean, AsyncValue, SyncValue = never> = IsSync extends true ? SyncValue : AsyncValue;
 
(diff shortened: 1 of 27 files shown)


sindresorhus:
I have not reviewed this in depth as it's quite dense, but I trust the extensive amount of tests.

sindresorhus:
Nice work 🎉

PR #915, opened 2024-03-18
Use `Uint8Array` instead of `Buffer`

ehmicky:
This PR forbids returning `Buffer` instead of `Uint8Array` in transforms.

Furthermore, when a transform in `objectMode` is used, `subprocess.stdout` can be any type. Therefore, it can still be `Buffer` then (since any type is allowed). However, we do not handle `Buffer` differently from other types then. In particular, those:
  - are not serialized in the `error.message`
  - cannot be passed to `${...}` when using the template string syntax

Instead, users should use `Uint8Array`.


lib/stdio/validate.js
```diff
@@ -12,7 +12,7 @@ const validateObjectTransformReturn = function * (optionName, chunk) {
 const validateStringTransformReturn = function * (optionName, chunk) {
 	validateEmptyReturn(optionName, chunk);
 
-	if (typeof chunk !== 'string' && !isBinary(chunk)) {
+	if (typeof chunk !== 'string' && !isUint8Array(chunk)) {
 		throw new TypeError(`The \`${optionName}\` option's function must yield a string or an Uint8Array, not ${typeof chunk}.`);
```
sindresorhus, line 16: `typeof chunk` will say `object` when Buffer is passed in. It may be a good idea to be a little bit more user-friendly and specifically detect Buffer and say that in the message.

sindresorhus (changes requested):

lib/stdio/validate.js
ehmicky: Fixed. :+1:

PR #919, opened 2024-03-21
Strip newlines in transforms

ehmicky:
This PR improves how newlines are handled in transforms:
  - The newline is now automatically stripped from each `line` argument passed to transforms
  - `yield string` statements do not need to add a newline at the end of `string` anymore. Each `yield` produces a new line (kindof like `console.log()`).

The above behavior can be disabled by passing a `{transform, newline: true}` plain object.

As an added bonus, the `stripFinalNewline` option now works with the `lines` option. If `lines: true` is used, and `stripFinalNewline: true` is used too, it will be applied to each line, i.e. each item of the `result.stdout`/`result.stderr` array.

### DIFF
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -54,6 +54,66 @@ This is more efficient and recommended if the data is either:
 
 Please note the [`lines`](../readme.md#lines) option is unrelated: it has no impact on transforms.
 
+## Newlines
+
+Unless [`{transform, binary: true}`](#binary-data) is used, the transform iterates over lines.
+By default, newlines are stripped from each `line` argument.
+
+```js
+// `line`'s value never ends with '\n'.
+const transform = function * (line) { /* ... */ };
+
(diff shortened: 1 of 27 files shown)


docs/transform.md
```diff
@@ -54,6 +54,66 @@ This is more efficient and recommended if the data is either:
...
+By default, newlines are stripped from each `line` argument.
+
+```js
+// `line`'s value never ends with '\n'.
+const transform = function * (line) { /* ... */ };
+
+await execa('./run.js', {stdout: transform});
+```
+
+However, if a `{transform, newline: true}` plain object is passed, newlines are kept.
```
sindresorhus, line 69: `{newline: true}` reads to me like it enables splitting on newlines. Maybe `preserveNewlines`?

ehmicky: Yes, that's much clearer. Fixed in 04345af7c848b7a319e075fa81836e051316a20b.

Issue #925, opened 2024-03-23
`encoding` option accepted values


ehmicky:
The `encoding` option accepts the same values as Node.js `child_process`:
  - `utf8`
  - `buffer`
  - `utf16le`
  - `hex`
  - `base64`
  - `base64url`
  - `latin1`
  - `ascii`

It also allows the following:
  1. `null` as an alias to `buffer`
  2. `utf-8` as an alias to `utf8` 
  3. `utf-16le`, `ucs2` and `ucs-2` as aliases to `utf16le`
  4. `binary` as an alias to `latin1`
  5. Any case, e.g. `UTF8` instead of `utf8`

@sindresorhus Do you think we should support the same options as core Node.js, or we should be stricter? If stricter, which of the above should be kept or not?

sindresorhus:
Stricter. I would drop 1, 2, 3, 4, 5. Needs a human-friendly error when any of those are passed.

ehmicky:
Alright, let's do this.
I'll make sure the validation error message includes the value to use instead.

PR #926, opened 2024-03-24
Improve handling of `encoding` option

ehmicky:
Some of our logic changes based on whether the subprocess output is binary or not: the `lines` option, transforms, `subprocess.readable()`, `subprocess.duplex()` and `subprocess.iterable()`.

We currently detect this by checking if the `encoding` option is `buffer`. However, if it is `hex`, it almost certainly also means the output is binary. The same applies to `base64`, `base64url`, `latin1` and `ascii` (those last 2 are improperly named, they are actually ways to represent binary data in text form).

This PR fixes this.

It also improves the documentation and testing of the `encoding` option.


index.d.ts
```diff
@@ -85,23 +85,27 @@ type StdioOptionsArray<IsSync extends boolean = boolean> = readonly [
...
+type BinaryEncodingOption =
+  | BufferEncodingOption
   | 'hex'
   | 'base64'
   | 'base64url'
-  | BufferEncodingOption
-  | undefined;
+  | 'latin1'
+  | 'binary'
+  | 'ascii';
```
ehmicky, line 104: This PR still handles the encoding aliases, `null` and case-insensitivity. 
Per #925, I'll remove those in a follow-up PR.

PR #928, opened 2024-03-25
Stricter values for the `encoding` option

### LINKED ISSUE #925: `encoding` option accepted values

ehmicky:
Fixes #925.


sindresorhus:
Shouldn't we remove the aliases here: https://github.com/sindresorhus/execa/blob/c9cdd2d27a1cb4a11bcfee2586defea0104bed12/index.d.ts#L91 ?

ehmicky:
Yes, we should, good catch! 
Fixed in 7fce4d1deabc507d37265e8fe0a28bf18d7fb2a0

PR #929, opened 2024-03-25
Simplify `lines` option

ehmicky:
The `lines` option currently does two different things:
1. Split `result.stdout`/`result.stderr` into arrays of lines
2. Make `subprocess.stdout`/`subprocess.stderr` streams iterate over chunks of lines instead of arbitrary chunks of data

If users only want to do `1.`, then getting `2.` as a side effect might be unwanted, since it sets `subprocess.stdout`/`subprocess.stderr` in object mode, and makes it iterate over smaller chunks of data.

Also, `2.` can now be achieved using either `subprocess.iterable()`, transforms or `subprocess.readable({binary: false})`. So it is now redundant with those methods. In fact, those methods are better because they create a new stream/iterable instead of directly modifying `subprocess.stdout`/`subprocess.stderr`.

Therefore, this PR makes the `lines` option do only `1.`. This also makes this option much simpler to understand, as the shortening of the `readme.md` description shows. Line-wise iteration over `subprocess.stdout`/`subprocess.stderr` can still be achieved, but using `subprocess.iterable()`, transforms or `subprocess.readable({binary: false})` instead.

### DIFF
--- a/docs/transform.md
+++ b/docs/transform.md
@@ -53,8 +53,6 @@ This is more efficient and recommended if the data is either:
 - Binary: Which does not have lines.
 - Text: But the transform works even if a line or word is split across multiple chunks.
 
-Please note the [`lines`](../readme.md#lines) option is unrelated: it has no impact on transforms.
-
 ## Newlines
 
 Unless [`{transform, binary: true}`](#binary-data) is used, the transform iterates over lines.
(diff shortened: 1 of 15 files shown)


sindresorhus:
Love how you keep iterating and improving these APIs 👍

ehmicky:
Thanks @sindresorhus!

PR #939, opened 2024-03-29
Fix types of `execaSync()`

ehmicky:
This PR improves the types of `execaSync()`, making it clearer that some features are not available with synchronous methods.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,11 @@
 import {type ChildProcess} from 'node:child_process';
 import {type Readable, type Writable, type Duplex} from 'node:stream';
 
-type IfAsync<IsSync extends boolean, AsyncValue, SyncValue = never> = IsSync extends true ? SyncValue : AsyncValue;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -13,11 +13,11 @@ type NoStreamStdioOption =
...
 	| [NoStreamStdioOption];
 
-type BaseStdioOption =
+type BaseStdioOption<IsSync extends boolean = boolean> =
 	| 'pipe'
-	| 'overlapped'
 	| 'ignore'
-	| 'inherit';
+	| 'inherit'
+	| IfAsync<IsSync, 'overlapped'>;
```
ehmicky, line 20: Note: `overlapped` currently crashes the process with `spawnSync()`, instead of just throwing an error, or just behaving like `pipe`.

I opened a bug in https://github.com/nodejs/node/issues/52265

ehmicky:
Fixed merge conflict.

PR #941, opened 2024-03-30
Improve validation of `execaSync()` `stdio` option

ehmicky:
An array of values can be passed to the `stdin`/`stdout`/`stderr` options.
This works well with `execa()`, and uses `subprocess.std*.pipe()` under-the-hood.
However, with `execaSync()`, this is harder to achieve since all we can use is the `input` option of `child_process.spawnSync()`.

This PR adds some validation logic to let users know of when they are hitting the above restrictions, i.e. passing an invalid array of values to the `stdin`/`stdout`/`stderr` options with `execaSync()`.

This PR also improves the related tests and documentation.


ehmicky:
Fixed merge conflict.

PR #942, opened 2024-03-31
Improve type tests

ehmicky:
This PR improves the tests for the TypeScript types.


index.d.ts
```diff
@@ -91,9 +91,11 @@ type StdinSingleOption<
 	| CommonStdioOption<IsSync, IsArray>
 	| InputStdioOption<IsSync, IsArray>;
 
-export type StdinOption<IsSync extends boolean = boolean> =
+type StdinOptionCommon<IsSync extends boolean = boolean> =
 	| StdinSingleOption<IsSync, false>
 	| Array<StdinSingleOption<IsSync, true>>;
+export type StdinOption = StdinOptionCommon<false>;
+export type StdinOptionSync = StdinOptionCommon<true>;
```
ehmicky, line 98: Use the same async vs sync (instead of using a generic parameter) export convention as the other types.

PR #962, opened 2024-04-06
Improve `verbose` option

ehmicky:
This PR refactors the `verbose: 'full'` option so that it does not modify `subprocess.stdout`.
It does not change its behavior otherwise.


ehmicky:
Fixed merge conflict.

PR #963, opened 2024-04-07
Add `error.isMaxBuffer`

ehmicky:
This adds `error.isMaxBuffer: boolean`. This is `true` when the subprocess failed due to the `maxBuffer` option.

Figuring out whether the subprocess failed due to the `maxBuffer` option is not completely straightforward otherwise:
  - `execa()` fails with `MaxBufferError` instances (which Execa does not export)
  - `execaSync()` fails with `error.code` `ENOBUFS`. However that code is also used for other errors by Node.js.

Also, this PR improves the error message when `maxBuffer` is hit. Previously, it just showed `maxBuffer exceeded`. Now, the error message is clearer. Also, it indicates:
  - Which stream was too big: `stdout`, `stderr`, `stdio[3]`, etc.
  - The `maxBuffer` value
  - The unit: bytes, characters, lines or objects


ehmicky:
Fixed merge conflict.

PR #967, opened 2024-04-08
Improve headers in `readme.md`

ehmicky:
The `readme.md` headers can be confusing. I often scroll up and down not knowing in which section I am.

I think one of the issues is that we re-use the same words for some options, subprocess properties and result properties. In general, that's a good thing, but it creates confusion in the `readme.md`. For example, we are using 3 times the "all" header: for the option, the stream, and the result property.

This PR fixes this by prefixing those headers. For example, `options.all`, `subprocess.all`, `result.all` and `error.all`. This makes it clearer which section the user is currently reading.


sindresorhus:
👍

sindresorhus:
Conflict

ehmicky:
Fixed.

PR #978, opened 2024-04-18
Split type files

ehmicky:
TypeScript types files are currently very big: ~2000 lines for the types and ~4000 lines for the type tests.

This makes editing those files very slow. VSCode takes 1+ minute to save the type tests file on my machine.
Also, this makes it hard to find specific types.

This PR splits those two big files into many smaller files.
It does not change the types, it just re-organizes them into smaller files.


package.json
```diff
@@ -25,7 +25,8 @@
 	"files": [
 		"index.js",
 		"index.d.ts",
-		"lib"
+		"lib",
+		"types"
```
ehmicky, line 29: Not forgetting this.

PR #984, opened 2024-04-23
Add more code comments

ehmicky:
This PR adds more code comments, to make it easier for contributors.

### DIFF
--- a/lib/arguments/command.js
+++ b/lib/arguments/command.js
@@ -4,6 +4,7 @@ import {getStartTime} from '../return/duration.js';
 import {joinCommand} from './escape.js';
 import {normalizeFdSpecificOption} from './specific.js';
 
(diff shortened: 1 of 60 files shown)


lib/arguments/command.js
```diff
@@ -4,6 +4,7 @@ import {getStartTime} from '../return/duration.js';
 import {joinCommand} from './escape.js';
 import {normalizeFdSpecificOption} from './specific.js';
 
+// Compute `result.command`, `result.escapedCommand` and `verbose`-related information
```
sindresorhus, line 7: Maybe standardize with `.` at the end? Some comments have it and some don't.

ehmicky: Currently the comments do not use a `.` when there is a single sentence, and use a `.` otherwise.

Would you prefer to always using a `.`, even when there is a single sentence?

PR #989, opened 2024-04-27
Add user guides

ehmicky:
This PR adds [user guides](https://github.com/sindresorhus/execa/tree/guides?tab=readme-ov-file#documentation).

This groups Execa features into sections, pages and groups of pages, which helps understanding those features. In a way, it is similar to the documentation approach taken by some of your other bigger projects like [AVA](https://github.com/avajs/ava/tree/main/docs) and [got](https://github.com/sindresorhus/got?tab=readme-ov-file#documentation).

The user guides are ordered, so they can be used as a user manual, reading one user guide at a time.

The user guides include many examples and add more details and context to each feature's documentation. The API reference points to those additional details, which allows that API reference to be smaller and less overwhelming.

Some features which failed to be properly documented are also covered by this PR. For example: 
- Shebangs support on Windows was previously briefly mentioned, but not fully documented.
- `uid`/`gid` options cannot be used on Windows, but this was documented.
- Null bytes not allowed in the command arguments, and how to work around it.
- And more.


sindresorhus:
Wow! This is really well done and comprehensive. Nice work! I read through it all now and could not really find anything to comment on.

ehmicky:
Thanks @sindresorhus, this took many days of work!
I hope this will help with highlighting all the features we offer.

PR #1004, opened 2024-05-01
Fix passing the same file or stream to both `stdout` and `stderr`

### LINKED ISSUE #1002: Fix passing the same file or stream to both `stdout` and `stderr`
It is not currently possible to redirect both `stdout` and `stderr` to the same file or stream when using the `stdout` and `stderr` options.

We should allow this. Otherwise, users have to use the `all` option then pipe `subprocess.all`, which is more verbose and error-prone.

Also, we should forbid passing the same file or stream to both `stdin` and either `stdout`/`stderr`, since that does not make much sense.

ehmicky:
Fixes #1002.

This works with both `execa()` and `execaSync()`.

### DIFF
--- a/docs/bash.md
+++ b/docs/bash.md
@@ -496,6 +496,34 @@ await $({stdout: {file: 'output.txt'}})`npm run build`;
 
 [More info.](output.md#file-output)
 
+### Piping interleaved stdout and stderr to a file
+
+```sh
+# Bash
+npm run build &> output.txt
+```
+
+```js
+// zx
+import {createWriteStream} from 'node:fs';
+
+const subprocess = $`npm run build`;
+const fileStream = createWriteStream('output.txt');
+subprocess.pipe(fileStream);
+subprocess.stderr.pipe(fileStream);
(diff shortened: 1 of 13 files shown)


docs/bash.md
```diff
@@ -493,6 +493,34 @@ await $({stdout: {file: 'output.txt'}})`echo example`;
...
+subprocess.stderr.pipe(fileStream);
+await subprocess;
+```
+
+```js
+// Execa
+await $({
+	stdout: {file: 'output.txt'},
+	stderr: {file: 'output.txt'},
+})`npm run build`;
```
sindresorhus, line 519:
```suggestion
const output = {file: 'output.txt'};
await $({
	stdout: output,
	stderr: output,
})`npm run build`;
```

?

sindresorhus (changes requested):

docs/bash.md
ehmicky: Yes, that's better. :+1:

Issue #413, opened 2020-02-16
Attach `.child` to the execa promise


sindresorhus:
Node.js does this now for `childProcess.` when you promisify it.

https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

The name is to be decided.

One day, I hope we can get rid of `execa()` being both a childProcess and promise and just have the `.child` property on the promise. But that's a huge breaking change. For now, the best we can do is to add it, and recommend new users to use it instead of the old way. We could even un-document the old way.

@ehmicky Thoughts?

ehmicky:
Yes this sounds good :+1:

sindresorhus:
Any thoughts on the name?

- `.child`
- `.process`
- `.subprocess`
- `.childProcess`

ehmicky:
I think it should be `.child` since this is the name used by Node.js `util.promisify(childProcess.exec)`?

sindresorhus:
👍

ehmicky:
@wesleytodd Based on your comment [here](https://github.com/sindresorhus/execa/issues/414#issuecomment-587171732), would you like to take a look at doing a PR for this? Please let us know if you need any help!

wesleytodd:
I plan to, just have other priorities at work. If someone else gets to it before me that’s fine, my guess is I will be able to do it late next week at the earliest.

ehmicky:
Per [this comment](https://github.com/sindresorhus/execa/issues/877#issuecomment-1982195398), we are going to name the property `subprocess` instead.

ehmicky:
We probably should keep the following specific methods on the return value, as opposed to the `.subprocess` property: `.pipe()` and `[Symbol.asyncIterator]`/`.iterable()`. This leads a much more natural syntax.

```js
const result = await execa`npm run build`
  .pipe`sort`
  .pipe`head -n 2`;

for await (const line of execa`npm run build`) {
  // ...
}
```

Instead of:

```js
const result = await execa`npm run build`
  .subprocess.pipe`sort`
  .subprocess.pipe`head -n 2`;

for await (const line of execa`npm run build`.subprocess) {
  // ...
}
```

All the other properties/methods though can be on the `.subprocess` property. But it feels like those ones might be made too verbose by the addition of `.subprocess`.

ehmicky:
I've been thinking about this issue, and would suggest the following:
[...]

PR #1007, opened 2024-05-01
Fix the type of the return value

ehmicky:
This PR fixes the type of `execa()`'s return value.

The names of `ExecaSubprocess` and `ExecaResultPromise` were inverted: the subprocess was called `ExecaResultPromise` and the promise was called `ExecaSubprocess`. We probably did not notice until now because right now the same variable uses both types.

```ts
const promiseOrSubprocess: ExecaSubprocess | ExecaResultPromise = execa(...);
```

However, once we split those two (according to #413) by adding a separate `.subprocess` property, this would currently be typed like this, which is incorrect:

```ts
const promise: ExecaSubprocess = execa(...);
const subprocess: ExecaResultPromise = promise.subprocess;
```

This PR swaps the two names. The upcoming release already changes the names of those types (those were previously called `ExecaChildPromise` and `ExecaChildProcess`), so this does not add any new breaking change.

Additionally, this PR improves the documentation of Execa's return value.

Finally, it adds more type tests for `ExecaSubprocess` and `ExecaResultPromise`, and renames some variables from `promise` to `subprocess` when what is being tested is actually the subprocess.


types/subprocess/subprocess.d.ts
```diff
@@ -115,6 +115,18 @@ export type ExecaResultPromise<OptionsType extends Options = Options> = {
...
+	& ExecaCustomSubprocess<OptionsType>;
+
+/**
+The return value of all asynchronous methods is both:
+- the subprocess.
+- a `Promise` either resolving with its successful `result`, or rejecting with its `error`.
+*/
+export type ExecaResultPromise<OptionsType extends Options = Options> =
+	& ExecaSubprocess<OptionsType>
+	& Promise<ExecaResult<OptionsType>>;
```
ehmicky, line 132: This file is the main change.

PR #1012, opened 2024-05-02
Rename a few more types

ehmicky:
This PR renames the following types:
  - `StdinOptionSync` -> `StdinSyncOption`
  - `StdoutStderrOptionSync` -> `StdoutStderrSyncOption`

This makes it consistent with the names of the other synchronous types, in particular `SyncOptions`, but also `SyncResult` and `ExecaSyncError`.

`StdinOptionSync` and `StdoutStderrOptionSync` are new types added by the upcoming release, so this is not a breaking change.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,8 @@
 export type {
 	StdinOption,
-	StdinOptionSync,
+	StdinSyncOption,
 	StdoutStderrOption,
-	StdoutStderrOptionSync,
+	StdoutStderrSyncOption,
 } from './types/stdio/type';
 export type {Options, SyncOptions} from './types/arguments/options';
 export type {Result, SyncResult} from './types/return/result';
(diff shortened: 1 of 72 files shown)


ehmicky:
Note: another option would be put `Sync` at the end of all of those types instead: `OptionsSync`, `StdinOptionSync`, `StdoutStderrOptionSync`, `ResultSync`.

The pro is that it is consistent with the naming of the `execaSync()`, `execaCommandSync()` and `$.sync()` methods.

The con is with `ExecaSyncError`. Either:
  - We would keep it as `ExecaSyncError`, in which case its would be inconsistent with others.
  - We would change it to `ExecaErrorSync`, which would be a breaking change. Also, it is usually convention to make error classes end with `Error`.

Another con is that the `SyncOptions` -> `OptionsSync` would be a breaking change (unlike the other types, which are already renamed types, or new types).

sindresorhus:
I think what you have chosen in this PR is the way to go. The other possibility is tempting, but not worth the breaking change.

PR #1019, opened 2024-05-04
Upgrade `merge-streams`

ehmicky:
This upgrades `merge-streams` to the latest version.


lib/io/iterate.js
```diff
@@ -73,8 +74,7 @@ const iterateOnStream = ({stream, controller, binary, shouldEncode, encoding, sh
 	});
 };
 
-// @todo: replace with `getDefaultHighWaterMark(true)` after dropping support for Node <18.17.0
-export const DEFAULT_OBJECT_HIGH_WATER_MARK = 16;
```
ehmicky, line 77: `getDefaultHighWaterMark()` is already used by `merge-streams`, so we can use it too.

PR #1021, opened 2024-05-04
Fix PR commit status failing due to Codecov

ehmicky:
Codecov seems to take more than 1 hour to merge the test coverage from Linux + macOS + Windows. Due to this, the PR commit status initially fails.

I cannot find an easy to fix this, so this PR fixes it by reverting to the old behavior of only uploading the test coverage on Linux. This means Codecov will report ~99.90% test coverage even though it is actually 100%, but that's ok. :/

This also means the test coverage will decrease with this PR, making the commit status of this PR fail with Codecov.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,3 @@
+codecov:
+  notify:
+    after_n_builds: 6


ehmicky:
If I'm not mistaken, it's the Windows test coverage that is uploaded too late to Codecov, because Windows is just slower on GitHub actions. This means the difference in test coverage are the lines of code that are only run on Windows.

At the moment, we only have one line like this. However, once we merge `node-cross-spawn`

ehmicky:
It seems like there might be a way to fix this with the following Codecov configuration property: [`after_n_builds`](https://docs.codecov.com/docs/notifications#preventing-notifications-until-after-n-builds). If I understand this correctly, this means Codecov would only change the commit status once all of the 6 builds (3 OS * 2 Node versions) have completed.

This also means we would not need to skip any test coverage.

I have tried it and it _seems_ to work. Shall we try this?

sindresorhus:
Lets try 👍

ehmicky:
Seems to work in the 2 other open PRs. Let's see if it keeps working.

PR #1025, opened 2024-05-05
Improve validation and typing of signals

### LINKED ISSUE #1013: Make signal names case-sensitive
We should make signal names case-sensitive, always uppercase.

Currently, signal names passed to `subprocess.kill()` or the `killSignal` option are case-insensitive. This is inherited from Node.js behavior which was added in Node v8 in the following PR (https://github.com/nodejs/node/pull/10423). This PR does not actually mention this change of behavior (although it's clear both the committer and the reviewer were aware of it), nor is this is mentioned in an issue.

The main cons is that this would be a breaking change. 

That being said, this behavior is undocumented in both Execa and Node.js.

Also, the current behavior is inconsistent with `process.kill()` and `os.constants.signals`, which are case-sensitive and always uppercase. For example, passing `'sigterm'` or `'Sigterm'` to `subprocess.kill()` works, but passing it to `process.kill()` fails.

Furthermore, uppercase is the convention for signal names. For example, passing a lowercase signal name to the `kill` Unix utility does not work.

Finally, being stricter would allow us to provide with stricter typing. Typos in signal names are likely, especially with signals like `SIGALRM`, `SIGSTKFLT`, `SIGCHLD` vs `SIGCLD` (yes, those are two different signals...), `SIGSTOP` vs `SIGTSTP` (also two different signals) or `SIGVTALRM`. It'd be helpful to let TypeScript warn of those typos, but that's not possible if the value is case-insensitive.

### LINKED ISSUE #1014: Improve validation of signal names
Passing invalid signal names currently fails. We should improve that validation:
  - The error message should indicate the list of possible signal names
  - The validation should happen early, before spawning the subprocess or creating any internal stream

Moreover, we should make the type of the signal names stricter. Currently, we allow any `string`. It should be a union of strings instead (see #1013).

This would be for the signals passed to `subprocess.kill()` and the `killSignal` option.

ehmicky:
Fixes #1013 
Fixes #1014


package.json
```diff
@@ -52,7 +52,7 @@
 		"cross-spawn": "^7.0.3",
 		"figures": "^6.1.0",
 		"get-stream": "^9.0.0",
-		"human-signals": "^6.0.0",
+		"human-signals": "^7.0.0",
```
ehmicky, line 55: https://github.com/ehmicky/human-signals/releases/tag/7.0.0

PR #1032, opened 2024-05-09
Avoid using NodeJS global type

ehmicky:
This PR uses imports for types that currently use the global `NodeJS` namespace.

This is mostly a matter of preference, but it feels like it's always a good idea to avoid globals and use explicit imports when possible.


sindresorhus:
Conflict

sindresorhus:
@ehmicky New release?

ehmicky:
Yes, but we might want to merge #1037 first.

PR #1038, opened 2024-05-09
Fix another complexity bug with types

### LINKED ISSUE #1034: TypeScript Error `TS2590: Expression produces a union type that is too complex to represent.`
Since Execa 9.0.0, when using TypeScript 5.1 or later, the following type error appears:

```
types/subprocess/stdout.d.ts:18:4 - error TS2590: Expression produces a union type that is too complex to represent.

18  : InputOutputStream<IsInputFd<FdNumber, OptionsType>>;
```

Or:

```
types/return/result-stdout.d.ts:33:4 - error TS2590: Expression produces a union type that is too complex to represent.

 33  : ResultStdioItem<
 34  IsObjectFd<ObjectFdNumber, OptionsType>,
... 
 36  OptionsType['encoding']
 37  >;
```

Apparently the error is not emitted when using `tsd`, only when using `tsc`, which is why our type tests did not catch this.

ehmicky:
Fixes #1034.

This time for TypeScript 4.8.

### DIFF
--- a/types/subprocess/all.d.ts
+++ b/types/subprocess/all.d.ts
@@ -3,22 +3,15 @@ import type {IgnoresSubprocessOutput} from '../return/ignore.js';
 import type {Options} from '../arguments/options.js';
 
 // `subprocess.all`
-export type SubprocessAll<OptionsType extends Options = Options> = AllStream<OptionsType['all'], OptionsType>;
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't think TS 4.8 is something we should care about.

ehmicky:
That sounds good to me. I'll open an issue about TypeScript versioning.

That being said, that PR might still be useful in that it reduces the types complexity, making them run a little faster. Also the code is a little smaller.

Issue #1040, opened 2024-05-09
error TS2304: Cannot find name 'TransformStream'  'ReadableStream' 'WritableStream'


qmonmert:
Since `9.0.1`, when i do `npm install`:

````
node_modules/execa/types/stdio/type.d.ts:56:52 - error TS2304: Cannot find name 'TransformStream'.

56  | Unless<IsSync, DuplexTransform | WebTransform | TransformStream>;
                                                      ~~~~~~~~~~~~~~~

node_modules/execa/types/stdio/type.d.ts:78:63 - error TS2304: Cannot find name 'ReadableStream'.

78  | Unless<IsSync, (AsyncIterable<unknown> & ProcessStdinFd) | ReadableStream>;
                                                                 ~~~~~~~~~~~~~~

node_modules/execa/types/stdio/type.d.ts:88:19 - error TS2304: Cannot find name 'WritableStream'.

88  | Unless<IsSync, WritableStream>;
                     ~~~~~~~~~~~~~~

node_modules/execa/types/transform/normalize.d.ts:54:22 - error TS2304: Cannot find name 'TransformStream'.

54  readonly transform: TransformStream;
                        ~~~~~~~~~~~~~~~


Found 4 errors in 2 files.

Errors  Files
     3  node_modules/execa/types/stdio/type.d.ts:56
     1  node_modules/execa/types/transform/normalize.d.ts:54
```

An idea? Thanks a lot

ehmicky:
Hi @qmonmert,

It appears you might have solved this problem.
Would you like to share how to solved it so that other users with the same problem can fix it too? Thanks!

qmonmert:
@ehmicky no I don't fix it, I was not sure if it was a real problem, but if you have the same problem I reopen

ehmicky:
Which `--target` are you using?
Also, what is your version of `@types/node`?

qmonmert:
@ehmicky `"target": "ES2022"` `"lib": ["ES2022"]` `"@types/node": "20.11.25",`

ehmicky:
This is a bug from Execa.

We now support web streams, but we are importing their types from the DOM instead of from `@types/node`. Therefore, users would now need to do `--lib es2022,dom`, which they should not have to. 

I opened a PR at #1043 to solve this.

sindresorhus:
> We now support web streams, but we are importing their types from the DOM instead of from @types/node.

[...]

PR #1042, opened 2024-05-09
Run TypeScript on the types

sindresorhus:
`tsd` only validates the types, it doesn't compile the code, so running TS too is helpful (and would have caught https://github.com/sindresorhus/execa/commit/3bdab60fd1986f9600653e97d81ac5240777cf12).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,12 @@
+{
+	"compilerOptions": {
+		"module": "nodenext",
+		"moduleResolution": "nodenext",
+		"moduleDetection": "force",
+		"target": "ES2022",
+		"strict": true
+	},
+	"files": [
+		"index.d.ts"
+	]
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,6 @@
 	"compilerOptions": {
 		"module": "nodenext",
 		"moduleResolution": "nodenext",
-		"moduleDetection": "force",
 		"target": "ES2022",
 		"strict": true
 	},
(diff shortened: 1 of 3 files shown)


sindresorhus:
Not sure why this affects XO.

ehmicky:
The change in xo behavior is probably due to the addition of `tsconfig.json`.

Maybe this is related to #1040? Let's see if the problem persists after merging #1043.

sindresorhus:
> I was thinking of opening an issue on tsd, since ideally those should have been caught by tsd.

Yeah, would be nice if `tsd` could run TS too.

sindresorhus:
> Maybe this is related to https://github.com/sindresorhus/execa/issues/1040? Let's see if the problem persists after merging https://github.com/sindresorhus/execa/pull/1043.

Still failing on the same.

ehmicky:
Looking into it now. :eyes:

ehmicky:
The problem was that `tsconfig.json` adds `strict: true` (which is good).

One of the type tests needed to be tweaked to avoid an implicit conversion to `any`, which cannot be avoided due to how `instanceof Error` narrowing works in TypeScript.

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+	"compilerOptions": {
+		"module": "nodenext",
+		"moduleResolution": "nodenext",
+		"moduleDetection": "force",
```
ehmicky, line 5: The default value of `moduleDectection` is `auto`, which uses the `package.json` `type` field when `module` is `nodenext`. Since we use `type: 'module'` in `package.json`, it seems like we could remove the `moduleDetection: "force"` line?

ehmicky (approved):

PR #1046, opened 2024-05-09
Fix type of result when using the `reject: false` option

### LINKED ISSUE #1045: Wrong type for the return value when using `reject: false`
When using the `reject: false` option, the return value of `execa()` is not assignable to the `Result` type.

```ts
const rejectsResult = await execa('unicorns');
expectAssignable<Result>(rejectsResult); // This passes

const noRejectsResult = await execa('unicorns', {reject: false});
expectAssignable<Result>(noRejectsResult); // This fails, but should pass
```

This also means that passing the return value (when using `reject: false`) to a command template string fails.

```ts
const noRejectsResult = await execa('unicorns', {reject: false});
const result = await execa`echo ${noRejectsResult}`; // This fails
```

This is a type-only bug.

ehmicky:
Fixes #1045.


ehmicky:
Should we make a new patch release?

sindresorhus:
Yes, feel free to do it.

PR #1049, opened 2024-05-10
Export `TemplateExpression` type

### LINKED ISSUE #1048: breaking changes not included in v9 release notes
Hello, and thanks for the library.
In execa version 9, it no longer exports some types that I was depending on:

image

Can you advise how to use them now? And also can we document/add those breaking changes to the release notes for 9.0?

ehmicky:
Fixes #1048.

### DIFF
--- a/types/methods/template.d.ts
+++ b/types/methods/template.d.ts
@@ -1,12 +1,14 @@
 import type {Result, SyncResult} from '../return/result.js';
 
-// Values allowed inside `...${...}...` template syntax
 type TemplateExpressionItem =
 	| string
 	| number
 	| Result
 	| SyncResult;
 
+/**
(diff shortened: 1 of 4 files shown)


types/methods/template.d.ts
```diff
@@ -1,12 +1,14 @@
...
 
-// Values allowed inside `...${...}...` template syntax
 type TemplateExpressionItem =
 	| string
 	| number
 	| Result
 	| SyncResult;
 
+/**
+Value allowed inside `${...}` when using the template string syntax.
```
Zamiell, line 10: Looks like it is missing the * on this line.

ehmicky: This is not required.

PR #1054, opened 2024-05-14
Add `parseCommandString()`

### LINKED ISSUE #1016: Add `parseCommandString()` method
### Behavior

`execaCommand()` takes a command string, splits it into an array of arguments (using spaces as delimiters), then forwards that array to `execa()`. 

It also trims whitespaces, allows consecutive spaces, and lets users escape spaces with a backslash.

```js
import {execa, execaCommand} from 'execa';

await execaCommand('file one two\\ three');
// Is same as:
await execa('file', ['one', 'two three']);
```

### Origin

The reason this method was introduced was to discourage the `shell` option. Many users used that option only to be able to pass the command and its arguments as a single string, as opposed to an array. However, the `shell` option comes with many pitfalls, including a risk for command injection. `execaCommand()` provides the single string syntax, but without using a shell.

### Recent changes

However, since then, we've introduced the template string syntax. Unless the command or arguments are user-defined, it is actually now better to use the template string syntax instead of `execaCommand()`. The template string syntax has more features and provides with a simpler way to escape spaces.

```js
import {execa, execaCommand} from 'execa';

await execa`file one ${'two three'}`;
await execaCommand('file one two\\ three');
```

So `execaCommand()` is now partially redundant, although it still has some use cases (e.g. user-defined input). Taking this new situation into account, this issue is a proposal for evolving `execaCommand()` to a new purpose.

## Proposal

Rename `execaCommand()` to `splitCommand()`. It performs the same command/arguments splitting than `execaCommand()` does. However, it does not execute the command. Instead, it just returns an array with the file and its arguments. The intention is to pass that array to the template string syntax of `execa`.

Before:

```js
import {execaCommand} from 'execa';

await execaCommand('file one two\\ three');
```

After:

```js
import {execa, splitCommand} from 'execa';

await execa`${splitCommand('file one two\\ three')}`;
```

For the above case, this results in more verbose syntax. However, this comes with the following pros.

### Orthogonality

Unlike `execaCommand()`, `splitCommand()` can be used with `$`, i.e. can be used in scripts.

```js
await $`${splitCommand('file one two\\ three')}`;
```

### Escaping

`execaCommand()` uses backslashes to escape spaces. It cannot use `${}` to escape spaces like the template string syntax does. On the other hand, `splitCommand()` allows mixing both.

```js
const variableWithSpaces = 'four five';
await execa`${variableWithSpaces} ${splitCommand('one two\\ three')}`;
```

### Subprocess interpolation

`execaCommand()` cannot re-use the result from another subprocess (while properly escaping it). `splitCommand()` does.

```js
const result = await execa`...`;
await execa`${result} ${splitCommand('one two\\ three')}`;
```

### Partial interpolation

`execaCommand()` splits the file and all its arguments. `splitCommand()` can apply to only the arguments (not the file), or to only some of the arguments.

```js
await execa`echo four ${splitCommand('one two\\ three')}`;
```

### Debugging

It is simpler to debug `splitCommand()`. Users will naturally just print the array being returned. There is less magic involved: users clearly understand what's happening in terms of parsing.

```js
const commandArguments = splitCommand('one two\\ three');
console.log(commandArguments); // ['one', 'two three']
await execa`echo ${commandArguments}`;
```

### Less confusion

Explaining the purpose of `execaCommand()`, as opposed to using the template string syntax, is not easy. I struggled a little when writing the new documentation. 

I also saw some examples of `execaCommand()` being misused. For example, Storybook [has a function](https://github.com/storybookjs/storybook/blob/7e1ee32d46075bed700b4ae193e54546b5a9b0e2/code/lib/core-common/src/js-package-manager/JsPackageManager.ts#L566) which takes a command and an array of arguments as input. It joins it into a single string, then passes it to `execaCommand()`, as opposed to just pass the array of arguments directly to `execa()`. In other words they are doing something like:

```js
await execaCommand([file, ...args].join(' '));
```

Instead of:

```js
await execa(file, args);
```

With the side effect that any argument with spaces is now accidentally split into several arguments.

On the other hand, `splitCommand()`'s purpose is straightforward. 

---

ehmicky:
Fixes #1016.

`execaCommand()` is now undocumented, but its implementation and types are kept. 
I have added https://github.com/sindresorhus/execa/issues/1055 to remember to remove them in the next major release.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -83,17 +83,12 @@ This is the preferred method when executing Node.js files.
 
 [More info.](node.md)
 
-### execaCommand(command, options?)
+### parseCommandString(command)
 
 `command`: `string`\
-`options`: [`Options`](#options)\
-_Returns_: [`ResultPromise`](#return-value)
+_Returns_: `string[]`
 
-Executes a command. `command` is a string that includes both the `file` and its `arguments`.
-
-This is only intended for very specific cases, such as a [REPL](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop). This should be avoided otherwise.
-
-Just like `execa()`, this can [bind options](execution.md#globalshared-options). It can also be [run synchronously](#execasyncfile-arguments-options) using `execaCommandSync()`.
+Split a `command` string into an array. For example, `'npm run build'` returns `['npm', 'run', 'build']` and `'argument otherArgument'` returns `['argument', 'otherArgument']`.
 
 [More info.](escaping.md#user-defined-input)
 
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -12,7 +12,7 @@
 
 [`error.escapedCommand`](api.md#resultescapedcommand) is the same, except control characters are escaped. This makes it safe to either print or copy and paste in a terminal, for debugging purposes.
 
-Since the escaping is fairly basic, neither `error.command` nor `error.escapedCommand` should be executed directly, including using [`execa()`](api.md#execafile-arguments-options) or [`execaCommand()`](api.md#execacommandcommand-options).
+Since the escaping is fairly basic, neither `error.command` nor `error.escapedCommand` should be executed directly, including using [`execa()`](api.md#execafile-arguments-options) or [`parseCommandString()`](api.md#parsecommandstringcommand).
 
 ```js
 import {execa} from 'execa';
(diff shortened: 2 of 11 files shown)


karlhorky:
@ehmicky @sindresorhus thanks for this!

Which release will this be published in?

- next minor: https://github.com/sindresorhus/execa/releases/tag/v9.2.0
- next major: https://github.com/sindresorhus/execa/releases/tag/v10.0.0

sindresorhus:
9.2.0

PR #1059, opened 2024-05-16
Add promise-based IPC

### LINKED ISSUE #1053: Promise-based IPC
## Current API (callbacks)

Currently, to use Node.js IPC, one must use callbacks. 

```js
// parent.js
const subprocess = execaNode('child.js');

subprocess.send(sentMessage, error => {
  // ...
});

subprocess.once('message', oneMessage => {
  // ...
});

subprocess.on('message', eachMessage => {
  // ...
});
```

```js
// child.js

process.send(sentMessage, error => {
  // ...
}); 

process.once('message', oneMessage => {
  // ...
});

process.on('message', eachMessage => {
  // ...
});
```

## New API (promises)

To make it work better with `async`/`await`, provide with better stack traces and avoid uncaught exceptions, we should expose that API using promises instead.

```js
// parent.js
const subprocess = execaNode('child.js');

await subprocess.ipc.send(sentMessage);

const oneMessage = await subprocess.ipc.receive();

for await (const eachMessage of subprocess.ipc.iterable()) {
  // ...
};
```

```js
// child.js
import {ipc} from 'execa';

await ipc.send(sentMessage); 

const oneMessage = await ipc.receive();

for await (const eachMessage of ipc.iterable()) {
  // ...
};
```

## Additional pros

I have just looked into Node.js/libuv source code for the IPC feature to check for edge cases. On top of just providing with a promise-based API instead of callbacks, this would also benefit the additional pros.

### Graceful exit

When using `process.on('message')` or `subprocess.on('message')`, the process is kept alive. My guess is that most users probably solve this by calling either:
  - `process.exit()`: this is bad because it interrupts any ongoing logic.
  - `process.disconnect()`: this is bad because it prevents calling `process.on('message')` later. In other words, it is not possible to stop listening then restart later.

The proper way is to remove the `message` event listener, which we would automatically do.

(Note: under the hood, [Node.js listens](https://github.com/nodejs/node/blob/78dbda1bd3399a1dcb454773ed518a94f05b329f/lib/child_process.js#L180) for `addListener` and `removeListener` events on the process.)

### Error handling

It's a little hard right now for users to both use the IPC callbacks API, while at the same time `await` the Execa promise in case the subprocess failed. Using a promise-based API would ensure no subprocess failure is creating unhandled promise rejections.

### Backpressure

`subprocess.send()` returns `false` when more than ~350KB has been buffered but not sent yet. This works like streams `highWaterMark`, for backpressure. I.e. the best practice would be for users to stop sending data when `false` is returned. 

I am guessing most users might not do this. Our API would automatically provide with backpressure, since the buffer is always emptied when the `subprocess.send()` callback is fired. So awaiting the `.send()` promise would be enough to ensure the IPC channel is not under pressure.

### Better validation

If the user forgot to use the `ipc: true` option, we can provide with a nice error message. Right now, they might be wondering why `process.send` crashes when it is  `undefined`.

### Simplicity

With that API, users should never have to use `process.disconnect()`, `process.on('disconnect')`, `process.connected` or `process.channel`. This is automatically handled for them.

## Backward compatibility

The callback-based API would still be available, but undocumented.

--- 

ehmicky:
Fixes #1053.

This adds the following methods:
  - In the current process: `subprocess.sendMessage(message)`, `subprocess.getOneMessage()`, `subprocess.getEachMessage()`.
  - In the subprocess, Execa exports the following methods instead: `sendMessage(message)`, `getOneMessage()`, `getEachMessage()`.

This also exports a new type `Message` representing IPC messages.

The old IPC methods are undocumented but kept. This PR does not have any breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
 		"c8": "^9.1.0",
 		"get-node": "^15.0.0",
 		"is-running": "^2.1.0",
-		"p-event": "^6.0.0",
 		"path-exists": "^5.0.0",
 		"path-key": "^4.0.0",
 		"tempfile": "^5.0.0",
--- a/types/subprocess/subprocess.d.ts
+++ b/types/subprocess/subprocess.d.ts
@@ -11,6 +11,7 @@ import type {
 	DuplexOptions,
 	SubprocessAsyncIterable,
 } from '../convert.js';
+import type {IpcMethods} from '../ipc.js';
 import type {SubprocessStdioStream} from './stdout.js';
 import type {SubprocessStdioArray} from './stdio.js';
 import type {SubprocessAll} from './all.js';
@@ -28,18 +29,6 @@ type ExecaCustomSubprocess<OptionsType extends Options> = {
 	This is `undefined` if the subprocess failed to spawn.
 	*/
 	pid?: number;
-
-	/**
-	Send a `message` to the subprocess. The type of `message` depends on the `serialization` option.
-	The subprocess receives it as a [`message` event](https://nodejs.org/api/process.html#event-message).
-
-	This returns `true` on success.
-
-	This requires the `ipc` option to be `true`.
-
-	[More info.](https://nodejs.org/api/child_process.html#subprocesssendmessage-sendhandle-options-callback)
-	*/
-	send: HasIpc<OptionsType> extends true ? ChildProcess['send'] : undefined;
 
 	/**
 	The subprocess [`stdin`](https://en.wikipedia.org/wiki/Standard_streams#Standard_input_(stdin)) as a stream.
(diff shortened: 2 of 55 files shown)


package.json
```diff
@@ -70,7 +70,6 @@
 		"c8": "^9.1.0",
 		"get-node": "^15.0.0",
 		"is-running": "^2.1.0",
-		"p-event": "^6.0.0",
```
ehmicky, line 73: `devDependencies` used only in tests. Replaced by `once()` from `node:events`.

types/subprocess/subprocess.d.ts
```diff
@@ -29,18 +30,6 @@ type ExecaCustomSubprocess<OptionsType extends Options> = {
...
-	Send a `message` to the subprocess. The type of `message` depends on the `serialization` option.
-	The subprocess receives it as a [`message` event](https://nodejs.org/api/process.html#event-message).
-
-	This returns `true` on success.
-
-	This requires the `ipc` option to be `true`.
-
-	[More info.](https://nodejs.org/api/child_process.html#subprocesssendmessage-sendhandle-options-callback)
-	*/
-	send: HasIpc<OptionsType> extends true ? ChildProcess['send'] : undefined;
```
ehmicky, line 42: This is not a breaking change: `Subprocess['send']` is still defined.

PR #1066, opened 2024-05-17
Add `Methods` type

### LINKED ISSUE #1062: Export `Execa` and `ExecaScript` types.
The "currying" capability built-in to `execa()` and `$` are nice, but I would like to be able to correctly type curried return values.

For example...

```ts
const executeSomething = async (exec: typeof execa = execa): Promise<void> => {
  // ...
});

// type error
await executeSomething(execa({ encoding: 'buffer' }));
```

The type error is...

```
Argument of type 'Execa<{ encoding: "buffer"; }>' is not assignable to parameter of type 'Execa<{}>'.
  Type 'ResultPromise<{ encoding: "buffer"; }>' is not assignable to type 'ResultPromise<{}>'.
    Type 'ResultPromise<{ encoding: "buffer"; }>' is not assignable to type '{ pid?: number | undefined; send: undefined; stdin: Writable; stdout: Readable; stderr: Readable; all: undefined; stdio: SubprocessStdioArray<{}>; ... 5 more ...; [Symbol.asyncIterator](): SubprocessAsyncIterable<...>; }'.
      The types returned by '[Symbol.asyncIterator]()' are incompatible between these types.
        Type 'SubprocessAsyncIterable<undefined, "buffer">' is not assignable to type 'SubprocessAsyncIterable<undefined, unknown>'.
          Type 'Uint8Array' is not assignable to type 'string'. ts(2345)
```

Instead of using `typeof execa`, it would be great to use `Execa<Options & { encoding: "buffer" }>` (or similar), but that's not possible because the `Execa` type is not exported. And the same goes for `ExecaScript`.

ehmicky:
Fixes #1062.

This PR adds the `Methods['execa']`, `Methods['execaNode']` and `Methods['$']` types. Those are the types of `execa`/`execa(options)`, and so on.

This is useful for TypeScript users passing bound/curried instances of `execa` around.

Using `ReturnType<execa>` does not work due to the overloading of the `execa()` function (`execa(options)`, `execa(file, args, options)`, template string syntax). So, `ReturnType<execa>` is the return type of `execa(options)` but also of `execa(file, args, options)` (i.e. a `ResultPromise`), which makes assigning it not work.

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -8,11 +8,11 @@
 
 ## Available types
 
-The following types can be imported: [`ResultPromise`](api.md#return-value), [`Subprocess`](api.md#subprocess), [`Result`](api.md#result), [`ExecaError`](api.md#execaerror), [`Options`](api.md#options-1), [`StdinOption`](api.md#optionsstdin), [`StdoutStderrOption`](api.md#optionsstdout), [`TemplateExpression`](api.md#execacommand) and [`Message`](api.md#subprocesssendmessagemessage-sendmessageoptions).
+The following types can be imported: [`ResultPromise`](api.md#return-value), [`Subprocess`](api.md#subprocess), [`Result`](api.md#result), [`ExecaError`](api.md#execaerror), [`Options`](api.md#options-1), [`StdinOption`](api.md#optionsstdin), [`StdoutStderrOption`](api.md#optionsstdout), [`TemplateExpression`](api.md#execacommand), [`Message`](api.md#subprocesssendmessagemessage-sendmessageoptions), [`ExecaMethod`](api.md#execaoptions), [`ExecaNodeMethod`](api.md#execanodeoptions) and [`ExecaScriptMethod`](api.md#options).
 
 ```ts
 import {
-	execa,
+	execa as execa_,
 	ExecaError,
 	type ResultPromise,
 	type Result,
@@ -21,7 +21,10 @@ import {
 	type StdoutStderrOption,
 	type TemplateExpression,
 	type Message,
+	type ExecaMethod,
 } from 'execa';
+
+const execa: ExecaMethod = execa_({preferLocal: true});
 
 const options: Options = {
 	stdin: 'inherit' satisfies StdinOption,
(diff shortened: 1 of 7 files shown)


types/methods/list.d.ts
```diff
@@ -0,0 +1,33 @@
+import type {Execa} from './main-async.js';
+import type {ExecaNode} from './node.js';
+import type {ExecaScriptCommon, ExecaScriptSync} from './script.js';
+
+// The types seem to fail unless type aliases are used.
+// We do not know why.
```
ehmicky, line 6: I cannot explain this. 
The types fail unless type aliases are used. :thinking:

Shakeskeyboarde:
On mobile right now, so I'll check tomorrow, but I'm a little skeptical that the new exports are subscript indexes of the Methods type, instead id just the types themselves. Also, the new exports aren't generic, so while the return type from currying might be assignable to them, I think we're loosing type information.

ehmicky:
> Also, the new exports aren't generic, so while the return type from currying might be assignable to them, I think we're loosing type information.

The generic type is the `Options` being used. For example, when the `encoding` option is `'buffer'`, `result.stdout` is a `Uint8Array`.

Those method types should not be used where [type inference](https://github.com/sindresorhus/execa/blob/main/docs/typescript.md#type-inference) is possible instead. When using type inference, the `Options` generic type is inferred and kept. For example:

```ts
import {execa as execa_} from 'execa';

// `execa` type is inferred as `Execa<{preferLocal: true}>`
const execa = execa_({preferLocal: true});
```

So the main reason for those new types is when the Execa methods are passed as arguments to a function. If we added the `Options` generic type, users could specify them to avoid losing type information, as you mention.

```ts
import {execa as execa_, type Methods} from 'execa';

const runCommand = async (execaMethod: Methods['execa']<{preferLocal: true}>) => {
  // ...
}

const execa = execa_({preferLocal: true});
await runCommand(execa);
```

However, using `typeof` would be simpler and result in more accurate types.

```ts
import {execa as execa_, type Methods} from 'execa';

const runCommand = async (execaMethod: typeof execa) => {
  // ...
}

const execa = execa_({preferLocal: true});
await runCommand(execa);
```

So the only use case IMHO would for if that function wants to enforce for the caller to use a specific option. For instance, the example above would enforce that `preferLocal` is `true`.

That being said, we are using `Options` as a generic type in most of the other types (`ResultPromise`, `Subprocess`, `Result`, `ExecaError`), so that does make sense to be consistent. I will update the PR accordingly.

> On mobile right now, so I'll check tomorrow, but I'm a little skeptical that the new exports are subscript indexes of the Methods type, instead of just the types themselves.

This is mostly a stylistic issue. The exported types would be the same, so this should not change your usage of those types. 

The reasoning behind grouping those types together under a `Methods` type is:
- Making it clear what those types using the word "method".
- Re-using the same name as the exported method. In particular, `$`'s type is named `$`, as opposed to be renamed something like `ExecaScript`.
- Grouping those types since they have the same purpose.

An alternative would be to have those types end with the same suffix, e.g. `ExecaMethod`, `ExecaNodeMethod` and `ExecaScriptMethod`. 

@sindresorhus Which names would you prefer? I am happy to rename them to whichever you like better.

ehmicky:
Adding the `Options` generic type seems to run into some type issue. This seems a little complicated, and might be some time to fix, so I am putting this PR back to draft.

ehmicky:
I found out two TypeScript bugs which were blocking this PR, but worked around it. 

The new exported types are: `ExecaMethod`, `ExecaNodeMethod`, `ExecaScriptMethod`, `ExecaSyncMethod`, `ExecaScriptSyncMethod`. They take an optional `Options`/`SyncOptions` as a generic type.

PR #1068, opened 2024-05-18
Add `ipcInput` option

ehmicky:
This PR adds an `ipcInput` option, which allows passing almost any value to a subprocess. It sends that value using `subprocess.sendMessage()` under the hood.

As opposed to `argv`, environment variables and `stdin`, it is not limited to strings/`Uint8Array`. Any value that works with the structured clone algorithm works.

Unlike them though, it only works if the subprocess uses Node.js.

Example:

```js
// main.js
import {execaNode} from 'execa';

const ipcInput = [
	{task: 'lint', ignore: /test\.js/},
	{task: 'copy', files: new Set(['main.js', 'index.js']),
}];
await execaNode({ipcInput})`build.js`;
```
```js
// build.js
import {getOneMessage} from 'execa';

const ipcInput = await getOneMessage();
```


sindresorhus:
Conflict

ehmicky:
Fixed. :+1:

Issue #1070, opened 2024-05-19
Avoid race conditions with IPC


ehmicky:
## Problem

The following example has a subtle bug:

```js
await subprocess.sendMessage('ping');
const pong = await subprocess.getOneMessage();
```

It has a race condition. It is technically possible for the `message` listener to be setup after the subprocess has already replied and the `message` event has been already emitted. This means `getOneMessage()` will hang forever.

This bug does happen in practice. For example, it's happened in our tests in GitHub actions on the Windows machines (since they are slower) and in test files running many tests in parallel.

## Solution

The proper way to handle this is:

```js
const [pong] = await Promise.all([
  subprocess.getOneMessage(),
  subprocess.sendMessage('ping'),
]);
```

To ensure users do not run in this race condition, we should create a new method that makes that pattern easy to use, and easy to explain in our documentation. 

For example, that method could be called `exchangeMessage()` and would run the same as above.

```js
const pong = await subprocess.exchangeMessage('ping');
```

---

Note: this is only a problem when the user first sends a message, then receives one. When they do the reverse, there is no race condition.

```js
const ping = await subprocess.getOneMessage();
await subprocess.sendMessage('pong');
```

Also, when the order is:
  - Send then receive: users usually just want to wait for the response. Therefore, a single `exchangeMessage()` method is intuitive.
  - Receive then send: users will usually do some logic in between those two calls. Either because the send call notifies that some operation has been performed, or because the sent message is based on the received message.

PR #1087, opened 2024-05-23
Debounce `message` events

ehmicky:
The following pattern has a race condition described in #1070:

```js
await subprocess.sendMessage('ping');
const pong = await subprocess.getOneMessage();
```

That's because the subprocess might respond between those two calls, in which case `subprocess.getOneMessage()` hangs forever. In #1077, we introduced a `subprocess.exchangeMessage()` to fix this problem.

However, to avoid race conditions, users must remember to use that method. Also, there is a series of "Best practices" that I documented: in essence, there are many pitfalls to avoid to make sure IPC does not run into similar race conditions as above.

This PR implements a better solution to the underlying problem: it debounces IPC messages. This removes any race condition not only from the example above, but also from the following additional situations.

```js
// Read single messages, multiple times in a row
while (/* ... */) {
  const message = await subprocess.getOneMessage();
  // ...
}
```

```js
// Send multiple messages in a row, then read one message
await subprocess.sendMessage('foo');
await subprocess.sendMessage('bar');
const message = await subprocess.getOneMessage();
```

This also removes the need for the `exchangeMessage()` method, which simplifies the API. It was a little confusing whether this method sent then received a message, or received then sent a message. Since sending and receiving messages are now separate methods, this is much clearer.

This PR also adds lots of tests to cover several edge cases.


sindresorhus:
Much nicer 🙏

PR #1098, opened 2024-05-25
Add `strict` option to `sendMessage()`

ehmicky:
This PR adds a `strict` option to `sendMessage()`.

When set to `true`, `sendMessage()` throws if the other process is not currently listening/receiving messages.

This is a measure that helps users identifying some timing-related race condition bugs, where they intend to listen to a message, but do it too late.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -113,7 +113,7 @@ _Type_: `object`
 _Type_: `boolean`\
 _Default_: `false`
 
-If `true`, throws when the other process is not receiving or listening to messages.
+Throw when the other process is not receiving or listening to messages.
 
 [More info.](ipc.md#ensure-messages-are-received)
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -92,9 +92,10 @@ Split a `command` string into an array. For example, `'npm run build'` returns `
 
 [More info.](escaping.md#user-defined-input)
 
-### sendMessage(message)
+### sendMessage(message, sendMessageOptions?)
 
 `message`: [`Message`](ipc.md#message-type)\
+`sendMessageOptions`: [`SendMessageOptions`](#sendmessageoptions)\
 _Returns_: `Promise<void>`
 
 Send a `message` to the parent process.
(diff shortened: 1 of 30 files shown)


docs/api.md
```diff
@@ -103,9 +104,22 @@ This requires the [`ipc`](#optionsipc) option to be `true`. The [type](ipc.md#me
...
+#### sendMessageOptions
+
+_Type_: `object`
+
+#### sendMessageOptions.strict
+
+_Type_: `boolean`\
+_Default_: `false`
+
+If `true`, throws when the other process is not receiving or listening to messages.
```
sindresorhus, line 116:
```suggestion
Throw when the other process is not receiving or listening to messages.
```

docs/ipc.md
```diff
@@ -91,6 +92,40 @@ for await (const message of getEachMessage()) {
 }
 ```
 
+## Ensure messages are received
+
+When a message is sent by one process, the other process must receive it using [`getOneMessage()`](#exchanging-messages), [`getEachMessage()`](#listening-to-messages), or automatically with [`result.ipcOutput`](api.md#resultipcoutput). If not, that message is silently discarded.
+
+If the [`strict: true`](api.md#sendmessageoptionsstrict) option is passed to [`subprocess.sendMessage(message)`](api.md#subprocesssendmessagemessage-sendmessageoptions) or [`sendMessage(message)`](api.md#sendmessagemessage-sendmessageoptions), an exception is thrown instead. This helps identifying subtle race conditions like the following example.
```
sindresorhus, line 99:
```suggestion
If the [`strict: true`](api.md#sendmessageoptionsstrict) option is passed to [`subprocess.sendMessage(message)`](api.md#subprocesssendmessagemessage-sendmessageoptions) or [`sendMessage(message)`](api.md#sendmessagemessage-sendmessageoptions), an error is thrown instead. This helps identifying subtle race conditions like the following example.
```

sindresorhus (changes requested):

Issue #1097, opened 2024-05-25
Node.js engine version in package.json is not up-to-date


iiroj:
The release logs for [Execa 9.0.0](https://github.com/sindresorhus/execa/releases/tag/v9.0.0) state the following Node.js version support:

> Dropped support for Node.js version <18.19.0 and 20.0.0 - 20.4.0. (https://github.com/sindresorhus/execa/commit/834e3726f0c7b7f23827b6e09b69b66dedbddd47)

But in the current `package.json` it only requires:

https://github.com/sindresorhus/execa/blob/b959030b5acacf10d2908fbe261670f7af46e355/package.json#L19-L21

Is this intentional, or could the engine version be updated to:

```json
{
    "engines": {
        "node": "^18.19 || >20.4.0"
    }
}
```

This range should only match the following Node.js versions:

```
18.19.0, 18.19.1, 18.20.0, 18.20.1, 18.20.2, 18.20.3, 20.10.0, 20.11.0, 20.11.1, 20.12.0, 20.12.1, 20.12.2, 20.13.0, 20.13.1, 20.5.0, 20.5.1, 20.6.0, 20.6.1, 20.7.0, 20.8.0, 20.8.1, 20.9.0, 21.0.0, 21.1.0, 21.2.0, 21.3.0, 21.4.0, 21.5.0, 21.6.0, 21.6.1, 21.6.2, 21.7.0, 21.7.1, 21.7.2, 21.7.3, 22.0.0, 22.1.0, 22.2.0
```

I can contribute a Pull Request for this if you think it would be a good change. The reason I'm interested in this that I'm trying to follow the minimum supported Node.js version in [lint-staged](https://github.com/lint-staged/lint-staged) and usually it's up to the direct dependencies like `execa`. I have this simple script in the repo which prints out the all the engine versions, but it feels like `execa` is not correct at the moment: https://github.com/lint-staged/lint-staged/blob/master/scripts/list-dependency-node-versions.js

ehmicky:
Hi @iiroj,

Thanks for opening this issue.
This sounds good to me. @sindresorhus What do you think?

iiroj:
If you want to avoid Node.js 19 entirely, I guess something like this might work: `^18.19 < 19 || >20.4.0`

ehmicky:
The `^` only includes versions within the same major version, so `^18.19 || >20.4.0` and `^18.19 <19 || >20.4.0` are equivalent.

```js
import {satisfies} from 'semver'

[...]

PR #1101, opened 2024-05-26
Update required Node.js semver range to "^18.19.0 || >=20.5.0"

iiroj:
Fixes https://github.com/sindresorhus/execa/issues/1097 by updating the `package.json` `engines.node` semver range to match the currently supported Node.js versions: `^18.19.0 || >=20.5.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
 	},
 	"sideEffects": false,
 	"engines": {
-		"node": ">=18"
+		"node": "^18.19.0 || >=20.5.0"
 	},
 	"scripts": {
 		"test": "npm run lint && npm run unit && npm run type",


package.json
```diff
@@ -17,7 +17,7 @@
 	},
 	"sideEffects": false,
 	"engines": {
-		"node": ">=18"
+		"node": "^18.9.0 || >=20.5.0"
```
ehmicky, line 20:
```suggestion
		"node": "^18.19.0 || >=20.5.0"
```

ehmicky:
@sindresorhus Is this technically a breaking change for users that use [`engine-strict: true`](https://docs.npmjs.com/cli/v6/using-npm/config#engine-strict)? In principle, that means they are using an unsupported Node.js version, which should not work at all. So I guess it's ok, just wanted to point it out.

package.json
iiroj: Oops sorry, I copy-pasted this from the issue. 🙈

iiroj: Fixed

iiroj:
My personal opinion is that it was communicated in the changelog for version 9, so this is just a _fix_ to make sure the value is up-to-date.

ehmicky (approved):

ehmicky:
Good point @iiroj.

PR #1103, opened 2024-05-26
fix: add default 0 exitCode when checking isFailedExit

### LINKED ISSUE #1102: Subprocess is considered as failed exit when using `process.kill` on windows only
# Reproduction

https://github.com/huang-julien/execa-repro

## Steps

- clone it on windows
- run index.js with WSL
- it exit after 5 sec without issue
- run again but with powershell or cmd
- it crash after 5 sec (terminated with SIGTERM)

# Description 

Hello :wave:

This is an issue we detected on https://github.com/nuxt/test-utils/pull/848

On windows, with using `process.exit()` without exitCode, the exitCode received by execa is `null` while it is `0` on linux.

huang-julien:
fix #1102

### DIFF
--- a/lib/resolve/exit-async.js
+++ b/lib/resolve/exit-async.js
@@ -43,6 +43,6 @@ export const waitForSuccessfulExit = async exitPromise => {
 };
 
 // When the subprocess fails due to an `error` event
-const isSubprocessErrorExit = (exitCode, signal) => exitCode === undefined && signal === undefined;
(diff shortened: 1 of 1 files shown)


ehmicky:
Thanks @huang-julien!

Let's first figure out the underlying problem in the issue at #1102. I am suspecting it to be slightly different that it seems.

ehmicky:
Closing per https://github.com/sindresorhus/execa/issues/1102#issuecomment-2146787524

PR #1107, opened 2024-05-28
Add `reference` option to `getOneMessage()` and `getEachMessage()`

ehmicky:
This PR adds a `ref` option to `getOneMessage()` and `getEachMessage()`. Setting it to `false` allows the subprocess to exit even if those methods are still waiting for a message to be received.

This is useful in situations where the subprocess does not know whether a message will be sent by the current process, or not. For example, a message to notify of an early termination / graceful exit.

This is called `ref` instead of `reference` to be consistent with the name of the `.ref()`/`.unref()` methods in the core Node.js API, and also with the `ref: boolean` option of [`setTimeout()`](https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options)/[`setImmediate()`](https://nodejs.org/api/timers.html#timerspromisessetimmediatevalue-options).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -140,8 +140,18 @@ Ignore any `message` that returns `false`.
 
 [More info.](ipc.md#filter-messages)
 
-### getEachMessage()
+#### getOneMessageOptions.reference
 
+_Type_: `boolean`\
+_Default_: `true`
+
+Keep the subprocess alive while `getOneMessage()` is waiting.
+
+[More info.](ipc.md#keeping-the-subprocess-alive)
+
+### getEachMessage(getEachMessageOptions?)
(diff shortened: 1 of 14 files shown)


sindresorhus:
> This is called ref instead of reference to be consistent with the name of the .ref()/.unref() methods in the core Node.js API, and also with the ref: boolean option of setTimeout()/setImmediate().

I don't think it's worth staying consistent with those APIs. We want Execa to be better than Node.js APIs.

sindresorhus:
I like `reference`, but `keepAlive` may also work.

ehmicky:
Renamed to `reference`. :+1:

PR #1113, opened 2024-06-03
Document passing `[]` or `''` to template string syntax

### LINKED ISSUE #1112: Document idiomatic ternary interpolation pattern?
Hey, thanks for execa!

Just wanted to ask whether it's worth documenting this ternary pattern:

```ts
let branch; // potentially undefined

await execa`git clone https://github.com/sindresorhus/execa.git ${
  branch ?
    `--branch ${branch}` :
    [] // Will not add anything to the command string, where an empty string will
}`;
```

I needed this today, and the only place that I could find the `[]` option was the tests:

https://github.com/sindresorhus/execa/blob/a5a4d69aa6f5d9171975d952dd045b49d93ed797/test/methods/template.js#L37

I tried to confirm my suspicion that `''` was causing a problem by inspecting the `escapedCommand`, which does indeed have an extra `''`:

```
> await execa`echo ${''} abc`
{
  command: 'echo  abc',
  escapedCommand: "echo '' abc",
  cwd: '/Users/k/p/preflight',
  durationMs: 5.451958,
  failed: false,
  timedOut: false,
  isCanceled: false,
  isTerminated: false,
  isMaxBuffer: false,
  exitCode: 0,
  stdout: ' abc',
  stderr: '',
  stdio: [ undefined, ' abc', '' ],
  pipedFrom: []
}
```

Using an empty array (`[]`) does not make anything extra show up in the `escapedCommand`:

```
> await execa`echo ${[]} abc`
{
  command: 'echo abc',
  escapedCommand: 'echo abc',
  cwd: '/Users/k/p/preflight',
  durationMs: 11.123583,
  failed: false,
  timedOut: false,
  isCanceled: false,
  isTerminated: false,
  isMaxBuffer: false,
  exitCode: 0,
  stdout: 'abc',
  stderr: '',
  stdio: [ undefined, 'abc', '' ],
  pipedFrom: []
}
```

ehmicky:
Fixes #1112.

### DIFF
--- a/docs/execution.md
+++ b/docs/execution.md
@@ -55,7 +55,30 @@ await execa`mkdir ${tmpDirectory}/filename`;
 ```js
 const result = await execa`get-concurrency`;
 
-await execa`npm ${['run', 'build', '--concurrency', 2]}`;
+await execa`npm ${['run', 'build', '--concurrency', result]}`;
+```
+
+### No arguments
+
+```js
+await execa`npm run build ${[]}`;
+// Same as:
+await execa('npm', ['run', 'build']);
+```
+
+### Empty string argument
+
+```js
+await execa`npm run build ${''}`;
+// Same as:
+await execa('npm', ['run', 'build', '']);
+```
+
+### Conditional argument
+
+```js
+const commandArguments = failFast ? ['--fail-fast'] : [];
+await execa`npm run build ${commandArguments}`;
 ```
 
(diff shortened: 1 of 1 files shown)


karlhorky (approved):
Looks good to me, thanks for this!

karlhorky:
I would suggest also documenting the ternary pattern if possible - could be a separate PR though

PR #1120, opened 2024-06-05
Automatically check Markdown links

ehmicky:
Our documentation has lots of links and anchors. It actually takes quite some time to ensure the links are correct. Also, external links might stop working without us knowing.

This adds a GitHub action to automatically check Markdown links, including anchors. 

This is hard to believe, but apparently it runs in only 2 seconds. This is written in Rust, but I would expect hundreds of parallel HTTP requests to take longer, but it seems not! :)

This PR currently fails (which is good since it shows the action works) due to the broken links that are fixed in a separate PR at #1119.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,6 +22,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
+      - uses: lycheeverse/lychee-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


sindresorhus:
> Failed: Network error: Too Many Requests

Seems to fail with too many requests towards a single host. Seems like the tool should limit concurrency per host to avoid this.

ehmicky:
There is a `max-concurrency` option, although it's not per-host. There's also an `exclude` option. I have added both, and it now seems to work.

PR #1139, opened 2024-08-06
refactor: remove `is-plain-obj`

Jayllyz:
## Changes

Removed `is-plain-obj` in favor of a oneline function.

## Context

As part of the ongoing [ecosystem cleanup](https://github.com/es-tooling/ecosystem-cleanup), we are migrating various projects to use lighter/faster package or alternatives.

image

*https://github.com/TheDevMinerTV/package-size-calculator*

## How I've tested my work

`npm run test` 4982 tests passed

### DIFF
--- a/lib/arguments/specific.js
+++ b/lib/arguments/specific.js
@@ -1,5 +1,5 @@
 import {debuglog} from 'node:util';
-import isPlainObject from 'is-plain-obj';
+import isPlainObject from '../utils/is-plain-object.js';
 import {STANDARD_STREAMS_ALIASES} from '../utils/standard-stream.js';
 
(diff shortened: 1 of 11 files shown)


sindresorhus:
Not sure I see the point of this. This will have very little impact on the cleanup effort. The real problems in the ecosystem is that certain packages support too low Node.js versions and include polyfills for everything. This also feels like a slippery slope to inline even more, which means code gets duplicated, and potential bug fixes or improvements don't get included.

The proper fix would be to convince TC39 to add a method like this to JS: https://es.discourse.group

ehmicky:
100% agree with @sindresorhus. :+1:

PR #1140, opened 2024-08-06
Document how to terminate hanging subprocesses

### LINKED ISSUE #1137: Timeout / cancel stuck process
Hey,

sorry for asking a question as an issue, I thought it might be a useful feature if it doesn't already exist. Execa has options for [`timeout`](https://github.com/sindresorhus/execa/blob/main/docs/termination.md#timeout) and [cancel signal](https://github.com/sindresorhus/execa/blob/main/docs/termination.md#canceling). 

I had a few cases where a process running a Node.js script got stuck because of an error, but the script didn't properly throw this error so the process kept running but didn't print anything to sdtout/stderr.

In such cases it would be nice to have timeout that resets after each update on on sdtout/stderr,  so that you can distinguish a long running process from a stuck process.

ehmicky:
Fixes #1137.

I reworked the example so it avoids `subprocess.stdout`/`subprocess.stderr` Node.js streams since we've been trying to recommend alternatives (transforms and/or async iteration), so that users can `await` the promise returned by `execa()` directly.

### DIFF
--- a/docs/termination.md
+++ b/docs/termination.md
@@ -124,6 +124,8 @@ If the subprocess is still alive after 5 seconds, it is forcefully terminated wi
 
 ## Timeout
 
+### Execution timeout
+
 If the subprocess lasts longer than the [`timeout`](api.md#optionstimeout) option, a [`SIGTERM` signal](#default-signal) is sent to it.
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/termination.md
```diff
@@ -138,6 +140,51 @@ try {
...
+	const controller = new AbortController();
+	const cancelSignal = controller.signal;
+
+	// Delay and debounce `cancelSignal` each time `controller.abort()` is called
+	const abort = debounceFn(controller.abort.bind(controller), {wait});
+
+	const onOutput = {
+		* transform(data) {
+			// When anything is printed, debounce `controller.abort()`
+			abort();
```
sindresorhus, line 164:
```suggestion
			scheduleAbort();
```

Just to make it clear it doesn't happen right away?

ehmicky: Fixed in 493b3661855bd3b1c5f85dde2352682d25f28f8f

Issue #1132, opened 2024-06-26
Unable to extend env when `ProcessEnv` has been manipulated


ehyland:
When a non optional property has been defined on `ProcessEnv` by [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces), it is not possible to use execa's `env` & `extendEnv` options to add additional environment variables to execa process. 

[Typescript playground example](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgUwB7IMYEM4F84BmUEIcA5GplmQNwCwAUKJLOQAJTIjCoC0UAVwB2AeiEQAJsjKNGIgFSMAKgAtgAZ0JZgAG01ZNHLj37Cxk5ClQxkQiZpgrLABWIZk69QFEhANzjAQjZQBFjusgwiIoTEpEbcfIKi4lKMgcGh7nCuEO6ePv4IjHBwAHIA8gAiXgD6XqUAagBccABEUr7IOhBgILYwrXAAPm1gxBICGDDAEEKDI6026gP0DLiM8iIRjBizy3AAJHAAvFZUABRFDCXLEjMtrYFOUMADADTFVjZ2BS0wgsgPtcUH4WlcSiVKgBBJRQgBCUIAyrUAKoAJQAMi0yNDYQjkTV0RiyECSrggbgAJSrCJYADu2ngBwABmMNCAcBIAEZwMACdQqOC8XhhdxgGC8CRYGBYXjdTzMmkMIA) 

## Example

```ts
import { config } from "~/config.server";
import { execa } from "execa";

// from @remix-run/node
interface ProcessEnv {
  NODE_ENV: "development" | "production" | "test";
}

// attempting to extend process env results in TS error "No overload matches this call."
const $ = execa({
  stdio: "inherit",
  extendEnv: true,
  env: {
    DATABASE_URL: config.DATABASE_URL,
  },
});

await $`prisma db push --accept-data-loss`;
``` 

## Suggestion

Instead of [`env` option](https://github.com/sindresorhus/execa/blob/6c6e8611a7834f52be9070b3de29e4de466acfc8/types/arguments/options.d.ts#L80) being typed as `typeof import('node:process').env`. Maybe it should be typed as `Record<string, string | undefined>`

ehmicky:
Hi @ehyland,

[...]

PR #1141, opened 2024-08-09
define env type

holic:
fixes https://github.com/sindresorhus/execa/issues/1132

I've seen the issue described in #1132 happen in both Remix and Next.js projects

### DIFF
--- a/types/arguments/options.d.ts
+++ b/types/arguments/options.d.ts
@@ -76,7 +76,7 @@ export type CommonOptions<IsSync extends boolean = boolean> = {
 
 	@default [process.env](https://nodejs.org/api/process.html#processenv)
 	*/
-	readonly env?: Partial<Record<string, string>>;
+	readonly env?: Readonly<Partial<Record<string, string>>>;
 
 	/**
 	If `true`, the subprocess uses both the `env` option and the current process' environment variables ([`process.env`](https://nodejs.org/api/process.html#processenv)).
--- a/types/arguments/options.d.ts
+++ b/types/arguments/options.d.ts
@@ -1,5 +1,4 @@
 import type {SignalConstants} from 'node:os';
-import type {env} from 'node:process';
 import type {Readable} from 'node:stream';
 import type {Unless} from '../utils.js';
 import type {Message} from '../ipc.js';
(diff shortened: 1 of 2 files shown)


ehmicky:
Hi @holic,

Thanks for this PR.

The issue at #1132 was specific to Remix. It is due to a bug in Remix currently tracked in https://github.com/remix-run/remix/issues/9718. It impacts `node:child_process`, and is not specific to Execa. So there's nothing to fix in Execa.

When it comes to Next.js, could you please show an example of a type error when using both Execa and Next.js?

ehmicky:
So basically the core problem is this: the following pattern is bad.

```ts
// Augment `ProcessEnv` type from Node.js
interface ProcessEnv { 
  NODE_ENV: "development" | "production" | "test"; 
} 
```

The intent is to change the type of `process.env.NODE_ENV`. However, `ProcessEnv` is also used by Node.js to type the `env` option of `node:child_process`. So the above creates unintentional problems.

Unfortunately, this pattern was introduced by Next.js, which was seemingly copied by Remix, and might have been copied by other libraries.

Although this is not a bug with Execa, and since getting that bug fixed in either Next.js or Remix seems difficult, I would suggest for Execa to patch that typing problem.

From that perspective, your solution looks good. However, could you please add some type tests? Specifically, a type test that does the `interface Process { ... }` augmentation, then calling Execa with the `env` option, and make sure the types work in that case. Thanks!

holic:
I agree the original pattern and source of this issue isn't great (haven't we learned enough that fiddling with globals is a bad idea?) but seems easy to simplify types here while maintaining intended use and compatibility.

Happy to add some tests! Will do in the next day or so.

holic:
Added! Thanks for making that an easy copy+paste job for me :)

types/arguments/options.d.ts
```diff
@@ -77,7 +76,7 @@ export type CommonOptions<IsSync extends boolean = boolean> = {
 
 	@default [process.env](https://nodejs.org/api/process.html#processenv)
 	*/
-	readonly env?: typeof env;
+	readonly env?: Partial<Record<string, string>>;
```
ehmicky, line 79:
```suggestion
	readonly env?: Readonly<Partial<Record<string, string>>>;
```

ehmicky (approved):
Looks good to me, except the one small suggestion above.

To be more explicit, the previous type of `options.env` was:

```ts
{ 
  TZ?: string;
  [key: string]: string | undefined;
}
```

And the new type is:

```ts
{ 
  readonly [key: string]: string | undefined;
}
```

This is backward compatible with the former one, i.e. this is not a breaking change.

sindresorhus:
👍

ehmicky:
Thanks @holic!

PR #1144, opened 2024-08-13
Fix Node.js without ICU support

### LINKED ISSUE #1143: Intl needed for v. 9
node builds without intl are failing due to this line

https://github.com/sindresorhus/execa/blob/c0b6efc9cbf8fa88398ccc863a8eb554d3a76f93/lib/arguments/escape.js#L41

execa is used by borp @mcollina. borp is used in undici, thus breaking the without intl unit tests. 

```js
> borp@0.17.0 unit
> node borp.js --ignore "fixtures/**/*" --coverage --coverage-exclude "fixtures/**/*" --coverage-exclude "test*/**/*"

file:///home/runner/work/borp/borp/node_modules/execa/lib/arguments/escape.js:41
const SPECIAL_CHAR_REGEXP = /\p{Separator}|\p{Other}/gu;
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^

SyntaxError: Invalid regular expression: /\p{Separator}|\p{Other}/: Invalid property name
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:152:18)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:299:14)
```

ehmicky:
Fixes #1143.

This PR does not add any documentation because this is a very tiny edge case that almost nobody will encounter:
  - Running Node.js without ICU support requires building your own Node.js binary. No pre-built binary without ICU support seems to be available.
  - Then, the user would need to use very special characters in the command, such as isolated surrogate characters or reserve Unicode codepoints (which is quite unlikely) to notice any difference.
  - Even then, the only difference would only be that those special characters would be printed as is on the terminal (when using `result.escapedCommand`) instead of being represented as a backslash sequence `\u...`, which is not a big problem.

### DIFF
--- a/test/arguments/escape-no-icu.js
+++ b/test/arguments/escape-no-icu.js
@@ -1,0 +1,16 @@
+// Mimics Node.js when built without ICU support
+// See https://github.com/sindresorhus/execa/issues/1143
+globalThis.RegExp = class extends RegExp {
+	constructor(regExpString, flags) {
+		if (flags?.includes('u') && regExpString.includes('\\p{')) {
+			throw new Error('Invalid property name');
+		}
+
+		super(regExpString, flags);
+	}
+
+	static isMocked = true;
+};
+
+// Execa computes the RegExp when first loaded, so we must delay this import
+await import('./escape.js');
--- a/lib/arguments/escape.js
+++ b/lib/arguments/escape.js
@@ -38,7 +38,23 @@ const escapeControlCharacter = character => {
 // Some shells do not even have a way to print those characters in an escaped fashion.
 // Therefore, we prioritize printing those safely, instead of allowing those to be copy-pasted.
 // List of Unicode character categories: https://www.fileformat.info/info/unicode/category/index.htm
-const SPECIAL_CHAR_REGEXP = /\p{Separator}|\p{Other}/gu;
+const getSpecialCharRegExp = () => {
+	try {
+		// This throws when using Node.js without ICU support.
+		// When using a RegExp literal, this would throw at parsing-time, instead of runtime.
+		// eslint-disable-next-line prefer-regex-literals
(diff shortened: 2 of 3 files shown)


test/arguments/escape-no-icu.js
```diff
@@ -0,0 +1,16 @@
...
+	constructor(regExpString, flags) {
+		if (flags?.includes('u') && regExpString.includes('\\p{')) {
+			throw new Error('Invalid property name');
+		}
+
+		super(regExpString, flags);
+	}
+
+	static isMocked = true;
+};
```
ehmicky, line 13: I cannot find a simple way to run tests using Node without ICU support. So this mimics it for our purpose.

test/arguments/escape-no-icu.js
```diff
@@ -0,0 +1,16 @@
...
+
+// Execa computes the RegExp when first loaded, so we must delay this import
+await import('./escape.js');
```
ehmicky, line 16: We could mock/unmock `RegExp` directly in `escape.js`, and run each test twice. But then each test would need to be serial. Also it makes the test file a little complex (this was my first approach). Using a separate test file that imports the other one results in less code.

Uzlopak:
When will it be released? :)

ehmicky:
This has now been released in [`9.3.1`](https://github.com/sindresorhus/execa/releases/tag/v9.3.1). Thanks again for your help.

Uzlopak:
Thank you. Yes :). Great Success. 

PS:
https://x.com/uzlopak_js/status/1823796028182814975

ehmicky:
Thanks for the shout out @Uzlopak, that's very kind! :pray:

Uzlopak:
Unfortunately, you are not on twitter, or else I would have linked you.

PR #1153, opened 2024-09-06
Fix a typo

### LINKED ISSUE #1152: Documentation markup problem in `execa/docs/api.md`
In the documentation of `execaSync`, the following appears: "[SyncOptions](typescript.md". Looks like some Markdown went a bit wrong!

Thanks for execa!

rrthomas:
Fixes #1152.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -754,7 +754,7 @@ If a signal terminated the subprocess, this property is defined and included in 
 
 ## Options
 
-_TypeScript:_ [`Options`](typescript.md) or [`SyncOptions`](typescript.md\
+_TypeScript:_ [`Options`](typescript.md) or [`SyncOptions`](typescript.md)
 _Type:_ `object`
 
 This lists all options for [`execa()`](#execafile-arguments-options) and the [other methods](#methods).
--- a/docs/api.md
+++ b/docs/api.md
@@ -754,7 +754,7 @@ If a signal terminated the subprocess, this property is defined and included in 
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


rrthomas (commented):
A one-character fix.

docs/api.md
```diff
@@ -754,7 +754,7 @@ If a signal terminated the subprocess, this property is defined and included in
 
 ## Options
 
-_TypeScript:_ [`Options`](typescript.md) or [`SyncOptions`](typescript.md\
+_TypeScript:_ [`Options`](typescript.md) or [`SyncOptions`](typescript.md)
```
ehmicky, line 757:
```suggestion
_TypeScript:_ [`Options`](typescript.md) or [`SyncOptions`](typescript.md)\
```

Otherwise, `Type` is on the same line, instead of a different one.

ehmicky:
Thanks @rrthomas. Just a small issue.

ehmicky (approved):

rrthomas:
Thanks, I didn't spot that the `\` was correct, not a typo, and the close paren was simply missing. This is the sort of issue where I think it's probably better for the upstream maintainers to take two minutes to fix a bug rather than 5 minutes to interact with the reporter. My perspective is that I'm using `execa` as part of other libre projects, and I think we'll all have more fun and give more to the world if we spend our time on our own projects *in this case*. I understand of course that the one thing we all lack is compensation and/or assistance, and also that `execa` and other projects you work on are wildly popular, unlike the projects I work on, so the load on your may well be correspondingly greater.

PR #1160, opened 2024-10-14
Fix deno node:process execPath compatibility

w3cj:
* When running `execa` from Deno the following error occurs:

```
The "nodePath" option must be a string or a file URL: /Users/username/deno/bin/deno.
```

* Traced this call back to here: https://github.com/sindresorhus/execa/blob/main/lib/methods/node.js#L29
  * `nodePath` defaults to `execPath` from `node:process`
  * `execPath` in the Deno process pollyfill is a special object, not just a string: https://github.com/denoland/deno/blob/main/ext/node/polyfills/process.ts#L344
* The check for string in `lib/arguments/file-url.js` fails, because execPath is not a string

Note: This issue currently occurs in the Nuxt CLI when trying to run with Deno: https://github.com/denoland/deno/issues/25779

Simple reproduction (run this with Deno v2):

```ts
import { execa } from "npm:execa";

const result = await execa("ls");
console.log({ result });
```

### DIFF
--- a/lib/arguments/file-url.js
+++ b/lib/arguments/file-url.js
@@ -5,6 +5,8 @@ export const safeNormalizeFileUrl = (file, name) => {
 	const fileString = normalizeFileUrl(file);
 
 	const isString = typeof fileString === 'string'
+		// In Deno node:process execPath is a special object, not just a string:
+		// https://github.com/denoland/deno/blob/f460188e583f00144000aa0d8ade08218d47c3c1/ext/node/polyfills/process.ts#L344
 		|| fileString?.__proto__ === String.prototype; // eslint-disable-line no-proto
 
 	if (!isString) {
--- a/lib/arguments/file-url.js
+++ b/lib/arguments/file-url.js
@@ -4,7 +4,10 @@ import {fileURLToPath} from 'node:url';
 export const safeNormalizeFileUrl = (file, name) => {
 	const fileString = normalizeFileUrl(file);
 
-	if (typeof fileString !== 'string') {
+	const isString = typeof fileString === 'string'
+		|| fileString?.__proto__ === String.prototype; // eslint-disable-line no-proto
+
+	if (!isString) {
 		throw new TypeError(`${name} must be a string or a file URL: ${fileString}.`);
 	}
 
--- a/lib/arguments/file-url.js
+++ b/lib/arguments/file-url.js
@@ -2,19 +2,24 @@ import {fileURLToPath} from 'node:url';
 
 // Allow some arguments/options to be either a file path string or a file URL
 export const safeNormalizeFileUrl = (file, name) => {
-	const fileString = normalizeFileUrl(file);
+	const fileString = normalizeFileUrl(normalizeDenoExecPath(file));
 
-	const isString = typeof fileString === 'string'
-		// In Deno node:process execPath is a special object, not just a string:
-		// https://github.com/denoland/deno/blob/f460188e583f00144000aa0d8ade08218d47c3c1/ext/node/polyfills/process.ts#L344
-		|| (fileString && Object.getPrototypeOf(fileString) === String.prototype);
-
-	if (!isString) {
+	if (typeof fileString !== 'string') {
 		throw new TypeError(`${name} must be a string or a file URL: ${fileString}.`);
 	}
 
 	return fileString;
 };
+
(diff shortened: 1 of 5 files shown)


lib/arguments/file-url.js
```diff
@@ -4,7 +4,10 @@ import {fileURLToPath} from 'node:url';
 export const safeNormalizeFileUrl = (file, name) => {
 	const fileString = normalizeFileUrl(file);
 
-	if (typeof fileString !== 'string') {
+	const isString = typeof fileString === 'string'
+		|| fileString?.__proto__ === String.prototype; // eslint-disable-line no-proto
```
sindresorhus, line 8: This needs a short comment that it's a Deno workaround.

w3cj: Updated.

lib/arguments/file-url.js
```diff
@@ -4,7 +4,12 @@ import {fileURLToPath} from 'node:url';
 export const safeNormalizeFileUrl = (file, name) => {
 	const fileString = normalizeFileUrl(file);
 
-	if (typeof fileString !== 'string') {
+	const isString = typeof fileString === 'string'
+		// In Deno node:process execPath is a special object, not just a string:
+		// https://github.com/denoland/deno/blob/f460188e583f00144000aa0d8ade08218d47c3c1/ext/node/polyfills/process.ts#L344
+		|| fileString?.__proto__ === String.prototype; // eslint-disable-line no-proto
```
ehmicky, line 10:
```suggestion
		|| Object.getPrototypeOf(fileString) === String.prototype;
```

This avoids the linting warning, and allows running with `node --disable-proto=throw`.

ehmicky: Sorry about forgetting to handle `undefined`/`null`, did not think of it. :)

ehmicky (approved):

ehmicky:
Looks good to me! 

Thanks a lot @w3cj for making this PR. :pray: 

Should we merge and release @sindresorhus?

sindresorhus:
> Should we merge and release @sindresorhus?

👍

ehmicky:
I have a few additional thoughts.

---

First, what Deno is doing here is quite hacky ([in their own words](https://github.com/denoland/deno/pull/25262#issue-2491922254)). Using an object like this where a string is expected is probably going to create quite many issues: not just that one solved in this PR, but probably with other projects. 

I am wondering whether it would be slightly safer to convert those to a string instead?

```js
export const safeNormalizeFileUrl = (file, name) => {
	const fileString = normalizeFileUrl(normalizeDenoExecPath(file));
	// ...
}

// In Deno node:process execPath is a special object, not just a string:
// https://github.com/denoland/deno/blob/f460188e583f00144000aa0d8ade08218d47c3c1/ext/node/polyfills/process.ts#L344
const normalizeDenoExecPath = file =>  isDenoExecPath(file)
	? file.toString()
	: file

const isDenoExecPath = file => typeof file !== 'string'
	&& file
	&& Object.getPrototypeOf(file) === String.prototype;
```

---

Then, the following syntax needs to be taken into account too:

```js
execa(...).pipe(process.execPath, ...)
```

Right now, it relies on the following check, which would not work with Deno:

https://github.com/sindresorhus/execa/blob/d3a146e2cc7809a59a914f5b8ca26751fc1d5b8a/lib/pipe/pipe-arguments.js#L59-L67

Adding some `isDenoExecPath(firstArgument) || ...` should be enough to fix this.

---

Finally, we should add automated tests for the two following cases:
  - `execa(process.execPath, ...)`
  - `execa(...).pipe(process.execPath, ...)`
  
When `process.execPath` is like Deno. Instead of running with Deno, we could monkey patch `process.execPath` to make it look like Deno.

If you need help with the automated tests, please let me know.

---

What do you think @w3cj?

w3cj:
Yeah that seems reasonable. I wasn't quite sure what else in the code base would be effected. I should have some time to work on this later today.

w3cj:
Updated `safeNormalizeFileUrl` as mentioned and added tests. Please let me know if this is what you had in mind.

ehmicky (approved):

ehmicky:
That's perfect, thanks @w3cj!

ehmicky:
Released in [`9.4.1`](https://github.com/sindresorhus/execa/releases/tag/v9.4.1).

PR #1171, opened 2024-11-14
Upgrade cross-spawn to 7.0.5+

apupier:

PR #1172, opened 2024-11-18
Bump cross-spawn npm ^7.0.6

camelmasa:
node-cross-spawn npm 7.0.3 has a bug, we have updated it to the latest 7.0.6.

Ref: https://github.com/moxystudio/node-cross-spawn/pull/166

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
 	],
 	"dependencies": {
 		"@sindresorhus/merge-streams": "^4.0.0",
-		"cross-spawn": "^7.0.3",
+		"cross-spawn": "^7.0.6",
 		"figures": "^6.1.0",
 		"get-stream": "^9.0.0",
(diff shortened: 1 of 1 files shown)


sindresorhus:
Duplicate of #1171

camelmasa:
@sindresorhus Thanks for reviewing. But, it's not duplicated. cross-spawn npm 7.0.5 is broken. We should bump that latest 7.0.6!

ehmicky:
Hi @camelmasa,

We're using the following version range, which targets `cross-spawn` `7.0.6` thanks to the usage of `^`:

https://github.com/sindresorhus/execa/blob/c8cff27a47b6e6f1cfbfec2bf7fa9dcd08cefed1/package.json#L55

camelmasa:
@ehmicky Okay, We'll adjust our package.json Thanks!

PR #1176, opened 2024-12-06
Fixes argument parsing for escaped newlines

### LINKED ISSUE #1175: Escaped newlines break arguments in version 9
After upgrading from version 8 to 9, some of my multi-line formatted commands have broken.

Reproduction:

```js
import {$} from "execa";

try {
  // 404 is just a stupid way to get an error for some details.
  // Some commands are large, so I break them up by escaping newlines in templates:
  //   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#multi-line_strings
  await $`404 abc\
  def`;
} catch (error) {
  console.log("command", error.command); // 404 abc  ef
  console.log("escaped", error.escapedCommand); // 404 "abc " ef
}
```

Expected command: `404 abc  def` or something close enough
Actual commands: `404 abc  ef` or `404 "abc " ef`

I suspect it may be related to the difference between the template strings and the raw strings passed to a tag function. Example:

```js
function poo(foo, ...bar) {
  console.log("strings", foo.slice());
  console.log("raw", foo.raw);
}

poo`404 abc def`;
poo`404 abc
  def`;
poo`404 abc\
  def`;
```

In the first two examples, the raw strings and the strings arrays match up.
```text
strings (1) ['404 abc def']
raw (1) ['404 abc def']

strings (1) ['404 abc
  def']
raw (1) ['404 abc
  def']
```

In the 3rd sample however the generated strings and the raw strings are different.
```text
strings (1) ['404 abc  def']
raw (1) ['404 abc\\
  def']
```

I suspect there may be some code that may be mixing up the raw strings with the regular strings causing some of the escaping to be offset by a character. But that is just a theory 🤷.

aarondandy:
Fixes #1175.

Trying to handle escaped line breaks in template strings.

### DIFF
--- a/lib/methods/template.js
+++ b/lib/methods/template.js
@@ -79,7 +79,6 @@ const splitByWhitespaces = (template, rawTemplate) => {
 				// Handles escaped newlines in templates
 				templateIndex -= 1;
 				rawIndex += 1;
-				continue;
 			} else if (nextRawCharacter === 'u' && rawTemplate[rawIndex + 2] === '{') {
 				rawIndex = rawTemplate.indexOf('}', rawIndex + 3);
 			} else {
--- a/lib/methods/template.js
+++ b/lib/methods/template.js
@@ -75,7 +75,12 @@ const splitByWhitespaces = (template, rawTemplate) => {
 			templateStart = templateIndex + 1;
 		} else if (rawCharacter === '\\') {
 			const nextRawCharacter = rawTemplate[rawIndex + 1];
-			if (nextRawCharacter === 'u' && rawTemplate[rawIndex + 2] === '{') {
+			if (nextRawCharacter === '\n') {
+				// Handles escaped newlines in templates
+				templateIndex -= 1;
+				rawIndex += 1;
+				continue;
+			} else if (nextRawCharacter === 'u' && rawTemplate[rawIndex + 2] === '{') {
 				rawIndex = rawTemplate.indexOf('}', rawIndex + 3);
 			} else {
 				rawIndex += ESCAPE_LENGTH[nextRawCharacter] ?? 1;
--- a/test/methods/template.js
+++ b/test/methods/template.js
@@ -24,6 +24,14 @@ test('$ can use newlines and space indentations', testScriptStdout, () => $`echo
   bar`, 'foo\nbar');
 test('$ can use escaped newlines and space indentations', testScriptStdout, () => $`echo.js foo\
   bar`, 'foo\nbar');
+test('$ can use escaped newlines and inline tab indentations', testScriptStdout, () => $`echo.js foo\
+	bar`, 'foo\nbar');
+test('$ can use escaped newlines and character escaped tab indentations', testScriptStdout, () => $`echo.js foo\
+\tbar`, 'foo\tbar');
+test('$ can use escaped newlines and character escaped newlines', testScriptStdout, () => $`echo.js foo\
+\n\nbar`, 'foo\n\nbar');
+test('$ can use escaped newlines without indentation', testScriptStdout, () => $`echo.js foo\
+bar`, 'foobar');
 test('$ can use Windows newlines and tab indentations', testScriptStdout, () => escapedCall('echo.js foo\r\n\tbar'), 'foo\nbar');
 test('$ can use Windows newlines and space indentations', testScriptStdout, () => escapedCall('echo.js foo\r\n  bar'), 'foo\nbar');
 test('$ does not ignore comments in expressions', testScriptStdout, () => $`echo.js foo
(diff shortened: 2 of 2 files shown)


aarondandy:
I'm not sure if I have this fix right, even though the test passes.

ehmicky:
Hi @aarondandy,

Thanks, but this should not be needed due to https://github.com/sindresorhus/execa/issues/1175#issuecomment-2523955681

test/methods/template.js
```diff
@@ -22,6 +22,8 @@ test('$ can use newlines and tab indentations', testScriptStdout, () => $`echo.j
 	bar`, 'foo\nbar');
 test('$ can use newlines and space indentations', testScriptStdout, () => $`echo.js foo
   bar`, 'foo\nbar');
+test('$ can use escaped newlines and space indentations', testScriptStdout, () => $`echo.js foo\
+  bar`, 'foo\nbar');
```
aarondandy, line 26: Should the expected actually be `foo bar`?

aarondandy:
Even though this is closed, I still think this indicates a bug in `splitByWhitespaces`. I think my current solution is insufficient though because it does not correctly treat these escaped newlines as token terminators which appears to be the intent of this method. An example that I think may highlight this would be a failure if the next line is unindented. I'm thinking that maybe the behavior that would better fit the intent of this method would be to delimit tokens on these escaped newlines while also correctly adjusting the index values.

test/methods/template.js
aarondandy: Apparently not 🤷.

lib/methods/template.js
```diff
@@ -75,7 +75,12 @@ const splitByWhitespaces = (template, rawTemplate) => {
 			templateStart = templateIndex + 1;
 		} else if (rawCharacter === '\\') {
 			const nextRawCharacter = rawTemplate[rawIndex + 1];
-			if (nextRawCharacter === 'u' && rawTemplate[rawIndex + 2] === '{') {
+			if (nextRawCharacter === '\n') {
```
aarondandy, line 78: This may need to also handle `\r\n` but I'm not sure how to write a test for that.

lib/methods/template.js
```diff
@@ -75,7 +75,12 @@ const splitByWhitespaces = (template, rawTemplate) => {
...
+				// Handles escaped newlines in templates
+				templateIndex -= 1;
+				rawIndex += 1;
+				continue;
```
aarondandy, line 82: This `continue` is probably going to make somebody crabby, but my intention here is to explicitly indicate that I'm doing weird stuff to the control flow of the loop because I'm mutating `templateIndex` and putting it temporarily into an invalid state, anticipating that the for loop will re-increment it again resulting in the value remaining stable while the newlines are skipped.

ehmicky:
Let's continue this discussion on #1175.

lib/methods/template.js
ehmicky: No because escaped newlines inside template strings always use Unix newlines (inside `template.raw`), even on Windows.

lib/methods/template.js
ehmicky: Yes, I think the code is clear enough without it, and the `continue` here is a noop. 

We have very thorough testing for that file, so if we were to add more logic after those lines in the future, we'd catch any potential problem.

Based on this, I think this `continue` should be removed.

test/methods/template.js
ehmicky, line 26: The test looks good. :+1: 

Could you please add the following tests as well?

```js
$`echo.js foo\
  bar` // two spaces, expects: 'foo\nbar'
```

```js
$`echo.js foo\
	bar` // inline tab (not spaces), expects: 'foo\nbar'
```

```js
$`echo.js foo\
\tbar` // escaped tab, expects: 'foo\tbar'
```

```js
$`echo.js foo\
\n\nbar` // escaped newlines, expects: 'foo\n\nbar'
```

aarondandy: Okay. I also added one for the `foobar` case.

test/methods/template.js
```diff
@@ -22,6 +22,16 @@ test('$ can use newlines and tab indentations', testScriptStdout, () => $`echo.j
...
+test('$ can use escaped newlines and inline tab indentations', testScriptStdout, () => $`echo.js foo\
+	bar`, 'foo\nbar');
+test('$ can use escaped newlines and character escaped tab indentations', testScriptStdout, () => $`echo.js foo\
+\tbar`, 'foo\tbar');
+test('$ can use escaped newlines and character escaped newlines', testScriptStdout, () => $`echo.js foo\
+\n\nbar`, 'foo\n\nbar');
+test('$ can use escaped newlines without indentation', testScriptStdout, () => $`echo.js foo\
+bar`, 'foobar');
```
ehmicky, line 34: This test case already exists below with test name `$ handles escaped newlines`.

test/methods/template.js
ehmicky: Thanks!

test/methods/template.js
aarondandy: Okay, I removed mine.

ehmicky (approved):

ehmicky:
Looks good to me. Thanks again @aarondandy.

@sindresorhus Is this good to merge?

sindresorhus (approved):

Issue #1196, opened 2025-04-09
cross-spawn: regular expression denial of service


damien-mathieu1:
The cross-spawn package is in version 7.0.3. This version contains the [CVE-2024-21538](https://avd.aquasec.com/nvd/cve-2024-21538) 

The fixed versions are : 7.0.5, 6.0.6

To fix this I suggest to update cross-spawn package to version 7.0.5

damien-mathieu1:
I will do the PR

damien-mathieu1:
Here is the [PR](https://github.com/sindresorhus/execa/pull/1197)

ehmicky:
Hi @damien-mathieu1,

This is a duplicate of https://github.com/sindresorhus/execa/pull/1172, https://github.com/sindresorhus/execa/pull/1171, https://github.com/sindresorhus/execa/issues/1170, https://github.com/sindresorhus/execa/issues/1180, https://github.com/sindresorhus/execa/pull/1182, https://github.com/sindresorhus/execa/issues/1183, https://github.com/sindresorhus/execa/issues/1184, #1185 and #1187.

See https://github.com/sindresorhus/execa/pull/1172#issuecomment-2484424731

PR #1197, opened 2025-04-09
fix: update cross-spawn version

damien-mathieu1:
Update cross-spawn version


ehmicky:
Hi @damien-mathieu1,

Thanks for the PR but I close it due to https://github.com/sindresorhus/execa/issues/1196#issuecomment-2795261772.

PR #1207, opened 2025-08-10
Fix CI tests

ehmicky:
This fixes some tests currently failing in CI.

### DIFF
--- a/test/arguments/cwd.js
+++ b/test/arguments/cwd.js
@@ -6,6 +6,7 @@ import tempfile from 'tempfile';
 import test from 'ava';
 import {execa, execaSync} from '../../index.js';
 import {FIXTURES_DIRECTORY, setFixtureDirectory} from '../helpers/fixtures-directory.js';
+import {majorNodeVersion} from '../helpers/node-version.js';
 
 setFixtureDirectory();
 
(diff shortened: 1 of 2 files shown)


test/arguments/cwd.js
```diff
@@ -70,7 +71,7 @@ if (!isWindows) {
 }
 
 const cwdNotExisting = {cwd: 'does_not_exist', expectedCode: 'ENOENT', expectedMessage: 'The "cwd" option is invalid'};
-const cwdTooLong = {cwd: '.'.repeat(1e5), expectedCode: 'ENAMETOOLONG', expectedMessage: 'The "cwd" option is invalid'};
+const cwdTooLong = {cwd: '.'.repeat(1e5), expectedCode: isWindows && majorNodeVersion >= 20 ? 'ENOENT' : 'ENAMETOOLONG', expectedMessage: 'The "cwd" option is invalid'};
```
ehmicky, line 74: When calling `child_process.spawn()` with a `cwd` that has a long name, the `error.code` is `ENAMETOOLONG` on Unix. It used to be that on Windows too, but since Node 24.2.0  it's `ENOENT`. I could not find the commit in the Node changelog that changed that behavior, but it's nothing that's impacted Execa's code, so it's ok.

test/ipc/send.js
```diff
@@ -133,7 +133,7 @@ test('Does not hold message events on I/O errors', async t => {
 	const {exitCode, isTerminated, message, ipcOutput} = await t.throwsAsync(subprocess);
 	t.is(exitCode, 1);
 	t.false(isTerminated);
-	t.true(message.includes('Error: getOneMessage() cannot be used: the parent process has already exited or disconnected'));
+	t.true(message.includes('Error: getOneMessage()'));
```
ehmicky, line 136: Those tests rely on a race condition between the time it takes to spawn a process vs to send an IPC message. Those fixes help prevent that.

PR #1208, opened 2025-08-10
[docs] A process is not a TTY usually, just connected to one.

mk-pmb:
I guess you meant that the TTY is connected.

Maybe we should clarify whether the loss of TTY control applies only to stdin/stdout/stderr, as the docs page only talks about those. I'd assume that TTYs on file descriptors 3 and up are inherited without interference, but it would be nice to have an explicit guarantee.

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -95,7 +95,7 @@ const {stdout} = await execa({stdout: ['inherit', {file: 'output.txt'}, 'pipe']}
 console.log(stdout);
 ```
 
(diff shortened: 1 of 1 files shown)


ehmicky (approved):

mk-pmb:
Does your "correct" extend to my assumptions in the 2nd paragraph (about FD 3+)?

ehmicky:
Sorry, I am not completely understanding what you mean. Could you please point me at the specific part of the documentation, and what the current issue might be? Thanks. :)

PR #1209, opened 2025-08-12
[docs] TTY control is lost only on the FD that uses mixed inherit.

mk-pmb:


docs/output.md
```diff
@@ -95,7 +95,7 @@ const {stdout} = await execa({stdout: ['inherit', {file: 'output.txt'}, 'pipe']}
 console.log(stdout);
 ```
 
-When combining [`'inherit'`](#terminal-output) with other values, please note that the subprocess will not be connected to an interactive TTY, even if the current process is.
+__Loss of TTY control:__ Please note that when an I/O channel (file descriptor) is configured with a combination of 'inherit' and other values, this I/O channel will never be a TTY in the subprocess, even if in the current process it is.
```
ehmicky, line 98: I think we might want to stick with "file descriptor" only instead of I/O channel as it is more precise. Also, it is consistent with how this is called in the rest of the documentation.

mk-pmb: fixed.

PR #1215, opened 2025-11-29
Export `VerboseOption`, since it's imported in another file

ndelangen:
We use `execa` in Storybook, and we ran into an issue where the type definitions are not valid.


sindresorhus:
The reason it's not caught is because `.d.ts` files are dumb and only work as ESM if you had `export {};` to all files.

https://github.com/microsoft/TypeScript/issues/57764

PR #1217, opened 2025-12-04
fix: make env value type unknown

dword-design:
Currently only `string` is allowed as values for `env`. But the variables will be stringified by Node.js, so actually anything can be passed.

I removed a test that imho isn't needed anymore. Alternatively, env could get a generic type arg, but as long as the type isn't used in any way in the lib, there isn't a need for it imho.


ehmicky:
Hi @dword-design,

This makes sense to me. Although arguably, some might consider relying on implicit transtyping to be bad practice, we still want to allow it with our types?

dword-design:
It's fine for me not to merge it. At least the discussion is here in case there are people asking. Alternatively, I thought about having the primitive types like `number` or `boolean` in it but again, discusable.

ehmicky:
Thanks for your understanding @dword-design. :pray: 
If someone else has some feedback on this, please feel free to continue discussing here or on a new issue.

PR #1220, opened 2025-12-19
[docs] add warnings about input, stdin, and stdio

### LINKED ISSUE #1218: Undocumented behavior when `input: 'string'` is mixed with `stdio: 'inherit'`
I'm not sure how to make a repro because I've only encountered this issue with SSH.  If I `ssh bash` piping in a script with `input: '<bash script>'` and I have `stdio: 'inherit'`, the command never exits after running the script.  But if I change to `stdio: ['pipe', 'inherit', 'inherit']` it does exit normally.

In retrospect it makes sense that mixing `input` with `stdio: 'inherit'` might not work, but this was fairly confusing, and I don't see any documentation of what's supposed to happen in this case.  Would it be better to error out, or automatically force `stdin` to be `'pipe'` when `input` is given?  Or is there some kind of legitimate use case for mixing `input` with `stdio: 'inherit'`?

jedwards1211:
- warn about and explain behavior when combining `input`/`fileInput` with `stdin: 'inherit'` or `stdio: 'inherit'`
- warn about upcoming `stdio: [..., 'ipc']` deprecation

fix #1218

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -759,6 +759,9 @@ _Type:_ `object`
 
 This lists all options for [`execa()`](#execafile-arguments-options) and the [other methods](#methods).
 
+> [!IMPORTANT] Differences from `child_process.spawn` API
+> - `stdio: [..., 'ipc']` is deprecated and will produce a runtime error in version 10
+
 The following options [can specify different values](output.md#stdoutstderr-specific-options) for [`stdout`](#optionsstdout) and [`stderr`](#optionsstderr): [`verbose`](#optionsverbose), [`lines`](#optionslines), [`stripFinalNewline`](#optionsstripfinalnewline), [`buffer`](#optionsbuffer), [`maxBuffer`](#optionsmaxbuffer).
 
 ### options.preferLocal
@@ -865,6 +868,10 @@ Write some input to the subprocess' [`stdin`](https://en.wikipedia.org/wiki/Stan
 
 See also the [`inputFile`](#optionsinputfile) and [`stdin`](#optionsstdin) options.
 
+> [!NOTE]
+> Be careful when combining `input` with `stdin: 'inherit'` or `stdio: 'inherit'`;
+> See [this section](input.md#combining-input-with-stdin-inherit) for details.
+
 [More info.](input.md#string-input)
 
 ### options.inputFile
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -759,6 +759,9 @@ _Type:_ `object`
 
 This lists all options for [`execa()`](#execafile-arguments-options) and the [other methods](#methods).
 
+> [!IMPORTANT] Differences from `child_process.spawn` API
+> - `stdio: [..., 'ipc']` is deprecated and will produce a runtime error in version 10
```
ehmicky: Could you please remove both `>` and `[!IMPORTANT]`/`[!NOTE]`? It's not a bad thing, but it's not consistent with the rest of the documentation, which does not use this style. Thanks!

docs/input.md
```diff
@@ -90,6 +98,10 @@ The parent process' input can be re-used in the subprocess by passing `'inherit'
 await execa({stdin: 'inherit'})`npm run scaffold`;
 ```
 
+> [!NOTE]
+> Be careful when combining `input` with `stdin: 'inherit'` or `stdio: 'inherit'`;
+> See [this section](#combining-input-with-stdin-inherit) for details.
```
ehmicky: For concision, let's assume users read this file in its entirety. If so, they will read that section below, so we do not need to point to it from those 3 sections above (and also the 2 places in `api.md`). This will also help keep this documentation page concise.

docs/api.md
```diff
@@ -925,6 +936,9 @@ A single string can be used [as a shortcut](output.md#shortcut).
 
 The array can have more than 3 items, to create [additional file descriptors](output.md#additional-file-descriptors) beyond [`stdin`](#optionsstdin)/[`stdout`](#optionsstdout)/[`stderr`](#optionsstderr).
 
+> [!IMPORTANT] Differences from `child_process.spawn` API
+> - `stdio: [..., 'ipc']` is deprecated and will produce a runtime error in version 10
```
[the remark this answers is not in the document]
jedwards1211: I don't really want to; this is the warning I would want to see as a busy developer already familiar with the child_process API encountering this library for the first time.  I see burying the warning as a disservice to other experienced developers.  I wish you took the issue of similarity to another API with subtle differences more seriously.

ehmicky:
Hi again @jedwards1211,

We're back from the holidays! :christmas_tree: 
I've pushed a documentation change at #1222 and #1221, which should solve this PR and #1218.

Thanks again for raising this issue and working on this PR. :pray:

PR #1229, opened 2026-03-20
Docs: Add custom delimiter examples to lines.md

### LINKED ISSUE #1210: How can I progressively iterate over stdout lines with a custom delimiter
I am attempting to monitor progress of `HandBrakeCLI`, which outputs its progress delimited by `\r` instead of `\n`. Is there a way that I can choose a custom delimiter when [progressively iterating](https://github.com/sindresorhus/execa/blob/main/docs/lines.md#progressive-splitting) over the process's `stdout`?

I am unable to find any docs on this, and the quick test I threw together seems to indicate that it is not easily possible.

Currently, my workaround is to pipe the output through `tr '\r', '\n'`, but being able to do this directly with `execa` would make things much simpler.

vincent067:
Hi there! 👋

This PR addresses issue #1210 by adding documentation for using custom delimiters when splitting subprocess output.

## What's added

A new **"Custom delimiters"** section in `docs/lines.md` with:

1. **Array output with custom delimiter** - Shows how to split output by null character (common for filenames with spaces)
2. **Progressive iteration with custom delimiter** - A reusable pattern for streaming processing with any delimiter
3. **Common delimiter examples** - Ready-to-use transforms for CSV, TSV, and paragraph-separated data

## Why this helps

The default newline-based splitting works for most cases, but users often need to process:
- Null-separated filenames (from `find -print0`)
- CSV/TSV data
- Custom protocol outputs

The examples show how to combine `lines: false` with `objectMode: true` transforms to handle these cases.

## Bonus fix

Also fixed a tiny syntax error in the `iterable()` example (had an extra `)`).

Let me know if you'd like any adjustments! Happy to iterate on this. 😊

---

**Related:** Closes #1210

### DIFF
--- a/docs/lines.md
+++ b/docs/lines.md
@@ -36,7 +36,7 @@ Alternatively, [`subprocess.iterable()`](api.md#subprocessiterablereadableoption
 The iteration waits for the subprocess to end (even when using [`break`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break) or [`return`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return)). It throws if the subprocess [fails](api.md#result). This means you do not need to `await` the subprocess' [promise](execution.md#result).
 
 ```js
-for await (const line of execa`npm run build`.iterable())) {
+for await (const line of execa`npm run build`.iterable()) {
 	/* ... */
 }
 ```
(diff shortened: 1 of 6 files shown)


ehmicky:
Hi @vincent067,

Thanks for your contribution.

It appears that this PR includes most of #1230, so could you remove the parts already included in the other PR? So we can review them at the same time, as opposed to one after the other?

docs/lines.md
```diff
@@ -51,6 +51,121 @@ for await (const stderrLine of execa`npm run build`.iterable({from: 'stderr'}))
...
+		if (para.trim()) yield para.trim();
+	}
+};
+
+// Example: parse comma-separated output
+const {stdout: items} = await execa({
+	stdout: {transform: csvTransform, objectMode: true},
+})`echo "apple,banana,cherry,date"`;
+console.log(items); // ['apple', 'banana', 'cherry', 'date']
+```
```
ehmicky, line 167: Those 3 examples are a little bigger in size and more specific than the other examples in our documentation, which are trying to explain the feature with as few lines as possible.

Would you please move them to the `examples` directory instead, then include some link to them from `lines.md`? Thanks!

ehmicky:
Also, please make sure the CI tests are passing. Thanks! :)

sindresorhus:
I'm going to close all of this as it's just AI-generated slop seemingly without any human review.

PR #1232, opened 2026-05-22
Fix input with inherited stdin

### LINKED ISSUE #1219: Fix using both `stdin: 'inherit'` and `input` options
See #1218.

When using any of the following options:
  - `{input: 'foo', stdio: 'inherit'}`   
  - `{input: 'foo', stdio: ['inherit', ...]}`
  - `{input: 'foo', stdin: 'inherit'}`

Execa should silently uses the following instead: `{input: 'foo', stdin: 'pipe', ...}`.

Please note that if users need multiple inputs, they can still use the `{stdin: ['inherit', {string: 'foo'}]` option.

This is technically a breaking change, so should be in a major release.

puneetdixit200:
Fixes #1219

## Summary
- Ignore a singular inherited stdin source when `input` or `inputFile` is provided, so explicit input becomes the only stdin source.
- Preserve explicit multi-source stdin arrays such as `['inherit', 'pipe']`.
- Add nested regression coverage for `stdin` and `stdio` forms in async and sync modes.

## Tests
- `npm_config_cache=/tmp/execa-npm-cache npx ava test/io/input-option.js --match='*input option ignores*'`
- `npm_config_cache=/tmp/execa-npm-cache npx ava test/io/input-option.js test/stdio/native-inherit-pipe.js test/stdio/file-path-main.js`
- `npm_config_cache=/tmp/execa-npm-cache npm run lint`
- `npm_config_cache=/tmp/execa-npm-cache npm run type`
- `git diff --check`

### DIFF
--- a/lib/stdio/handle.js
+++ b/lib/stdio/handle.js
@@ -61,9 +61,10 @@ const getFileDescriptor = ({stdioOption, fdNumber, options, isSync}) => {
 // For example, `stdout: ['ignore']` behaves the same as `stdout: 'ignore'`.
 const initializeStdioItems = ({stdioOption, fdNumber, options, optionName}) => {
 	const values = Array.isArray(stdioOption) ? stdioOption : [stdioOption];
+	const inputStdioItems = handleInputOptions(options, fdNumber);
 	const initialStdioItems = [
-		...values.map(value => initializeStdioItem(value, optionName)),
-		...handleInputOptions(options, fdNumber),
+		...omitInheritedStdin(values.map(value => initializeStdioItem(value, optionName)), inputStdioItems),
+		...inputStdioItems,
 	];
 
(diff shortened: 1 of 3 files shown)


ehmicky (approved):

ehmicky:
Hi @puneetdixit200,

Thanks a lot!
Perfect PR, nothing to fault! :+1: 

This is technically a breaking change, so would require a major release.
We should combine the next major release with dropping support for Node 18 and 20.
We should also fix #1233 (run tests on Node 26) as well.

To allow other non-breaking PRs to still be merged until we do the above, then we probably should wait to merge this PR, but mark it as ready to merge.
@sindresorhus Thoughts?

sindresorhus:
We can start work on the next major now.

PR #1241, opened 2026-07-01
Fix fd-specific options to account for `ipc`

### LINKED ISSUE #1239: `maxBuffer.fd3` and `maxBuffer.ipc` normalize to the same slot when `ipc: true`
## Summary

There appears to be a possible fd-specific option collision between `fd3` and `ipc` for `maxBuffer`.

This is only partially reproduced: the source-level/normalizer behavior is reproducible, but the runtime `maxBuffer` failure was not confirmed locally. The local runtime fixture timed out with `isMaxBuffer: false`, so this is best treated as a semantic confirmation / potential bug report whose runtime impact needs maintainer confirmation.

## Steps to reproduce

From the repository root, compare how `fd3` and `ipc` are normalized:

```js
import { normalizeFdSpecificOption } from './lib/arguments/specific.js'

const fd3 = normalizeFdSpecificOption({ ipc: true, maxBuffer: { fd3: 1 } }, 'maxBuffer')
const ipc = normalizeFdSpecificOption({ ipc: true, maxBuffer: { ipc: 1 } }, 'maxBuffer')

console.log(JSON.stringify({ fd3, ipc }))
```

Observed local output:

```json
{
  "fd3": [100000000, 100000000, 100000000, 1],
  "ipc": [100000000, 100000000, 100000000, 1]
}
```

A runtime check using `ipc-send-twice.js` did not confirm the maxBuffer failure locally; it timed out after 5000ms and reported `isMaxBuffer: false`.

## Expected behavior

`maxBuffer.fd3` and `maxBuffer.ipc` should remain distinct.

If `stdio[3]` does not exist, `{ maxBuffer: { fd3: 1 } }` should probably either trigger the existing "file descriptor does not exist" validation error or otherwise not affect IPC buffering. Only `{ maxBuffer: { ipc: 1 } }` should limit `ipcOutput`.

## Actual behavior

At the normalizer level, `{ maxBuffer: { fd3: 1 } }` and `{ maxBuffer: { ipc: 1 } }` produce the same normalized array:

```json
[100000000, 100000000, 100000000, 1]
```

This appears to happen because the fd-specific array includes an extra IPC slot. With default stdio length, numeric `fd3` writes index `3`, while `ipc` also reads the last slot, which is index `3`.

## Evidence

Relevant source anchors:

- `types/arguments/specific.d.ts`: public fd-specific keys include both fd names and `ipc`, implying distinct option keys.
- `lib/arguments/specific.js`: `normalizeFdSpecificOption()` creates `getStdioLength(options) + 1` slots.
- `lib/arguments/specific.js`: `parseFdName('ipc')` maps IPC to `optionArray.length - 1`.
- `lib/arguments/specific.js`: `parseFdName('fd3')` maps to numeric index `3`; with default stdio plus the extra IPC slot, this passes validation.
- `lib/arguments/specific.js`: `getFdSpecificValue(optionArray, 'ipc')` reads `optionArray.at(-1)`.
- `lib/ipc/buffer-messages.js`: IPC buffering reads `maxBuffer` via `getFdSpecificValue(maxBufferArray, 'ipc')`.
- Existing `maxBuffer` coverage checks explicit `{ maxBuffer: { ipc: 1 } }`, but not the `fd3` collision case.

## Suggested fix

One possible fix is to validate numeric fd names against the real stdio length before adding the internal IPC slot.

Another option is to keep IPC-specific values outside the numeric fd array, so numeric fd keys cannot alias the IPC slot.

Suggested regression coverage:

- `{ ipc: true, maxBuffer: { fd3: 1 } }` without `stdio[3]` should not limit `ipcOutput`.
- `{ ipc: true, maxBuffer: { ipc: 1 } }` should continue to limit `ipcOutput`.

## Additional context / Related coverage

A current GitHub search did not identify exact issue/PR coverage for the `fd3` + `ipc` + `maxBuffer` collision.

Again, the runtime impact still needs maintainer confirmation because the latest local runtime fixture attempt timed out instead of producing a maxBuffer error.

---
Submitted with Codex.

sindresorhus:
Fixes https://github.com/sindresorhus/execa/issues/1239

### DIFF
--- a/lib/arguments/specific.js
+++ b/lib/arguments/specific.js
@@ -16,8 +16,9 @@ export const normalizeFdSpecificOptions = options => {
 };
 
 export const normalizeFdSpecificOption = (options, optionName) => {
-	const optionBaseArray = Array.from({length: getStdioLength(options) + 1});
-	const optionArray = normalizeFdSpecificValue(options[optionName], optionBaseArray, optionName);
+	const stdioLength = getStdioLength(options);
+	const optionBaseArray = Array.from({length: stdioLength + 1});
+	const optionArray = normalizeFdSpecificValue(options[optionName], optionBaseArray, optionName, stdioLength);
 	return addDefaultValue(optionArray, optionName);
 };
 
@@ -25,13 +26,13 @@ const getStdioLength = ({stdio}) => Array.isArray(stdio)
 	? Math.max(stdio.length, STANDARD_STREAMS_ALIASES.length)
 	: STANDARD_STREAMS_ALIASES.length;
 
-const normalizeFdSpecificValue = (optionValue, optionArray, optionName) => isPlainObject(optionValue)
-	? normalizeOptionObject(optionValue, optionArray, optionName)
+const normalizeFdSpecificValue = (optionValue, optionArray, optionName, stdioLength) => isPlainObject(optionValue)
+	? normalizeOptionObject(optionValue, optionArray, optionName, stdioLength)
 	: optionArray.fill(optionValue);
 
-const normalizeOptionObject = (optionValue, optionArray, optionName) => {
+const normalizeOptionObject = (optionValue, optionArray, optionName, stdioLength) => {
 	for (const fdName of Object.keys(optionValue).sort(compareFdName)) {
-		for (const fdNumber of parseFdName(fdName, optionName, optionArray)) {
+		for (const fdNumber of parseFdName(fdName, optionName, stdioLength)) {
 			optionArray[fdNumber] = optionValue[fdName];
 		}
 	}
@@ -50,9 +51,9 @@ const getFdNameOrder = fdName => {
 	return fdName === 'all' ? 2 : 1;
 };
 
-const parseFdName = (fdName, optionName, optionArray) => {
+const parseFdName = (fdName, optionName, stdioLength) => {
 	if (fdName === 'ipc') {
-		return [optionArray.length - 1];
+		return [stdioLength];
 	}
 
 	const fdNumber = parseFd(fdName);
@@ -61,7 +62,7 @@ const parseFdName = (fdName, optionName, optionArray) => {
 It must be "${optionName}.stdout", "${optionName}.stderr", "${optionName}.all", "${optionName}.ipc", or "${optionName}.fd3", "${optionName}.fd4" (and so on).`);
 	}
 
-	if (fdNumber >= optionArray.length) {
+	if (fdNumber !== 'all' && fdNumber >= stdioLength) {
 		throw new TypeError(`"${optionName}.${fdName}" is invalid: that file descriptor does not exist.
 Please set the "stdio" option to ensure that file descriptor exists.`);
 	}
--- a/test-d/arguments/specific.test-d.ts
+++ b/test-d/arguments/specific.test-d.ts
@@ -9,7 +9,7 @@ await execa('unicorns', {maxBuffer: {stdout: 0, stderr: 0} as const});
 await execa('unicorns', {maxBuffer: {all: 0}});
 await execa('unicorns', {maxBuffer: {fd1: 0}});
 await execa('unicorns', {maxBuffer: {fd2: 0}});
-await execa('unicorns', {maxBuffer: {fd3: 0}});
+await execa('unicorns', {maxBuffer: {fd3: 0}, stdio: ['pipe', 'pipe', 'pipe', 'pipe']});
 await execa('unicorns', {maxBuffer: {ipc: 0}});
 expectError(await execa('unicorns', {maxBuffer: {stdout: '0'}}));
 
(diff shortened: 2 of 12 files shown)


sindresorhus:
@ehmicky Going to merge some PRs right away to get things done, but if you have time, a review is always welcome, but I don't want to wait or stress you about it.

lib/arguments/specific.js
```diff
@@ -61,7 +62,7 @@ const parseFdName = (fdName, optionName, optionArray) => {
 It must be "${optionName}.stdout", "${optionName}.stderr", "${optionName}.all", "${optionName}.ipc", or "${optionName}.fd3", "${optionName}.fd4" (and so on).`);
 	}
 
-	if (fdNumber >= optionArray.length) {
+	if (fdNumber !== 'all' && fdNumber >= stdioLength) {
```
ehmicky, line 65: Good catch.

test/verbose/output-enable.js
```diff
@@ -105,7 +106,7 @@ test('Does not escape internal characters from stdout', async t => {
 });
 
 test('Strips color sequences from stdout', async t => {
-	const {stderr} = await nestedSubprocess('noop.js', [foobarRed], {verbose: 'full'});
+	const {stderr} = await nestedSubprocess('noop.js', [red(foobarString)], {verbose: 'full'}, {env: {FORCE_COLOR: '1', NO_COLOR: undefined}});
```
ehmicky, line 109: It appears that some of this PR is accidentally reverting #1242?
I.e. it re-introduces `yoctocolors` in the tests and removes the usage of the `foobarRed` variable.

test/verbose/start.js
```diff
@@ -26,8 +27,15 @@ import {
 
 setFixtureDirectory();
 
+const redFoobarString = `\u001B[31m${foobarString}\u001B[39m`;
```
ehmicky, line 30: Isn't this the same as `foobarRed`?

test-d/arguments/specific.test-d.ts
```diff
@@ -9,7 +9,7 @@ await execa('unicorns', {maxBuffer: {stdout: 0, stderr: 0} as const});
 await execa('unicorns', {maxBuffer: {all: 0}});
 await execa('unicorns', {maxBuffer: {fd1: 0}});
 await execa('unicorns', {maxBuffer: {fd2: 0}});
-await execa('unicorns', {maxBuffer: {fd3: 0}});
+await execa('unicorns', {maxBuffer: {fd3: 0}, stdio: ['pipe', 'pipe', 'pipe', 'pipe']});
```
ehmicky, line 12: Maybe we should keep the original assertion as well (without the `stdio` option), to make sure `maxBuffer.fd3` type works, even when `stdio` is not set?
Same for the other changes in that same file.

lib/arguments/specific.js
```diff
@@ -50,9 +51,9 @@ const getFdNameOrder = fdName => {
 	return fdName === 'all' ? 2 : 1;
 };
 
-const parseFdName = (fdName, optionName, optionArray) => {
+const parseFdName = (fdName, optionName, stdioLength) => {
 	if (fdName === 'ipc') {
-		return [optionArray.length - 1];
+		return [stdioLength];
```
ehmicky, line 56: Looks good. 👍

test/helpers/verbose.js
```diff
@@ -57,6 +60,10 @@ export const runVerboseSubprocess = ({
 	...options,
 });
 
+export const getStdioForFd3Option = (...values) => values.some(value => hasFd3Option(value)) ? fullStdio : {};
```
ehmicky, line 63: If I understand correctly, this is added to tests that use some `{option: {fd3: ....}}` option, to make them also pass `{stdio: ['pipe', 'pipe', 'pipe', 'pipe']}`. However, shouldn't `fd3` work even without users having to pass that `stdio` option?

types/arguments/specific.d.ts
```diff
@@ -29,26 +29,28 @@ type FdSpecificObjectOption<
...
+		: FdNumber extends '2'
+			? 'stderr' extends GenericOptionKeys
+				? 'stderr'
+				: 'fd2' extends GenericOptionKeys
+					? 'fd2'
+					: 'all' extends GenericOptionKeys
+						? 'all'
+						: never
+			: `fd${FdNumber}` extends GenericOptionKeys
+				? `fd${FdNumber}`
```
ehmicky, line 53: Do you remember the reason behind this change? Wouldn't the new code be equivalent, in principle? 

I am guessing this might have to do with TypeScript doing something weird, as it sometimes does with complex logic, especially with nested conditionals. :)

test/helpers/verbose.js
sindresorhus: Not for runtime behavior. `fd3` should only be valid when the child actually has fd 3, for xample via `stdio: ['pipe', 'pipe', 'pipe', 'pipe']`. With default stdio, `fd3` should throw. The bug was that `{ipc: true}` added an internal pseudo-slot at index 3, so `fd3` accidentally targeted IPC. This helper just makes the fd3 verbose matrix tests create a real fd3; the regression tests cover the default-stdio throw case.

test/verbose/start.js
sindresorhus: Yeah, merge conflict mistake. Will fix.

types/arguments/specific.d.ts
sindresorhus: They are not equivalent. The old fallback let numeric fd lookups fall through to `ipc`, so `buffer: {ipc: false}` could make `stdio[3]` typed as `undefined`.

This keeps `ipc` separate: `ipc` only affects `ipcOutput`, while numeric fds only match real `fdN`/`stdout`/`stderr`/`all`. The tsd regression covers tat.

ehmicky: :+1:

test/helpers/verbose.js
ehmicky: :+1:

PR #1243, opened 2026-07-01
Require Node.js 22

sindresorhus:
Most of the changes are lint fixes. Some ignores will be able to be removed when XO 4 is out.


ehmicky:
Nice cleanup.
I checked it and it all looks good. :+1:

PR #1245, opened 2026-07-02
Remove the old `stdio: [..., 'ipc']` syntax

### LINKED ISSUE #1056: Deprecating old `ipc` option
We've replaced the `{stdio: [..., 'ipc']}` option with `{ipc: true}`. The old syntax has been undocumented. However, it's still currently supporting, for backward compatibility. 

In the next major release, should we remove support the old syntax? Possible options:
  1. Keep as is: undocumented only
  2. Remove the types only
  3. Remove the types, and also throw an error message when using that old syntax at runtime

Which one do you prefer @sindresorhus?

Note: we probably should hold on the next major release for many months. We just had a big major release with many breaking changes. So this issue is for future planning, nor immediate action.

sindresorhus:
It was undocumented and replaced by the `ipc: true` option. Using `'ipc'` as a `stdio` value now throws.

Fixes #1056


lib/stdio/stdio-option.js
```diff
@@ -6,11 +6,19 @@ import {isFullVerbose} from '../verbose/values.js';
...
+	validateIpcStdioOption(stdioArray);
 	return isSync
 		? normalizeStdioSync(stdioArray, buffer, verboseInfo)
 		: normalizeIpcStdioArray(stdioArray, ipc);
 };
 
+// The `stdio: 'ipc'` value (raw `child_process` syntax) was replaced by the `ipc: true` option
+const validateIpcStdioOption = stdioArray => {
+	if (stdioArray.some(stdioOption => Array.isArray(stdioOption) ? stdioOption.includes('ipc') : stdioOption === 'ipc')) {
+		throw new Error('The `ipc: true` option must be used instead of `stdio: \'ipc\'`.');
```
ehmicky, line 18: Nice warning message.

ehmicky:
Great! This looks good. :+1:

PR #1246, opened 2026-07-03
Add `{value: 'pipe', input: true}` to use an additional file descriptor as input

### LINKED ISSUE #1214: Using duplex() with fd3: it must be a writable stream
I am trying to use `execa` within a `pipeline`, which results in very elegant code.

However, I am facing a problem when using an input file descriptor different from stdin. The following snippet contains a 

<details><summary>minimal reproducer</summary>
<p>

```js
// Importing dependencies
import fs from 'node:fs';
import dedent from 'dedent';
import { execa } from 'execa';
import { pipeline } from 'node:stream/promises';
import { Readable } from 'node:stream';

// Run execa as duplex stream in a pipeline
// The pipeline source and sink, as well as the subprocess (Python here) are not important,
// instead the focus is on the "to" option for duplex()
await pipeline(
    Readable.from('This is a test'),
    execa(
        'python3',
        {
            input: dedent(`
                import os
                with os.fdopen(3, 'r') as file:
                    content = json.load(file)
                print(content)
            `),
            stdio: ['pipe', 'pipe', 'pipe', 'pipe'],
        },
    ).duplex({ to: 'fd3' }),
    process.stdout,
);
```

</p>
</details> 

Running the code on GNU/Linux 6.14.0 with Node.js v22.15.0 and execa 9.6.0 causes the following

<details><summary>error message</summary>
<p>

```
file:///tmp/test/node_modules/execa/lib/arguments/fd-options.js:66
                throw new TypeError(`"${getOptionName(isWritable)}" must not be ${fdName}. It must be a writable stream, not readable.`);
                      ^

TypeError: "to" must not be fd3. It must be a writable stream, not readable.
    at validateFdNumber (file:///tmp/test/node_modules/execa/lib/arguments/fd-options.js:66:9)
    at getFdNumber (file:///tmp/test/node_modules/execa/lib/arguments/fd-options.js:36:2)
    at getToStream (file:///tmp/test/node_modules/execa/lib/arguments/fd-options.js:7:19)
    at getSubprocessStdin (file:///tmp/test/node_modules/execa/lib/convert/writable.js:32:26)
    at createDuplex (file:///tmp/test/node_modules/execa/lib/convert/duplex.js:22:68)
    at main (file:///tmp/test/[eval1]:29:11)
    at file:///tmp/test/[eval1]:37:2
    at ModuleJob.run (node:internal/modules/esm/module_job:274:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:268:26)
    at async ModuleLoader.executeModuleJob (node:internal/modules/esm/loader:264:20)
```

</p>
</details> 

I tracked the source, and noticed that by simply removing the [validation rules](https://github.com/sindresorhus/execa/blob/103095f2c9a459ae3500fa7d861f162c21a94b3a/lib/arguments/fd-options.js#L61-L67), the code runs and behaves as expected. Unfortunately, I am not familiar enough with the source code to suggest a solution, so I would appreciate any help by the maintainers or more experienced users.

Thank you for a wonderful library, and its extensive documentation.

sindresorhus:
For file dscriptors beyond `stdin`/`stdout`/`stderr`, `'pipe'` is ambiguous: Execa cannot tell whether it is meant for input or output, so it defaults to output. This made it impossible to write into an extra file descriptor, e.g. `.duplex({to: 'fd3'})`, without the buggy `['pipe', []]` workaround.

The new `{value: 'pipe', input: true}` object form explicitly marks such a file descriptor as input.

Fixes #1214

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -916,14 +916,14 @@ More info on [available values](output.md), [streaming](streams.md) and [transfo
 ### options.stdio
 
 _TypeScript:_ [`Options['stdio']`](typescript.md) or [`SyncOptions['stdio']`](typescript.md)\
-_Type:_ `string | Array<string | number | stream.Readable | stream.Writable | ReadableStream | WritableStream | TransformStream | URL | {file: string} | Uint8Array | Iterable<string> | Iterable<Uint8Array> | Iterable<unknown> | AsyncIterable<string | Uint8Array | unknown> | GeneratorFunction<string | Uint8Array | unknown> | AsyncGeneratorFunction<string | Uint8Array | unknown> | {transform: GeneratorFunction | AsyncGeneratorFunction | Duplex | TransformStream}>` (or a tuple of those types)\
+_Type:_ `string | Array<string | number | stream.Readable | stream.Writable | ReadableStream | WritableStream | TransformStream | URL | {file: string} | Uint8Array | Iterable<string> | Iterable<Uint8Array> | Iterable<unknown> | AsyncIterable<string | Uint8Array | unknown> | GeneratorFunction<string | Uint8Array | unknown> | AsyncGeneratorFunction<string | Uint8Array | unknown> | {transform: GeneratorFunction | AsyncGeneratorFunction | Duplex | TransformStream} | {value: 'pipe'; input?: boolean}>` (or a tuple of those types)\
 _Default:_ `pipe`
 
 Like the [`stdin`](#optionsstdin), [`stdout`](#optionsstdout) and [`stderr`](#optionsstderr) options but for all [file descriptors](https://en.wikipedia.org/wiki/File_descriptor) at once. For example, `{stdio: ['ignore', 'pipe', 'pipe']}` is the same as `{stdin: 'ignore', stdout: 'pipe', stderr: 'pipe'}`.
 
 A single string can be used [as a shortcut](output.md#shortcut).
 
-The array can have more than 3 items, to create [additional file descriptors](output.md#additional-file-descriptors) beyond [`stdin`](#optionsstdin)/[`stdout`](#optionsstdout)/[`stderr`](#optionsstderr).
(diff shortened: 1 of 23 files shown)


docs/api.md
```diff
@@ -916,14 +916,14 @@ More info on [available values](output.md), [streaming](streams.md) and [transfo
...
 _Default:_ `pipe`
 
 Like the [`stdin`](#optionsstdin), [`stdout`](#optionsstdout) and [`stderr`](#optionsstderr) options but for all [file descriptors](https://en.wikipedia.org/wiki/File_descriptor) at once. For example, `{stdio: ['ignore', 'pipe', 'pipe']}` is the same as `{stdin: 'ignore', stdout: 'pipe', stderr: 'pipe'}`.
 
 A single string can be used [as a shortcut](output.md#shortcut).
 
-The array can have more than 3 items, to create [additional file descriptors](output.md#additional-file-descriptors) beyond [`stdin`](#optionsstdin)/[`stdout`](#optionsstdout)/[`stderr`](#optionsstderr).
+The array can have more than 3 items, to create [additional file descriptors](output.md#additional-file-descriptors) beyond [`stdin`](#optionsstdin)/[`stdout`](#optionsstdout)/[`stderr`](#optionsstderr). Passing `'pipe'` to an additional file descriptor defaults it to output. To use it for [input](input.md#additional-file-descriptors) instead, pass `{value: 'pipe', input: true}`.
```
ehmicky, line 926: Can only `'pipe'` be ambiguous? For example, I would imagine `'inherit'` could be possibly used for both input and output in fd3? Same with streams that are both readable and writable, or with files.

sindresorhus: I considered that, but I scoped it to 'pipe'/'overlapped' because that's what #1214 was about (the buggy ['pipe', []] workaround), but 'inherit' and files indeed share the same limitation. I'll look into it.

ehmicky: I was only mentioning `'pipe'` in the issue because that was the specific instance the user was experiencing, but the problem was more generic, in that it impacts other values.

I am actually wondering whether allowing any value would simplify the implementation, since some of the logic right now is making sure `'pipe'` is used. Technically, some values do not make sense as input, namely `1`, `2` or any non-readable stream (Node.js or web). But using fd3 + `{input: true}` + one of those values seems like such an edge case, which is going to fail runtime anyway, that it might not deserve any safeguard.

sindresorhus: I generalized this in #1249: `{value, input}`

I still kept the cheap guard for intrinsically output-only values like writable streams, since that remains useful validation and gives a clearer error.

PR #1248, opened 2026-07-03
Document and test Windows fixes

### LINKED ISSUE #997: Document and tests Windows fixes
We use `node-cross-spawn` to fix a few issues with cross-platform support, specifically Windows. This includes shebang support, `PATHEXT`, running commands with spaces in them, and [more](https://github.com/moxystudio/node-cross-spawn?tab=readme-ov-file#why).

My understanding of what `node-cross-spawn` is:
- Call `node-which` to resolve the executable's absolute file path. This fixes `PATHEXT` support.
- Read the first 150 bytes of that file, to detect any shebang. This fixes shebang support.
- Unless the file is `*.exe` or `*.com`, runs the file with `cmd.exe /d /s /c`. Unlike the `shell: true` option implemented by Node.js, this does proper escaping.

All of the above is only done on Windows. Also, it is not done if `shell: true` is used. Running `cmd.exe` fixes `PATHEXT` support, so using `shell: true` is enough, although I don't think it supports shebangs, and also the escaping done by Node.js is not as good.

From reading the [original Node.js issue](https://github.com/nodejs/node-v0.x-archive/issues/2318) that both introduced `shell: true` and led to the creation of `node-cross-spawn`, this is a very complex topic. The [following comment](https://github.com/nodejs/node-v0.x-archive/issues/2318#issuecomment-32706753) has a complex list of Windows' behavior, some of it appears to be undocumented. The Windows syscalls apparently have some inherent limitations. That issue describes actually many different problems, which probably should have been separate issues as pointed out by [this comment](https://github.com/nodejs/node-v0.x-archive/issues/2318#issuecomment-44227527).

We need to get a better understand on what `node-cross-spawn` is actually fixing and:
- Add tests in Execa covering those cases
- Better document them in the [Windows guide](https://github.com/sindresorhus/execa/blob/main/docs/windows.md) (and also in the `readme.md`). At the moment, we only mention shebang support.
- In the [Shells guide](https://github.com/sindresorhus/execa/blob/main/docs/shell.md), better document why `shell: true` might, or might not, be needed in specific cases based on the above.

sindresorhus:
Fixes #997

### DIFF
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -15,6 +15,16 @@ In general, [shells](https://en.wikipedia.org/wiki/Shell_(computing)) should be 
 
 In almost all cases, plain JavaScript is a better alternative to shells. The [following page](bash.md) shows how to convert Bash into JavaScript.
 
+## When a shell is not needed
+
+On Windows, a shell is often used to work around limitations of the OS. This is not needed with Execa. In particular, `shell: true` is not required to:
+- run `.cmd` and `.bat` files, such as `npm.cmd`
+- [resolve a command](windows.md#file-extensions) by name using `PATHEXT`
+- run [shebang](windows.md#shebang) scripts
+- [escape](windows.md#escaping) file paths and arguments
+
+A shell is only needed to use shell-specific syntax, such as pipes `|`, redirections `>`, globbing `*` or environment variable expansion `$VAR`.
+
 ## Specific shell
 
 ```js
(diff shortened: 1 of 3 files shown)


ahm1982mbs-stack (approved):

ahm1982mbs-stack (approved):

ehmicky:
Nice, concise documentation. :+1: 
Good to add more tests too. In the future, we probably should add more Windows-related tests if we're going to inline `node-cross-spawn`, to cover all the complex cases that library is supporting.

PR #1249, opened 2026-07-04
Generalize `{value, input}` to any direction-ambiguous stdio value

sindresorhus:
In previouss changes, `{value, input: true}` only worked for `'pipe'`/`'overlapped'`. Other values with an ambiguous direction on an additional file descriptor (`'inherit'`, file descriptor numbers, files, transforms) also default to output with no clean way to force input.


ehmicky:
Nice! :)

PR #1251, opened 2026-07-04
Move away from `cross-spawn`

### LINKED ISSUE #578: Get rid of `cross-spawn` dependency
It's pretty unmaintained.

We do a lot to the shell, so would be simpler to do that too in execa ourselves.

https://github.com/moxystudio/node-cross-spawn/blob/master/lib/parse.js

sindresorhus:
Fixes #578

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -35,7 +35,7 @@ import {execa} from 'execa';
 await execa`npm run build`;
 ```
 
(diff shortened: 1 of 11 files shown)


ehmicky:
That's amazing. I can't believe all our current tests are passing!
Also the new tests are great.

